PR #4, opened 2019-08-14
Reports page

fardarter:


input/number/reports.md
```diff
@@ -0,0 +1,28 @@
+## This page should be used for casual reporting of issues.
```
gregwhitworth: I don't want pages that are casual reports of issues. More of investigation into the current state of things that we can ultimately bubble up into solutions. So I'd delete this - Also, take the template currently being used in `<select>`. Ultimately I'd like to get @mfreed7 content into a similar reporting page of the current state of things into the repo. That said, there is valuable discussion occuring in the Google doc so I'm not sure if it's worth the time yet.

input/number/reports.md
```diff
@@ -0,0 +1,28 @@
...
+
+For help on formatting, see:
```
gregwhitworth: Remove this section regarding formatting

input/number/reports.md
```diff
@@ -0,0 +1,28 @@
...
+
+For help on formatting, see: 
+- [How to put multiple lines in a Markdown table cell (multiline table)](https://alvinalexander.com/text/how-put-multiple-lines-markdown-table-cell-multiline-table)
+- [Markdown table generator](https://www.tablesgenerator.com/markdown_tables)
+
+### <a name="valid-issues">Valid issue types:</a> 
+- Inconsistency with other similar input elements
+- Non-extensibly (functionality)
+- Non-extensibly (styling)
+- Lacks built-in accessibility
```
gregwhitworth: This is a good example of something that would need to be dug into as I actually don't believe this is true. This may however be true on certain OSes or certain browsers which is why doing a current state dig into the input is valuable

input/number/reports.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Structure of a report
+
+#### Please use a <a href="#valid-issues">valid issue type</a>.
+
+##### At least one link is mandatory. 
+
+|  An example report                                                                                                                    |
+|---------------------------------------------------------------------------------------------------------------------|
+| <p><strong>Issue</strong>: `<valid issue type>`<br><strong>Links</strong>:<br>- `<link1>`<br>- `<link2>`<br>- `<link3>`<br><strong>Description</strong>:`<some description>`</p> |
```
gregwhitworth: I'd remove all of this, if we feel a specific structure is desired (which I'm thinking we will) then we should spin up a template file that folks copy and start from

input/number/reports.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+---
+
+|  Reports                                                                                                                    |
+|---------------------------------------------------------------------------------------------------------------------|
+| <p><strong>Issue</strong>: `Inconsistency with other similar input elements`<br> <strong>Links</strong>:<br>- [Input type “number” won't resize](https://stackoverflow.com/questions/22709792/input-type-number-wont-resize)<br>- [Why doesn't the <input> element respect min-width?](https://stackoverflow.com/questions/29470676/why-doesnt-the-input-element-respect-min-width)<br><strong>Description</strong>: Can't use the same CSS to handle inputs type='number' as the others.</p> |
```
gregwhitworth: Adjust this to known issues not reports.

gregwhitworth: The stackoverflow issue is actually per [spec](https://html.spec.whatwg.org/multipage/input.html#number-state-(type=number)), specifically this line:

> The following content attributes must not be specified and do not apply to the element: accept, alt, checked, dirname, formaction, formenctype, formmethod, formnovalidate, formtarget, height, maxlength, minlength, multiple, pattern, size, src, and width.

Because of this I'd recommend filing an issue against the [HTML Spec](https://github.com/whatwg/html/issues)

input/number/reports.md
fardarter: I can work more closely to the template on the `select` page, and could rephrase, but the idea of a place where someone can very casually drop an item seems a good one to me.

It should be curated in that way. Simply as a research/notepad for people to make observations.

I agree with the deep-dive structure, so I think an `overview.md` might be good for this folder also, but this isn't intended to create that sort of space.

Do you agree that kind of space is something you might want?

input/number/reports.md
fardarter: Right, so my idea for a report is that it isn't necessarily true. It's just someone saying something. It would need further investigation were we/you to recognise it as an issue.

input/number/reports.md
fardarter: Fair, enough, but what would you like in the meantime?

input/number/reports.md
fardarter: As per above, not sure it is an issue yet. You commented below that it's per spec, which I'll ask about below, but feel like I should be able to drop something here that bothers me in working with forms.

Or if there were a different space?

fardarter: Right, maybe I'm not understanding the project, but if it's about tracking normalisation issues, if the spec isn't normalised, then shouldn't it be tracked here until the spec is updated?

input/number/reports.md
fardarter: Will make (and resolve) changes once/if we agree my idea has some value.

input/number/reports.md
gregwhitworth: @fardarter Starting off I think I'd prefer to have a label of issues, eg: I can open issues and filter to input=number to see issues people are hitting (these could both be bugs or limitations people are hitting [eg: I can't style the spinners, each browser ships different parts, etc]). The area where what you're proposing becomes valuable, at least to me, is when we've determined that to create an improved input=number (based on diggind and taking into account the known bugs [as GH issues]) requires us to solve the `size` attribute issue which should then be tracked on the document as tasks that need to land to complete the improved control. Whether that be a new feature & spec or bug fixing, each control will be different. I also still recommend opening the issue on WHATWG as well but I think having a list of known number issues within this repo will be valuable to inform the document but also allow them to be referenced across repos (this is one thing I find valuable about issues, not just a notepad). Does that make sense?

**To hopefully clarify it fully, the flow would look like this:**
* Look through filed issues on our repo, as well as bug databases, usecases, rants, work arounds, etc
* Build out the solid list of unique use cases
* Identify missing platform features & known issues
* Scope the problem based on impact (eg: do we actually need to spec and implement all missing platform features in order to improve either the end user or developer experience)
* Begin determining how to solve that problem

gregwhitworth:
Based on where the current state of this project is going I'm going to close this one down since I do still think the issues should be filed against the WHATWG.

PR #14, opened 2019-10-01
Consider Adding Accessibility section

micahgodbolt:
We've worked with a team to collect accessibility recommendations for individual controls. Controls that adhere to the 'open ui' spec should also follow these accessibility specs. 

These initial rules are certainly up for debate. This PR is mostly to kick off the discussion around the addition of these guidances.

### DIFF
--- a/research/src/pages/button.mdx
+++ b/research/src/pages/button.mdx
@@ -16,6 +16,19 @@ import ComponentCoverage from '../components/component-coverage'
 
 <Anatomy component="Button" />
 
+## Accessibility
+
+- Use a touch target area of `44px` by `44px`
+- For multiple buttons provided on the page that perform the same function, one of the targets should be `44px` by `44px`.
+- Buttons should get activated using `Enter` key as well as `SPACE` key.
+- For custom buttons made using `<div>`, `<span>`, `<svg>`or `<img>`,
+  - `role="button"` must be applied.
+  - Tabindex property should be set to "0", so that it could get tab focus.
+  - When secondary action is present, use the `aria-describedby` property.
+  - Provide tooltips for button icons that don’t have a visible label or hint text about the event it triggers.
+  - Provide an accessible label using `aria-label`, and make sure the accessible label string starts with the visible label string
+- Con
+
 ## Concepts
 
 <Concepts component="Button" />
--- a/research/src/pages/button.mdx
+++ b/research/src/pages/button.mdx
@@ -27,7 +27,8 @@ import ComponentCoverage from '../components/component-coverage'
   - When secondary action is present, use the `aria-describedby` property.
   - Provide tooltips for button icons that don’t have a visible label or hint text about the event it triggers.
   - Provide an accessible label using `aria-label`, and make sure the accessible label string starts with the visible label string
-- Con
+- Consider putting a comma `,` wherever required to break the string, in case of longer description, to make a screen reader take a little pause.
+
 
 ## Concepts
 
--- a/research/src/pages/button.mdx
+++ b/research/src/pages/button.mdx
@@ -23,7 +23,7 @@ import ComponentCoverage from '../components/component-coverage'
 - Buttons should get activated using `Enter` key as well as `SPACE` key.
 - For custom buttons made using `<div>`, `<span>`, `<svg>`or `<img>`,
   - `role="button"` must be applied.
-  - Tabindex property should be set to "0", so that it could get tab focus.
+  - `tabindex` property should be set to "0", so that it could get tab focus.
   - When secondary action is present, use the `aria-describedby` property.
   - Provide tooltips for button icons that don’t have a visible label or hint text about the event it triggers.
   - Provide an accessible label using `aria-label`, and make sure the accessible label string starts with the visible label string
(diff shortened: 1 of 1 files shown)


marcoscaceres:
@micahgodbolt could you kindly please join the WICG (button on the right-hand side):
https://www.w3.org/community/wicg/

If you've already joined, could you please link your GitHub account to your w3c account:
https://www.w3.org/users/myprofile/connectedaccounts

That should resolve the IPR issues.

angallison:
Will this appear before the "Implementation" section?

research/src/pages/button.mdx
```diff
@@ -16,6 +16,20 @@ import ComponentCoverage from '../components/component-coverage'
 
 <Anatomy component="Button" />
 
+## Accessibility
+
+- Use a touch target area of `44px` by `44px`
+- For multiple buttons provided on the page that perform the same function, one of the targets should be `44px` by `44px`.
+- Buttons should get activated using `Enter` key as well as `SPACE` key.
```
chrisdholt: @micahgodbolt since semantic elements (button and input with type=button) already handle this behavior, this seems more specific to custom implementations (the bullet below). What are your thoughts to moving this as a bullet under "custom" below?

research/src/pages/button.mdx
```diff
@@ -16,6 +16,20 @@ import ComponentCoverage from '../components/component-coverage'
...
+- For custom buttons made using `<div>`, `<span>`, `<svg>`or `<img>`,
+  - `role="button"` must be applied.
+  - Tabindex property should be set to "0", so that it could get tab focus.
```
chrisdholt: nit: should this be `tabindex` since it's an attribute (similar to others)?

research/src/pages/button.mdx
micahgodbolt: good call.

micahgodbolt:
@angallison i currently have it placed between anatomy and concepts. 

image

micahgodbolt:
@marcoscaceres connected

research/src/pages/button.mdx
```diff
@@ -16,6 +16,20 @@ import ComponentCoverage from '../components/component-coverage'
 
 <Anatomy component="Button" />
 
+## Accessibility
+
+- Use a touch target area of `44px` by `44px`
```
dzearing: This one needs to be refined, questioned, or removed.

Nearly every button on the web is smaller than 44x44. Where did this guidance come from? Is it correct?

For example, github buttons are 34px tall.

Fabric buttons I believe are 32.

research/src/pages/button.mdx
```diff
@@ -16,6 +16,20 @@ import ComponentCoverage from '../components/component-coverage'
...
+- For multiple buttons provided on the page that perform the same function, one of the targets should be `44px` by `44px`.
+- For custom buttons made using `<div>`, `<span>`, `<svg>`or `<img>`,
+  - `role="button"` must be applied.
+  - `tabindex` property should be set to "0", so that it could get tab focus.
+  - Custom buttons should be activated by both `Enter` and `SPACE` key.
```
dzearing: It probably needs to be clear on what activation means; on keydown or keyup or keypress? I believe keydown.

chrisdholt: 👍 - technically *all* buttons should have this behavior. I think the important call out here is that custom buttons need them defined programmatically (`onKeydown`).

dzearing: One other minor note that I think would be good to call out here; if a keydown is handled, `preventDefault` should be called to prevent default browser behaviors. Otherwise, space might end up scrolling the browser window. I don't think `stopPropagation` should be called though.

I'm thinking that native buttons already do this, so this is more about custom buttons

research/src/pages/button.mdx
levithomason: Also, since these are the "Open UI" specs, we need to document what is existent in the broader open community.  This will have to be validated with some sources.  The specs here should descriptively document what exists in the community already opposed to prescribe something.

Think of Wikipedia, which doesn't allow original content but instead links to sources.  In a similar way, phase 1 of Open UI needs to link to sources and aggregate them.  We should not prescribe things we want.

If Microsoft has a certain accessibility requirement then it should live here but we need to introduce a way of noting that this is Microsoft's requirement.  Think of how an MDN page shows APIs and properties that exist in HTML/CSS/JS specs, but also allows vendor specific APIs and properties to be surfaced.  However, it is clearly noted that these are vendor specific variants of a thing, opposed to presenting it as if it were the spec itself.

This capability of the Open UI doc build does not yet exist.  So, we need to figure how we want to display this kind of information, accessibility information that comes from specific sources.

research/src/pages/button.mdx
levithomason: See https://github.com/WICG/open-ui/pull/14/files#r338290817

gregwhitworth:
@micahgodbolt I haven't seen any movement on this since 2019, and I'd expect the template to include accessibility information. Our <select> has this and this will be required for any component/control that is within Open UI. Let me know how and if you plan to get this into place or if I should close this down as dupe to the template issue.

gregwhitworth:
This was added in the spec template. Closing this.

PR #17, opened 2019-11-05
Adding spec template.

dzearing:
Taking a stab at adding this. I'm curious how we should curate opinionated specs and what makes sense to include here vs in a separate implementation repo.

### DIFF
--- a/research/src/pages/component-spec-template.mdx
+++ b/research/src/pages/component-spec-template.mdx
@@ -11,7 +11,33 @@
 
 ## Visual examples
 
-> Provide, if possible, visual examples of the component being built.
+> Provide, if possible, visual examples of the component being built, to provide an idea of implementations.
+
+## DOM hierarchy/structure
+
+> Define the recommended DOM shape to represent the component, and how important attributes (aria attributes especially) are applied to the various parts. In cases where a component nests other components, ideally, you expand the full dom structure to understand both the expectation and if there are shortcommings in the child components' customizability.
+>
+> Example:
+>
+> ```html
+> <div class="root">
+>   <div class="rail"></div>
+>
+>   <div class="mark">
+>     <label class="markLabel" />
+>   </div>
+>   <div class="track" />
+>   <div
+>     class="thumb"
+>     tabindex="0"
+>     role="slider"
+>     aria-valuenow="0"
+>     aria-valuemin="0"
+>     aria-valuemax="10"
+>   />
+>   <input name="{name}" type="hidden" value="0" />
+> </div>
+> ```
 
 ## Component anatomy
 
@@ -39,32 +65,6 @@
 > | `className`    | `string`                                     | Allows the author to add a class name to the root element.                                 |
 > | `defaultValue` | `number` \| `number[]`                       | Provides the uncontrolled default value of the component, mutually exclusive with `value`. |
 > | `disabled`     | `boolean`                                    | Specifies whether a component should be disabled.                                          |  |
-
-## DOM hierarchy/structure
-
-> Define the recommended DOM shape to represent the component, and how important attributes (aria attributes especially) are applied to the various parts. In cases where a component nests other components, ideally, you expand the full dom structure to understand both the expectation and if there are shortcommings in the child components' customizability.
->
-> Example:
->
-> ```html
-> <div class="root">
->   <div class="rail"></div>
->
->   <div class="mark">
->     <label class="markLabel" />
->   </div>
->   <div class="track" />
->   <div
->     class="thumb"
->     tabindex="0"
->     role="{ComponentName}"
->     aria-valuenow="0"
->     aria-valuemin="0"
->     aria-valuemax="10"
->   />
->   <input name="{name}" type="hidden" value="0" />
-> </div>
-> ```
 
 ## Behaviors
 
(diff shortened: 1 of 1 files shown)


research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+> Components should be built to withstand multiple aesthetic looks. This ensures customizability and reuse in the long term.
+>
+> At minimum, each component should consider the following theming requirements: light mode, dark mode, high contrast light and high contrast dark.
+>
+> If applicable, include some screenshots of themed variants of the component to get an idea of what is to be achieved through customization.
+
+### Class names
+
+> Components should include a class name contract to achieve maximum reusability. The contract should include:
```
chrisdholt: Is there a correlation here to the DOM structure as well? Not sure how we tie the two together but these seem linked if the goal is reusability.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,233 @@
...
+> | `rail`         | Rail slot class.                        |
+> | `thumb`        | Thumb slot class.                       |
+> | `rootDisabled` | Applied when the component is disabled. |
+> | `rootSelected` | Applied when the component is selected. |
+
+### Component design tokens
+
+> Assuming the Fluent UI styling system, a component's styling may reference design tokens.
+>
+> Tokens represent the general look and feel of the various visual slots. Tokens feed values into the styling during applicable conditions. For example, `railBackgroundHovered` would apply to the `rail` background when the mouse cursor is hovered over the component.
```
chrisdholt: Since this is being set as a template, does it make sense to start a conversation on terminology? For instance, what do we mean when we say `slot` - are we following the Web Component pattern here or is it slightly different? May not be the time, but figured I would float it.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,233 @@
...
+> - Are states changed as a result of a keystroke?
+> - Are some events triggered on key down vs key press?
+> - RTL settings?
+>
+> Example:
+>
+> | Key       | Description                                                                                                                                                    |
+> | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+> | Up/Right  | Increments the value of the {ComponentName} by the amount specified by the `step` prop. If the `shift` modifier is pressed, increases by 10x the `step` value. |
+> | Down/Left | Decrements the value of the {ComponentName} by the amount specified by the `step` prop. If the `shift` modifier is pressed, increases by 10x the `step` value. |
```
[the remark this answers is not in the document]
dzearing: nice

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+## Visual examples
+
+> Provide, if possible, visual examples of the component being built.
+
+## Component anatomy
+
+> Outline the parts of a component; what their names and purposes are. Ideally order them in DOM order. Also consider if the component should be broken into subcomponents. Example:
+>
+> | Name        | Considerations                                                       |
```
dzearing: Move anatomy after DOM

research/src/pages/component-spec-template.mdx
dzearing: DOM consists of elements which would be 1 to 1 with slot names. A slot would be 1 to 1 with class name, so yes there is a correlation. One slot = one class.

Additionally for stateful things, we would keep state based classes attached to the root of a component, so that selectors can be used to adjust child content based on component state.

Does that seem reasonable?

research/src/pages/component-spec-template.mdx
dzearing: A `slot` is a named part of a component representing a placeholder location to replace with content, but usually having a default element type. It is typically represented by a single DOM element.

research/src/pages/component-spec-template.mdx
dzearing: Class names may be too much about implementation detail in terms of the styling system.

research/src/pages/component-spec-template.mdx
dzearing: Should we add terminology in the template? Or somewhere else?

I could certainly take a stab at it. I could open a bug, and get this checked in and follow up in a separate pr

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,153 @@
...
+> - should be focusable via `tabindex=0`
+> - role set to `{ComponentName}`
+> - receives `aria-valuemin` and `aria-valuemax` representing min/max
+> - receives `aria-valuenow` representing the current value
+
+### Accessibility concerns for the user
+
+> Consider which props, mapped to attribute values, must be provided for the user to make the component accessible. Often props which require localization or calculating a value must be provided by the user.
+>
+> For React props, consider using `ariaLabel` and `ariaLabelledBy`, rather than the natural kebab-cased variants (`aria-label`) so that they can be provided without destroying type safety.
```
chrisdholt: @dzearing I'm torn on this. In our implementation we extend existing type interfaces for DOM elements and still get type safety. Is the important thing here to ensure that these can be enumerated, or that the implementation is 1:1? I would think that the goal is less the details of the interface (how we accomplish passing an aria-label) and more as to whether it is *required* or *beneficial* for a given implementation.

research/src/pages/component-spec-template.mdx
chrisdholt: For sure get it here; I'm open, more asking the question. I'm not sure styles/classNames are reusable if the structure isn't 1:1, but that's also more of something we've run into specific to our implementation. I'm good either way.

research/src/pages/component-spec-template.mdx
chrisdholt: I'm good with either - I don't think it needs to block this, but we should definitely have the conversation as I think semantics will play a big part of what we're trying to accomplish here.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,153 @@
...
+### Touch interaction
+
+> Define the touch behaviors of the component.
+>
+> - Are there different behaviors than in mouse scenarios?
+> - Are touch targets large enough for fat fingers? (44x44px is the recommended minimum size.) For example, the `thumb` of a slider might appear to be small, but the touchable area should be larger.
+
+### Screen reader accessibility
+
+> Review/link aria specifications and list which aria labels should appear on which parts of the component.
```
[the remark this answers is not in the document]
gregwhitworth: I agree with @nicholasrice here - I'd really like to ensure that the behaviors are intertwined with necessary changes if these invoke necessary AT adjustments.

research/src/pages/component-spec-template.mdx
gregwhitworth: Also, should we be getting into any react specifics? For example, shouldn't we just use the standardized variant if it is a standardized attribute and the implementor can map this to their framework/lib or web component.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,153 @@
...
+> - receives `aria-valuemin` and `aria-valuemax` representing min/max
+> - receives `aria-valuenow` representing the current value
+
+### Accessibility concerns for the user
+
+> Consider which props, mapped to attribute values, must be provided for the user to make the component accessible. Often props which require localization or calculating a value must be provided by the user.
+>
+> For React props, consider using `ariaLabel` and `ariaLabelledBy`, rather than the natural kebab-cased variants (`aria-label`) so that they can be provided without destroying type safety.
+
+## Implementation research
```
gregwhitworth: Shouldn't this be on the research page? My expectation is that this should follow a similar standards model in that the colated document represents content that has been agreed upon. We should have a draft status as well so folks know whether it has been agreed upon or not. For example, I think each item in each section should be discussed and agreed upon in a Github issue/PR.

I did ask for a resources section that can contain a usecase doc, other research that doesn't fit into the spec but I think this should be concise and not a complete dump in spec of research that was done.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,153 @@
...
+> - `mouseup` should remove the `mousemove` event.
+
+### Touch interaction
+
+> Define the touch behaviors of the component.
+>
+> - Are there different behaviors than in mouse scenarios?
+> - Are touch targets large enough for fat fingers? (44x44px is the recommended minimum size.) For example, the `thumb` of a slider might appear to be small, but the touchable area should be larger.
+
+### Screen reader accessibility
```
gregwhitworth: I understand this starts getting into some design, but I think we should possibly have a default recommendation for system colors of the anatomy for `forced-color` support. I think this deserves its own issue to discuss so don't block the PR on this.

research/src/pages/component-spec-template.mdx
chrisdholt: Yeah, I think we should try as hard as possible to stay implementation agnostic - aligning to web standards will help with that and allow us to scale as the next new, cool thing comes out.

dzearing: good calls

research/src/pages/component-spec-template.mdx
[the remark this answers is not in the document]
dzearing: oops! nice call

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,153 @@
...
+>     aria-valuemin="0"
+>     aria-valuemax="10"
+>   />
+>   <input name="{name}" type="hidden" value="0" />
+> </div>
+> ```
+
+## Component anatomy
+
+> Outline the parts of a component; what their names and purposes are. Ideally order them in DOM order. Also consider if the component should be broken into subcomponents. Example:
```
wdencker: In addition to the part definitions, perhaps the template should include a diagram of the anatomy as well? (This could also be categorized under the DOM hierarchy / structure section above). Currently, the WIP specs for <select> and <carousel> both include a diagram.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,153 @@
...
+> | Name        | Considerations                                                       |
+> | ----------- | -------------------------------------------------------------------- |
+> | `root`      | The root element.                                                    |
+> | `rail`      | The line behind the Slider thumb and rail.                           |
+> | `mark`      | Optional mark line on top of the rail, below the track.              |
+> | `markLabel` | A label aligned with a mark to render the textual value of the mark. |
+> | `track`     | The selected area of the Slider.                                     |
+> | `thumb`     | The focusable draggable knob within the Slider.                      |
+> | `tooltip`   | The tooltip rendered above a thumb.                                  |
+
```
wdencker: Would add a subsection here for events at the anatomy level (which are distinct from the input-specific events like onClick or onKeyDown, already included below).

For example, in the WIP `<select>` spec, one event is "`::host` Listening for `toggle-pop-up` event to hide/show `::part-pop-up`."

research/src/pages/component-spec-template.mdx
levithomason: I second the diagram.  Perhaps we can do this in another follow PR so as to get this one merged.

Here's a sample diagram @gregwhitworth shared with me previously that another team/effort was working with:

image

And we also have a `research/carousel` branch here in Open UI that is toying with a CSS Grid approach so that we can write these in our MDX files:

image

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,153 @@
...
+> Outline the recommended props for the component; taking into consideration conformance guidelines. Make sure to discuss differences and rationalizations.
+>
+> Example:
+>
+> | Name           | Type (default value if clarification needed) | Purpose                                                                                    |
+> | -------------- | -------------------------------------------- | ------------------------------------------------------------------------------------------ |
+> | `as`           | `string` (`"div"`)                           | Replaces the primitive root element type.                                                  |
+> | `className`    | `string`                                     | Allows the author to add a class name to the root element.                                 |
+> | `defaultValue` | `number` \| `number[]`                       | Provides the uncontrolled default value of the component, mutually exclusive with `value`. |
```
levithomason: Open UI specs are framework agnostic so we should avoid React-isms.  That would mean "props" would probably just be "attributes" and the camelCased React-isms would not be present.

research/src/pages/component-spec-template.mdx
levithomason: Yep, agreed.  The final spec should just be a statement of what it is.  All the supporting info and research would be on the "Research" page.

research/src/pages/component-spec-template.mdx
gregwhitworth: We copied the premise of the Carousel variant for the <select> one - it looks like this:

image

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,153 @@
...
+> `thumb`:
+>
+> - should be focusable via `tabindex=0`
+> - role set to `{ComponentName}`
+> - receives `aria-valuemin` and `aria-valuemax` representing min/max
+> - receives `aria-valuenow` representing the current value
+
+### Accessibility concerns for the user
+
+> Consider which props, mapped to attribute values, must be provided for the user to make the component accessible. Often props which require localization or calculating a value must be provided by the user.
```
chrisdholt:
```suggestion
> Consider which attributes must be provided for the user to make the component accessible.
```

dzearing:
Coordinating with @EisenbergEffect, he will post another PR to supersede this one.

PR #19, opened 2019-11-06
Add initial curated page and research for <select>

wdencker:


research/src/components/select-anatomy-styles.css
```diff
@@ -0,0 +1,68 @@
+/* General */
```
gregwhitworth: rename this to select-anatomy.css (similar to your .js)

gregwhitworth:
There are several sections that are missing that we had discussed, granted I know accessibility you had a question or two on, but I'd like to see the following:

## Accessibility

### Keyboard
* Is it focusable by default (in the case of select, yes)?
* How many tab stops does it have?
* What is the default behavior when using the arrows - I'd like to denote that pressing the down key traverses to the next <option> in the list and the inverse for the up key.
* We should also ensure that all attributes that the `<select>` [supports](https://html.spec.whatwg.org/multipage/form-elements.html#the-select-element) are listed here as well.
* Address the popup dismissal (eg: it's the escape key)

### Assistive technology
* Place a link to the HTML AAM [spec](https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings)
* And the combox [role](https://www.w3.org/TR/core-aam-1.1/#role-map-combobox)

### Forced colors
Put this in as a placeholder, ultimately once we have the parts defined we should be able to provide system color keywords relevant for when forced-colors is active.

### Parts
We need to ensure that each part is defined, currently the anatomy has a ::host, a pop-up and a button. We should split all of these out so that we can propose them and discuss them on their own([here's an example](https://drafts.csswg.org/css-pseudo-4/#selectordef-spelling-error)). We don't need to be in depth but this may be a good way to denote some of the characteristics (eg: does it get a tabstop? Just thinking out loud though - this may not be useful).

### Resources
Put in here any resources that you think are valuable for the reader of the document

wdencker:
I've added placeholders for now -- will start working on curating the additional content.

wdencker:
@gregwhitworth I added content for the sections you mentioned -- let me know what revisions you would suggest

research/src/pages/select.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+- `::part-button` Dispatches custom event `split-button-clicked`
+
+## Accessibility
+
+### Keyboard
+
+Keyboard users can operate the `<select>` as follows:
+
+- Focus on the `<select>` element using the `Tab` key (there is a tab stop on the button). The control can be
```
gregwhitworth: this shouldn't be 'tab stop on the button' but should be `part-button`.

research/src/pages/select.mdx
```diff
@@ -0,0 +1,132 @@
...
+Implements the [combobox role](https://www.w3.org/TR/core-aam-1.1/#role-map-combobox) per the
+[HTML AAM spec](https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings).
+
+### Forced Colors
+
+Placeholder for now, dependent on standardization on parts definitions.
+
+## States / Properties
+
+### &ltselect&gt
```
gregwhitworth: for some reason markdown makes headers render the &lt; and doesn't convert them - so for headers use <select>

research/src/pages/select.mdx
```diff
@@ -0,0 +1,132 @@
...
+  focused by default by setting the `autofocus` attribute.
+- Having focus, but with the pop-up still closed, users can:
+  - Change the selection to the next option using the `Down` arrow
+  - Change the selection to the previous option using the `Up` arrow
+  - Change the selection to the first option starting with a certain sequence by typing that sequence
+  - Cycle the selection through options starting with a certain letter by continually typing that letter
+- Having focus, the user can open the popup using the `Enter` key or the `Space` key. The popup can be dismissed using
+  the `Enter` key or the `Esc` key.
+- Having focus and with the pop-up open, the user can select options in the same manner as above. As the user traverses
+  the options, the currently-selected option is always visible and highlighted, with the menu scrolling as necessary.
```
gregwhitworth: We should also denote that the `value` field is updated upon selection.

research/src/pages/select.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+### Assistive Technology
+
+Implements the [combobox role](https://www.w3.org/TR/core-aam-1.1/#role-map-combobox) per the
```
gregwhitworth: This is great, thanks for adding this in!!!

gregwhitworth:
@dzearing can you take a look at this PR. Let's try to not dig into the naming or behaviors themselves. I'd like to open individual issues for those; but the general structure and any glaring gaps. Thanks.

cc: @levithomason

gregwhitworth:
@levithomason can you take a look at this. This has the anatomy and other items I know we need to implement the new select. Also, there are things we know we want to change but I'd really like to have those discussions in issues to begin the flywheel of how these things will be handled and suss out the process. EG: I want to adjust the anatomy some but I want to have an issue for that. I want to bikeshed names but want that in issues - I don't want all of this buried in PRs. I'll be merging this in by EOD tomorrow as I plan to show this in a local huddle. Also I have had meetings with Apple and have one with Moz and I want a strawman proposal to share with these teams as we've begun to gain momentum. cc: @dzearing

gregwhitworth:
pinging @chrisdholt too as I'm simply looking for anyone on the design system side to weigh in if we're missing something egregious. I expect us to continue to evolve this which is why we added the draft status.

research/src/components/select-anatomy.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import './select-anatomy.css'
+
+const SelectAnatomy = () => {
+  return (
+    <host>
+      <part name="button">
+        <slot>Default = Split Button Component</slot>
```
chrisdholt: Split Button or Select?

research/src/pages/select.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+- `::host` Listening for `toggle-pop-up` event to hide/show `::part-pop-up`
+
+- `::part-button` Listens to `click` event and dispatches `toggle-pop-up` event
+
+- `::part-button` Listening for `option-change` event to update selected value shown
+
+- `::part-pop-up` Listening for `option-change` event to dispatch `toggle-pop-up` event
+
+- `::part-button` Dispatches custom event `split-button-clicked`
```
chrisdholt: Split button or select?

chrisdholt: This could be confusion on my part; I wouldn't equate the button here to a split button as it's a single action with the sole intent of toggling the pop-up (whether on keydown of enter/space or click)

chrisdholt:
> pinging @chrisdholt too as I'm simply looking for anyone on the design system side to weigh in if we're missing something egregious. I expect us to continue to evolve this which is why we added the draft status.

Thanks @gregwhitworth - I don't see anything egregious. I think one scenario I've seen sometimes is a desire to customize the actual carat for the dropdown itself. While I don't see it called out, I'm assuming that can be supported through a pseudo element on the button perhaps or another part. Only a couple questions on the use of `split button` from a semantic standpoint. Looks like either copy/paste or a misunderstanding of terms on my end.

research/src/components/select-anatomy.js
gregwhitworth: This is actually something that I want to cover in an issue. This is a slot so technically the content is replaceable. I initially had it as a split button and I already plan to have it as a <button> or we'll need to create new element that has the psuedo elements if the desire is truly there to have the current button design as seperate parts as you're correct that a button and split button are very different paradigms.

research/src/pages/select.mdx
gregwhitworth: agreed, as noted above I'd expect this to be a button or a new element/sub pieces that still live within a slot that allow the content to be replaced (eg: be able to build a combo box from the select)

gregwhitworth:
Thanks Chris, I'll go ahead and merge it in as I noted I want the default discussion to occur in issues.

PR #31, opened 2019-11-22
Add the Primer Design System 

emplums:
👋 Hey there!

Adding GitHub's design system, Primer. Just adding the buttons for now. :)


gregwhitworth:
Just in case notifications aren't on - @levithomason can you review this please. Thanks.

research/src/sources/primer.json5
```diff
@@ -15,69 +15,84 @@
...
         {
           "name": "block",
-          "category": "type",
           "image": "primer-block-button.png",
-          "$comment": "Full width button"
+          "$comment": "Full width button."
+        },
+        {
+          "name": "octicon",
+          "openUIName": "icon",
```
levithomason: The `name` key will always capture the original named in the source material.  In this case, the design system called this an "Octicon Button". 

We are playing with this "openUIName" idea right now when we need to specify a name that a design system uses, such as `octicon button`, and also surface the variant along with other commonly named things, like `icon button`.

The automated page layout will show the `octicon` button in the `icon` concept now, but we still have a record that the original `name` was something else.  We'll likely show this differing information in the automated page soon.

research/src/sources/primer.json5
```diff
@@ -15,69 +15,84 @@
...
         },
         {
           "name": "outline",
-          "category": "type",
           "image": "primer-outline-button.png",
           "$comment": "Outline buttons downplay an action as they appear like boxy links."
         },
         {
           "name": "link",
-          "category": "type",
```
levithomason: If a design system has a variant value that is part of a category of values, like `type="primary"`, then we'll record the variant name as `primary` and the category as `type`.  More on this and why here: https://open-ui.org/contribute/design-system.

levithomason:
😺!

I've added a few more missing buttons and make a couple minor cleanup items.  See https://github.com/WICG/open-ui/pull/31/commits/e6f9a6927c1e2f8c84f71e6831e7791235129c30.

levithomason:
@gregwhitworth it seems we're running into more IPR issues 😿 Anything we can do about that?  It will be a huge blocker here.  Since we are not writing specs, but design research in a community group, it seems this requirement doesn't actually apply to Open UI research. The error reads:

>...Some of the contributors in this pull request were not recognized as having made the required IPR commitment to make **substantive changes to the specification** in this repository.

Because of this I'm marking it non-substantiative until we can get it resolved. Please correct me if I'm wrong.

levithomason:
Marked as non substantive for IPR from ash-nazg.

gregwhitworth:
@levithomason they have to join the WICG group to pass the IPR, @emplums did you do this?

gregwhitworth:
Here is the link to join the WICG: https://www.w3.org/community/wicg/ I'll go add it to the contribution doc

PR #32, opened 2019-11-22
Fix Ant Design tooltip source link

coreybruyere:


levithomason:
Duplicate #39, thanks!

PR #37, opened 2019-12-12
Adding Table component research

pompomon:
PR contains research of Table component and describes various concepts of Table component 
This includes anatomy, images/animations of interactions, as well as research notes on design systems reviewed, including links to code sandboxes for future references


research/src/pages/table.research.mdx
```diff
@@ -0,0 +1,19 @@
+---
+name: Table
```
nicholasrice: I noticed that Select is tagged as a working draft - should this be of the same status? Do we have a formalized process or enumerated criteria to inform what *state* this should be in?

research/src/pages/table.mdx
```diff
@@ -0,0 +1,656 @@
...
+Data grid -- tabular information that has column titles, row titles, or both and enables users to navigate interactive elements it contains using directional navigation keys. The grid pattern is particularly useful if the tabular information is editable or interactive (per [WAI-ARIA](https://www.w3.org/TR/wai-aria-practices/#grid))
+
+## Anatomy
+
+### Diagram
+
+<TableAnatomy />
+
+### Part Definitions
+
```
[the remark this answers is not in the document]
pompomon: Good point, I'll do kebab-case as other components :)

research/src/pages/table.research.mdx
pompomon: I'm not aware of the formalized process, but it's a good point, I'll mark Table as working draft as well

levithomason:
Hi @pompomon, thanks for the great work and apologies for the crazy delay!  I've done some cleanup here and review this with @nicholasrice and we can merge now.

We'll tackle some of these issues in more detail later, but here's the gist of the changes:

- Normalized the anatomy to work with Select anatomy a bit (we should have a formal way of doing this in the future)
- Remove code snippets from proposal. In general, we should link out to material instead of copying it over. This way it stays fresh at the source and is also clear that it is not our content, but theirs.

PR #44, opened 2020-02-25
Add design token page

nicholasrice:
Adds a Design Token page to address alignment of design token names, values, and uses across UI components. This was identified by @levithomason and myself as an alignment point for UI components that OpenUI could help standardize.

Looking forward to thoughts and edits.


gregwhitworth:
What's the status of this? Also, @levithomason it would be good to understand how this interacts with the design token CG, if at all?

gregwhitworth:
Based on the lack of response I'm going to close this PR. @nicholasrice please feel free to re-open if things change.

levithomason:
@gregwhitworth Whoops, yea this is still ongoing. It will be long running.

@nicholasrice Per our sync today, we've agreed to not prefix token names by default.  In order to handle inevitable collisions, we'll ensure systems have the option to prefix token names. This will also allow interop and embed scenarios to play nicely together.

PR #45, opened 2020-02-26
Attempt to improve alt text for images

### LINKED ISSUE #40: Images need useful text descriptions
None of the images on pages like the one for [button](https://open-ui.org/components/button) have useful text descriptions. They use the image file name instead. For example:
```
<img alt="andt-carousel-dotposition.png" ...>
```

hiwelo:
Hello folks :) 

Following the issue #40 opened by @LJWatson regarding the lack of usefulness of the `alt` present on images in pages like [button](https://open-ui.org/components/button), I made this PR as an attempt to fix this.

From what I saw, the image name is following a structure like `system-component-variation.extension`. My idea was to simply transform this kebab case name to a real sentence.
For example, this PR is transforming the `carbon-button-primary` alt attribute to be `Carbon button primary`. 

In my opinion, this script is not replacing what could be offered by a having a real alt description for each image but it might provide a better experience for screen readers based on the current information available.

If you're happy with this solution, I was wondering if this script should be part of the `Specimens` component (as proposed by this pull request) or if it should be part of the `image` component, by replacing the current `alt = src` default value attribution.

What do you think about it?

--- 

Could close #40

### DIFF
--- a/research/src/components/specimens.js
+++ b/research/src/components/specimens.js
@@ -1,5 +1,6 @@
 import _ from 'lodash'
 import React from 'react'
+import { kebabCaseToSentence, removeExtensionFromFilename } from '../utils/text'
 import { getImagesForComponentConcept } from '../sources'
 import Image from './image'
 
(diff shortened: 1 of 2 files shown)


LJWatson:
I agree they're still not proper text descriptions, but they are a definite improvement over what's there at the moment. Thanks.

levithomason:
Thanks much for assisting here!  I've take this a bit further and included the design system data with the function that returns the image specimens.  We now generate alt tags like so:

```html
<img alt="An image of the disabled concept on a Dropdown component in Ant Design." >
```

I'll close this as a dupe now.  Please add any comments to #52.

PR #49, opened 2020-03-06
Link component source page to names

### LINKED ISSUE #43: Names section of component pages should be anchors
Under the "Names" section of of each component page (eg. https://open-ui.org/components/button) we list the name of the Design System but we don't link to the resource. I believe we *have* this information because each component entry in a design system is required to have a URL. 

I think we should add anchors to the design system names, linking them to the information resource. Directly linking to the resource brings more transparency about where the information is coming from and allows site users and authors to investigate the design system further if desired.

levithomason:
Fixes #43

### DIFF
--- a/research/src/components/component-coverage.js
+++ b/research/src/components/component-coverage.js
@@ -17,13 +17,26 @@ const ComponentCoverage = ({ component }) => {
       <ul style={{ marginLeft: 0 }}>
         {_.map(withDifferentNamesUniq, (component, i) => {
           const components = withDifferentNamesGrouped[component.name]
-          const sourceNames = _.map(components, 'sourceName').join(', ')
 
           return (
             <li key={component.name} style={{ listStyleType: 'none' }}>
               {component.name}
               {' - '}
-              <small style={{ opacity: 0.75 }}>{sourceNames}</small>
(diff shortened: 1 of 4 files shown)


research/src/utils/typography.js
```diff
@@ -43,6 +43,10 @@ const typography = new Typography({
         borderLeft: '4px solid #ccc',
       },
 
+      li: {
+        marginBottom: rhythm(0.25),
+      },
```
levithomason: Make lists less ugly

research/src/components/component-coverage.js
```diff
@@ -17,13 +17,26 @@ const ComponentCoverage = ({ component }) => {
...
+                {components.map((component, i, arr) => (
+                  <>
+                    <a
+                      key={component.sourceName}
+                      target="_blank"
+                      rel="noopener noreferrer"
+                      href={component.url}
+                    >
+                      {component.sourceName}
+                    </a>
```
levithomason: Here's the actual change, we now link the design system for component names back to the design system page for that component:

image

research/src/utils/typography.js
chrisdholt: this would make a really good commit message :)

Issue #18, opened 2019-11-05
Investigate modal/dialogue element


gregwhitworth:
There is an increasing desire/demand for a <dialogue> or <modal> component in the web platform. There is also an [explainer ](https://github.com/adrianhopebailie/modal-window/blob/master/explainer.md)that is geared to addressing needs presented by the Payments WG for other items using a custom <dialogue>. We should ensure that needs outlined by the various design systems is also covered by this solution. The cross origin needs outlined there are real but we should figure out a way to converge the needs so that webdevs aren't having to create a modal/dialogue <N> different ways with different capabilities of styling, extensibility, etc.

chrisdholt:
@gregwhitworth I'd like to take a shot at writing a spec for this. I'm working on a similar issue on our end. Going to self assign.

gregwhitworth:
@chrisdholt is there any plan to keep moving forward with this. cc @melanierichards as I know she's been looking at things like this

chrisdholt:
@gregwhitworth I'll pull up the branch and see if I can get something up in the next week or so. I think I primarily need to update to align with the latest spec template and update a reference for light dismiss to the issue created the other week.

melanierichards:
Happy to coordinate with you on this, @chrisdholt! The work I'm tracking is popups all-up, which includes dialog: #116 Currently planning on a mega-page for popups, and we can discuss as a group how to break that up if needed. AFAICT that's the first instance of nested concepts.

chrisdholt:
@melanierichards would be great to sync on this! Do you want to setup some time next week to walk through what I have currently and we can refine and get a PR out?

gregwhitworth:
@melanierichards @chrisdholt ping

keithamus:
`<dialog>` now exists, and it has a `showModal()` API which shows the given dialog as modal, in the top layer. Congrats everyone!

PR #51, opened 2020-03-11
doc(component-spec-template): add initial template for component specs

EisenbergEffect:
This pull request adds a spec template for components. It supersedes PR #17 by taking into account the template from that PR, addressing most if not all feedback from that PR, and merging it with the preliminary template that fast-dna has been successfully using to spec out its web components.


EisenbergEffect:
/cc @dzearing @chrisdholt for visibility.

Note: The above failed ipr check will need to be rerun. I didn't have my GitHub account associated with w3 until after I submitted this PR.

chrisdholt:
@levithomason I'm having trouble re-running IPR validation. Can you try and run it? 

@EisenbergEffect this looks good to me, though as it's based on our template I am a bit biased. Thanks for submitting!

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
+- *A designer customizing the component.*
+
+### API
+
+*Outline the key elements of the component's public API surface, taking into consideration conformance guidelines. Make sure to discuss differences and rationalizations. Consider high and low-level APIs. Attempt to design a powerful and extensible low-level API with a high-level API for developer/designer ergonomics and simplicity.*
+
+#### Tag Name
+
+*The tag that would be used in HTML. e.g. `<my-slider></my-slider>`*
```
chrisdholt: @EisenbergEffect - I'm wondering if we need this for the initial specification. This seems more along the lines a team implementing this spec would care about, whereas the intent here is to document the overall pattern for alignment. I think we can probably remove this as it's going to be specific to the implementation (as WC's require).

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+*What is the plan for testing the component?*
+
+### Tooling
+
+*Are there any special considerations for tooling? Will tooling changes need to be made? Is there a special way to light up this component in dev tools that would be compelling for developers/designers?*
+
+### Documentation
+
+*What additions or changes are needed for user documentation and demos? Are there any architectural/engineering docs we should create as well, perhaps due to some interesting technical challenge or design decisions related to this component?*
```
chrisdholt: Given the current intent (documentation), I think this could likely be removed. Although, the description of "documentation" does cause me to wonder if there's an opportunity for a specific section on things that would be required from the web platform.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+*Will implementing the component require taking on any dependencies?*
+
+- *3rd party libraries*
+- *Upcoming standards we need to polyfill*
+- *Do any dependencies bring along an associated timeline?*
+
+### Test Plan
+
+*What is the plan for testing the component?*
```
dzearing: Would be good to include an example table to guide spec writers what to write here. The section may be redundant with some of the previous sections, so I'm not totally sure what to include.

| Scenario | Expectation |
|-|-|
| User clicks on enabled Checkbox with mouse | Checked state is toggled |
| User clicks on disabled Checkbox with mouse | Interaction is ignored, (is focus set to disabled checkbox?) |

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+- *Swapping of internal icons/visuals*
+- *Localization*
+
+### Security
+
+*Are there any security implications surrounding the component?*
+
+### Performance
+
+*Are there any performance pitfalls or challenges with implementing the component?*
```
dzearing: Might be good to provide some examples of common pitfalls:

* If the component renders a loop of items, should certain areas be considered for virtualization?
* Does the component need to measure container area before rendering? This could cause rendering performance or jank issues.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+*Example Table*
+
+| Key | Description |
+| --- | ----------- |
+| Up/Right  | Increments the value of the component by calling the `increment` method. If the `shift` modifier is pressed, increases by 10x the `step` value. |
+| Down/Left | Decrements the value of the component by calling the `decrement` method. If the `shift` modifier is pressed, decreases by 10x the `step` value. |
+
+#### Form Input
+
+*Does this component integrate with form submission, form validation, etc.?*
```
dzearing: This is a good question to include; maybe more questions or guidance on things to think about would be valuable if so:

* Does the component need to render a hidden form `input` element to ensure the value is read correctly into the form?

(What other things would a developer need to consider?)

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+*Example Table*
+
+| State Group | States | Initial State | Description | Interaction/Transition |
+| ----------- | ------ | ------------- | ----------- | ---------------------- |
+| `disabled` | `true`/`false` | `false` | When `true`, disables the control, preventing user interaction and displaying the control with a disabled appearance. | No interaction. Controlled programmatically. |
+
+### Accessibility
+
+*Consider the accessibility of the component.*
```
dzearing: This is very vague. Accessibility is a combo of: screen reader compatibility, keyboard model, contrast ratios, avoiding color-only to indicate some state (focus rectangles, for example).

Should this be specific to screen reader concerns and which aria attributes need to be distributed across which elements?

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
+> This document represents a component specification template. Its purpose
+> is to provide a point-in-time guide as to how a component should be
+> authored, in light of common anatomies, behaviors, and accessibility.
+> The primary audience is a developer authoring a given component.
+
+# {ComponentName} component specification
```
dzearing:
```suggestion
# {ComponentName} Component Specification
```

research/src/pages/component-spec-template.mdx
EisenbergEffect: I wonder if phrasing it that was doesn't get too far into the implementation. For example, the new ElementInternals APIs enable Web Component form participation without hidden inputs.

research/src/pages/component-spec-template.mdx
chrisdholt: I like the ambiguity here personally. By going broad here I think someone needs to really internalize and consider what all is necessary to include here. For instance, for something like a dialog (which I plan on using this template for) it may be enough to link to the W3C aria-practices documentation as a pattern already exists in standards form. For something like a split-button though, that doesn't exist and I would expect much more thought to go into defining how that might need to be built given the existing standards.

research/src/pages/component-spec-template.mdx
EisenbergEffect: Makes sense. I'll remove this section. A general name would be included in the spec title and overview, so I think that's sufficient for the spec's purpose.

research/src/pages/component-spec-template.mdx
EisenbergEffect: I'll update this to "Platform Requirements" with an appropriate description.

research/src/pages/component-spec-template.mdx
EisenbergEffect: I'm not entirely sure either. I think what I had in my head was are there any testing details that are non-standard; special test considerations. I think most of what to test would be covered in the above sections. I'll see if I can clarify some wording here.

research/src/pages/component-spec-template.mdx
EisenbergEffect: I added these examples with a slight rewording for flow.

research/src/pages/component-spec-template.mdx
EisenbergEffect: I've added some wording which I think is in the spirit of the initial idea: "*Does this component integrate with form submission, form validation, etc.? Can integration be accomplish with standard APIs or are special work-arounds required?*"

levithomason:
I think we should book a call on this PR as there is a lot to understand and discuss here.  The main concern I'd have is the amount of opinion and prescription it adds to Open UI in terms of how to implement a component.

To reference Open UI's stated goals:

>### Goals
>These are the goals Open UI believes will realize the above vision.
>
>- Document component names as they exist today
>- A common language for describing UIs and design systems
>- Eventual browser standards for web app components
>- Converging designer processes and developer workflows

These are currently and intentionally void of framework and implementation details.  The current mission is more about unifying and less about creating something new.  Think Wikipedia (no original sources) or the dictionary (linguistic descriptivism) opposed to a prescriptive specification.

That said, there's a lot in this template that I think does fit with the current goals and are areas that we're currently lacking in our component pages.  Specifically around use cases, non-goals, design, accessibility, keyboard nav, etc.  These are all things that can and should be studied in Open UI and are currently missing.

We're also up for amending and extending the goals to make sense.  Especially given that Open UI must have some level of "blessing" and guidance to leverage against the research in order to push the community toward the place we want to be.  I think we need to have a larger convo around that before merging as the PR in its current form implicitly changes the stated goals without due process.

research/src/pages/component-spec-template.mdx
EisenbergEffect: Fixed :)

EisenbergEffect:
@levithomason Sounds good to me. I'll finish making a few updates based on the feedback from @dzearing and @chrisdholt. After that we could get on a call and walk through the latest version together. Does that sound reasonable?

levithomason:
Definitely.  Also, please don't feel like this is a "stop ship."  I love the momentum and deep thought here and we'll certainly get much of this integrated right away.  I think there's room for the prescriptive stuff too, so we'll discuss how to fit that in.

research/src/pages/component-spec-template.mdx
EisenbergEffect: I was treating that as a broad overview statement with the sections below on keyboard nav, focus, forms, and AT tech being the areas where more details were fleshed out. We can add something maybe about how states would be represented in a color-independent way. I'll leave this section as is for now and we can discuss f2f as part of a broader review.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+| ---------- | ----------- | ------- | -------- | ----------- | ----------------- |
+| `change` | none | `true` | `true` | `false` | Fired when the slider's `value` changes. |
+
+### Appearance
+
+*Screenshots and/or description of the basic appearance of the component.*
+
+### Anatomy
+
+*Outline the component's structure with a diagram of its visual tree (shadow dom). Enumerate key areas of visual customization.*
```
nicholasrice:
```suggestion
*Outline the component's structure with a diagram of its visual tree. Enumerate key areas of visual customization.*
```

I think referencing "shadow dom" gets further into the implementation specifics than we want to get.

EisenbergEffect: I've reworded this sentence to remove that reference and be a bit more general.

research/src/pages/component-spec-template.mdx
dzearing: Yep I retract my statement - this is more of a group of accessibility concerns organized by the subtitles below :)

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+- *3rd party libraries*
+- *Upcoming standards we need to polyfill*
+- *Do any dependencies bring along an associated timeline?*
+
+#### Platform Requirements
+
+*Are there any core web platform features that are needed to implement this component well?*
+
+### Test Plan
```
gregwhitworth: I'd like more details on this as well because you're not going to be able to get into every JS manner/library in which to test these across JS libs and also not including the web platform testing collateral which uses a variety of methods both client, manual and C++ dumps.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+*Are there any additional details that need to be tested which aren't covered in the spec above? Are there any expected nuances or challenges with testing this component?*
+
+### Tooling
```
gregwhitworth: Again, not entirely sure this makes sense for the definition of the control/component as this vary drastically based on the libs, build steps, and what you consider "tools".

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+---
+
+## Resources
+
+*Any related resource links such as web standards, discussion threads, diagrams, etc.*
+
+---
+
+## Next Steps
```
gregwhitworth: I would expect these to be in Github issues. Possibly we should have a generic statement but link to the label for this specific specification so that the user can filter on enhancements, issues, etc. Something like:

> Open UI track all feature additions, modifications, bugs or even editorial changes in our Github repo. Click here to review this specification's issues.

research/src/pages/component-spec-template.mdx
gregwhitworth: Also, the a11y testing alone across ATs for a given component would be a few pages alone. I'd probably leave this up to the framework/UA to define.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+*A list of the key features unique to this component.*
+
+### Risks and Challenges
+
+*Notable risks or challenges associated with implementing or shipping the component.*
+
+### Prior Art/Examples
+
+*Screenshots and/or links to existing, canonical, or exemplary implementations of the component.*
```
levithomason: Here, we should just be able to link to the research page.  It has ample examples of names, anatomy, concepts, and links to the implementations.  We plan to continue extending this page to show more and more information as well.

levithomason:
In order to help get this to the next step, I propose we lock comments, address them, and merge this.  The next step is to see what a component or two actually look like here.  I have hard time seeing how some of these headings will be used specifically.  I also have a hunch an actual spec for a real component will be a better basis for continuing comments and iterating.

Thoughts?

chrisdholt:
> In order to help get this to the next step, I propose we stop comments, address them, and merge this. The next step is to see what a component or two actually look like here. I have hard time seeing how some of these headings will be used specifically. I also have a hunch an actual spec for a real component will be a better basis for continuing comments and iterating.
> 
> Thoughts?

If we get this merged in, I'd be happy to try my hand at the [dialog issue](#18) and push up a spec for that.

levithomason:
Sure, why don't @chrisdholt and/or @EisenbergEffect make a final pass here and merge as you see fit.  Let's stay flexible in mindset on it but move on to getting a spec or two up and we can see how it is working out.

EisenbergEffect:
I should be able to address the latest comments and get this merged next week.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+### Use Cases
+
+*Primary use cases for this component.*
+
+### Non-goals
+
+*A list of use cases, features, or functionality which are **not** goals for the component.*
+  
+### Features
```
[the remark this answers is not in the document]
EisenbergEffect: I see overview as a quick summary, along with background if there is any. Use cases provide the specific scenarios (I could rename that User Stories). Features, however, describes how all of that maps to specific capabilities of the component. A given user case could be accomplished through different conceptualization of features, for example.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+*A list of the key features unique to this component.*
+
+### Risks and Challenges
```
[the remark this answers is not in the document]
EisenbergEffect: I can certainly think of risks for various components. For example, if there's a component that relates to a web standard that's in flux. There's a risk there. Something like a select or list control might have a technical risk related to handling large sets of data, etc. I find it's worth it to at least think through these sorts of things.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+#### Slots
+
+*Example Table*
+
+| Slot Name | Description | Fallback Content |
+| --------- | ----------- | ---------------- |
+| `thumb` | Enables replacing the component's `thumb` part with custom markup. | A default thumb implementation is provided. |
+
+#### Host Classes
```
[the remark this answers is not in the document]
EisenbergEffect: In web components parlance, this is a css class that the component may add to its host element. They are part of the public API of the component as it affects user-land styling.

research/src/pages/component-spec-template.mdx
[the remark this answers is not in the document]
gregwhitworth: I guess I'll need to some specific examples because it feels like they maybe should be combined rather than disjointed. Let's not block on this however.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+*Example Table*
+
+| Part Name | Description |
+| --------- | ----------- |
+| `thumb` | Enables styling the `thumb` part without requiring it to be completely replaced via its slot. |
+
+---
+
+## Implementation Plan
```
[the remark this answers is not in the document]
EisenbergEffect: I don't know that removing this is a good idea. I can be renamed. The content below this header is definitely different than the content above.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+#### Form Input
+
+*Does this component integrate with form submission, form validation, etc.? Can integration be accomplish with standard APIs or are special work-arounds required?*
+
+#### Use with Assistive Technology
+
+*Are there any details about shadow dom, focus delegation, or aria attributes that need special attention?*
+
+### Globalization
```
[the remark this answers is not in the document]
EisenbergEffect: I would consider global readiness to be on a similar level of importance as accessibility. I don't think we should be designing any UI/UX without thinking about it. Depending on the component, there may be more or less to call out though. For example, a select may want to call out how the chevron should be positioned when in RTL mode. It may also want to point out that the chevron needs to be replaceable with a different visual.

research/src/pages/component-spec-template.mdx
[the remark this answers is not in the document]
gregwhitworth: @EisenbergEffect yeah there are definately risks for any component or it's sub features and am not stating that they shouldn't be thought out. That's precisely what issues are for which is to track things that are blocking a standard such as perf for large datasets, etc. Normally, in other specs these issues are linked inline to the area that has the Issue. I'd probably be more in favor of simply changing this to Issues and possibly having the important ones inline to their items that have the issue (here's an [example](https://drafts.csswg.org/css-color-4/#resolving-color-values) in BikeShed). This could then link to said issues. If there are challenges as it relates to specific sections of implementation then I think those should be noted inline so the implementor isn't having to go from implementation text to challenges that may/may not exist (it would be similar to comments in review not being inlined but just a combination at the bottom that you then had to join up with the code yourself). Again, I wouldn't block on this.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+- *Special RTL handling*
+- *Swapping of internal icons/visuals*
+- *Localization*
+
+### Security
+
+*Are there any security implications surrounding the component?*
+
+### Performance
```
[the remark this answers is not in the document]
EisenbergEffect: I think there are performance considerations with certain components that are pretty common across platforms with fairly standard solutions. UI virtualization is a common technique used by multiple platforms to handle long lists of data. In this section, the spec writer would call out the potential for large data sets as something to consider related to performance. They could note virtualization as a common technique to mitigate the potential issue.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+*Are there any performance pitfalls or challenges with implementing the component? (examples below)*
+
+* If the component renders a loop of items, should certain areas be considered for virtualization?
+* If the component needs to measure an area before rendering, how will jank be avoided?
+* If any measuring is needed at all, can rAF techniques help queue measures and prevent synchronous reflows?
+
+### Dependencies
```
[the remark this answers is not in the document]
EisenbergEffect: See response above. Most of these sections are here to make the spec writer think through things thoroughly. Often times "features" or bugs are accidentally implemented because someone didn't think about the various implications of things, be it perf, security, globalization, etc.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+*Will implementing the component require taking on any dependencies?*
+
+- *3rd party libraries*
+- *Upcoming standards we need to polyfill*
+- *Do any dependencies bring along an associated timeline?*
+
+#### Platform Requirements
```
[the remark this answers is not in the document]
EisenbergEffect: If it's no possible to implement the component or the implementation is dependent on a not-yet-implemented proposal, I think that should be called out here. Or, if there are major downsides to working around a missing platform feature, that could be called out here as well.

research/src/pages/component-spec-template.mdx
EisenbergEffect: I can remove this section. It was initially intended to call out if there were any abnormal or non-obvious testing considerations related to validating the correctness according to the spec.

research/src/pages/component-spec-template.mdx
EisenbergEffect: The intent was related to DevTools and how they could be enhanced in order to better work with the component.

research/src/pages/component-spec-template.mdx
EisenbergEffect: Yes, this could be a list of links to follow-on issues or also just thoughts on what could be done with the component beyond the initial MVP outlined by the spec.

research/src/pages/component-spec-template.mdx
EisenbergEffect: Makes sense.

research/src/pages/component-spec-template.mdx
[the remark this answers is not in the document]
EisenbergEffect: I've changed some of the wording to remove anything related to implementation. I've made this into general risks/challenges and added a note about linking issues for platform blockers.

research/src/pages/component-spec-template.mdx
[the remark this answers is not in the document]
EisenbergEffect: I've added some clarifying wording into this section as well as similar sections.

research/src/pages/component-spec-template.mdx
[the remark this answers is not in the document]
EisenbergEffect: I've renamed this to "Behavior" since that's mostly what is covered.

research/src/pages/component-spec-template.mdx
EisenbergEffect: I cleaned up some wording to make it clearer that this was related to DevTools.

research/src/pages/component-spec-template.mdx
EisenbergEffect: I added some clarifying words based on your recommendation.

research/src/pages/component-spec-template.mdx
EisenbergEffect: I've updated the wording to indicate the research page should be linked and that if any additional examples are found while writing the spec, they should be submitted through a separate PR to the research page.

EisenbergEffect:
@chrisdholt I think I've got most of the feedback addressed now. When you get a chance, do you want to do a final review? If nothing else major comes up, we should be good to merge.

research/src/pages/component-spec-template.mdx
```diff
@@ -0,0 +1,217 @@
...
+| `min` | `min` | `number` | `0` | The minimum value of the slider. |
+
+#### Methods
+
+*Example Table*
+
+| Signature | Description |
+| ----------- | ---------- |
+| `increment(value: number = this.step): void` | Increments the `value` of the component by the amount specified by `step`, clamped within `min`/`max` values. |
+| `decrement(value: number = this.step): void` | Decrements the `value` of the component by the amount specified by `step`, clamped within `min`/`max` values. |
```
chrisdholt, line 66: Great examples!

EisenbergEffect:
Appears to need @levithomason to approve before it can be merged.

PR #52, opened 2020-03-19
Improve image alt tags and specimens

### LINKED ISSUE #40: Images need useful text descriptions

levithomason:
Fixes #40 

This PR fixes image alt tags.  In order to do this, the function for returning images was updated to include the design data.  We now have nice alt tags and images grouped by design system.

# Before

image

image

# After

image

image

### DIFF
--- a/research/src/components/specimens.js
+++ b/research/src/components/specimens.js
@@ -10,13 +10,46 @@ const Specimens = ({ component, concept }) => {
   return (
     <div
       style={{
-        padding: hasImages ? '1em' : '2em',
-        marginBottom: '2em',
+        display: 'flex',
+        flexWrap: 'wrap',
+        marginBottom: '8px',
         border: '1px solid #ccc',
       }}
     >
-      {images.map((src, i) => (
-        <Image key={i} src={src} title={src} />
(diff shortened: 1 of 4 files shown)


research/src/components/specimens.js
```diff
@@ -10,13 +10,46 @@ const Specimens = ({ component, concept }) => {
...
+            flexDirection: 'column',
+            padding: hasImages ? '8px' : '16px',
+            textAlign: 'center',
+          }}
+        >
+          <div style={{ margin: 'auto', padding: '4px' }}>
+            <Image
+              src={image.image}
+              title={image.image}
+              alt={`An image of the ${concept} concept on a ${component} component in ${image.sourceName}.`}
```
nicholasrice: It might be good to add an alt data field also, and make this formatted string the fallback if that field doesn't exist.

levithomason: Agreed, let's do in follow up PRs.

PR #54, opened 2020-03-23
Research: adds checkbox draft proposal

nicholasrice:
Adds a draft checkbox proposal based on checkbox research. Also updates several fields in the research page for checkbox.


research/src/components/checkbox-anatomy.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import './anatomy.css'
+
+export default () => {
+  return (
+    <div className="component-anatomy">
+      <host>
+        <part name="label">
```
gregwhitworth, line 8: Do the frameworks really have a label built into the checkbox control? From the webplat perspective the label is its own label and the relationship is made either through descendant or for attribute. I'm not saying this isn't the case but I think this may warrant a discussion/issue.

research/src/components/checkbox-anatomy.js
```diff
@@ -0,0 +1,18 @@
...
+          <slot>{'Default'}</slot>
+        </part>
+        <part name="checkbox">
+          <part name="checked-indicator" />
```
gregwhitworth, line 12: I'm intrigued at the need for different parts. Is this to inform that there are two states of which can both be on/off? Cause ultimately this could just be a series of icons that have their visibility toggled upon state change, no?

research/src/pages/checkbox.proposal.mdx
```diff
@@ -0,0 +1,82 @@
+---
+menu: Components
+name: Checkbox (Working Draft)
+path: /components/checkbox
+pathToResearch: /components/checkbox.research
+---
+
+import CheckboxAnatomy from '../components/checkbox-anatomy'
+
+The `<checkbox>` element is a control that allows the user to select a binary state, either checked or unchecked.
```
gregwhitworth, line 10: probably worth denoting indeterminate state

research/src/pages/checkbox.proposal.mdx
```diff
@@ -0,0 +1,82 @@
...
+Implements the [checkbox role](https://www.w3.org/TR/wai-aria-1.1/#checkbox) per the
+[HTML AAM spec](https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings).
+
+### Forced Colors
+
+Placeholder for now, dependent on standardization on parts definitions.
+
+## States / Properties
+
+### &ltcheckbox&gt
```
gregwhitworth, line 58: Should be updated using @EisenbergEffect template

research/src/components/checkbox-anatomy.js
```diff
@@ -0,0 +1,18 @@
...
+import './anatomy.css'
+
+export default () => {
+  return (
+    <div className="component-anatomy">
+      <host>
+        <part name="label">
+          <slot>{'Default'}</slot>
+        </part>
+        <part name="checkbox">
```
gregwhitworth, line 11: Also, I noticed that there is no slot in here? How are you expecting the user of this checkbox to adjust the content/colors, etc?

research/src/components/checkbox-anatomy.js
nicholasrice: Typically the label is *optional* or there is a mechanism to make the label visible only to AT, but generally Checkbox components have a simple, first-class mechanism to create a label for the checkbox.

I anticipated discussion here - it certainly differs from the way the platform works but it *is* aligned with the research of component libraries.

research/src/components/checkbox-anatomy.js
nicholasrice: My understanding is the "part" convention denotes an element selectable via CSS, similar to how a pseudo is targetable. I'm not sure there is a need to provide *content* to this element, though I think there is an argument for providing content for the indicators themselves.

I'm not really fond of this name because of the redundancy. Functionally this represents the container for the checkbox control itself. This complication arises because, in this model (and in many component libraries), the "checkbox" is a checkbox + label pairing. In FAST we've adjusted this to be "control", but perhaps there is a better name? Happy to hear ideas here.

research/src/components/checkbox-anatomy.js
nicholasrice: I think thats a good point. Another point that was brought up was that, being different elements, it would make it much harder to *transition* between visual treatments of checked vs indeterminate.

But yes, the original idea was that because `checked` and `indeterminate` are distinct states, there could be mechanisms to provide indicators for each. Internally, default behavior of the control would show indeterminate when indeterminate, and when indeterminate is false, would show checked when checked. Exposing a single part / slot w/ a set of child icons would require more work by the app-author to indicate which should be shown when.

research/src/pages/checkbox.proposal.mdx
nicholasrice: Good catch - it's missing from the states / properties section also. I'll add.

research/src/components/checkbox-anatomy.js
gregwhitworth: And it is tied to the individual host of type `checkbox`? I think I can live with this although I find it kind of awkward but I get your overall point and we always recommend that folks have a label.

research/src/components/checkbox-anatomy.js
gregwhitworth: I get your overall premise with regards to the parts, but it's making an assumption of what the checkbox will contain and how it will want to be styled/modified. If everything that anyone could do with a checkbox could be done solely in CSS then a part would be sufficient. If someone needs potentially additional layers for particle effects or some other additive design effect they shouldn't be inhibited from doing so. The slot doesn't need to be leveraged but should be possible.

nicholasrice: @gregwhitworth can you help me understand your recommendation? 

Are you proposing that an app author can slot arbitrary content *inside* this element or are you proposing that the entire element implementation can be whole-sale replaced. In my mind, the later would *also* remove the checked/indeterminate indicators because they are structurally children of the implementation.

research/src/components/checkbox-anatomy.js
nicholasrice: As you've suggested I've created an [issue ](https://github.com/WICG/open-ui/issues/93) to discuss.

research/src/components/checkbox-anatomy.js
EisenbergEffect: With slots for both indicators, I also wasn't quite sure what it would mean to had a slot for the entire checkox. I'm also not sure that you can keep the slots for the indicators if you do that. So, I'd be worried that we would increase the complexity of customization for the most common cases.

Just thinking out loud here...by combining Houdini custom paints via CSS with CSS parts, is there anything that this configuration really can't handle that absolutely needs the "everything" slot?

PR #59, opened 2020-03-26
Specific low level goals for web specifications

gregwhitworth:
Bo Cupp and I were reviewing the vision and he asked a good question if we had specific aspects we wanted to standardize on. As such, I added specific goals and non-goals which I think are worth agreeing on since the template has some of these sections but we should be sure to not duplicate other specs and end up being an [XKCD comic strip](https://xkcd.com/927/). I welcome any and all thoughts. @chrisdholt @levithomason @EisenbergEffect @dzearing @nicholasrice

### DIFF
--- a/research/src/pages/mission.research.mdx
+++ b/research/src/pages/mission.research.mdx
@@ -39,7 +39,6 @@ For web app component specifications here are the specific goals:
 - Standardize on the names of the component parts and anatomy structure
 - Standardize the name of states which are applicable to the component and its parts
 - Standardize the behaviors of components
-- Standardize the base style upon which authors can build a custom appearance
 - Update other specifications in other standards bodies when gaps are discovered (eg: If
   discover the need for a new HTML element or CSS property, propose and have it standardized in
   the WHATWG HTML spec and CSSWG respectively)
--- a/research/src/pages/mission.research.mdx
+++ b/research/src/pages/mission.research.mdx
@@ -32,6 +32,24 @@ These are the goals Open UI believes will realize the above vision.
 - Eventual browser standards for web app components
 - Converging designer processes and developer workflows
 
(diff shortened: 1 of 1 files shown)


research/src/pages/mission.research.mdx
```diff
@@ -32,6 +32,24 @@ These are the goals Open UI believes will realize the above vision.
...
 - Converging designer processes and developer workflows
 
+## Goals for web specifications
+
+For web app component specifications here are the specific goals:
+
+- Standardize on the names of the component parts and anatomy structure
+- Standardize the name of states which are applicable to the component and its parts
+- Standardize the behaviors of components
+- Standardize the base style upon which authors can build a custom appearance
```
EisenbergEffect: Can you give an example of "base style". I'm not sure how to draw the line between that and the non-goal of non-standardization of "default appearance".

gregwhitworth: Good question, the base style are things that may/may not be necessary for this group but whether or not the part should be BFCs, Inlines, etc. From a platform perspective we would want to align on this and then probably have a discussion regarding how to negate all of these and set a base set of styles (such as appearance: none does not necessarily set inlines to block, etc. It removes a certain set of styles to ensure that the component still functions). @BoCupp-Microsoft do you have anything to add to this?

gregwhitworth: Bo and I discussed this a bit and this is definitely a platform requirement but not sure it needs to be in there for frameworks for now I'm going to remove it.

gregwhitworth:
@EisenbergEffect @chrisdholt can one of you please review this again now that I have updated it. I removed the aspect that slighly confused Rob and I think that it is specific to UAs and may not apply to frameworks and as such should be a non-goal

PR #60, opened 2020-03-27
Clean up buttons and improve concepts

levithomason:
This PR adds many missing button images.  It also fixes many of the images to be more appropriate.

I also updated the `openUIName` which is used to link similar specimens.  Here's an example of how we display grouped images now:

image

- The `openUIName` is used for the heading
- There is a coverage meter for each other name it is known by
- The specimens that are known by a different name in their design system show their other name in **bold**


research/gatsby-browser.js
```diff
@@ -0,0 +1,17 @@
...
+const anchors = new AnchorJS({
+  placement: 'left',
+})
+
+exports.onRouteUpdate = (location, prevLocation) => {
+  anchors.add()
+}
+
+exports.onClientEntry = (...args) => {
+  console.log('onClientEntry')
```
nicholasrice, line 12: still need the console log?

research/src/components/concepts.js
```diff
@@ -1,22 +1,77 @@
...
 import Specimens from './specimens'
 
 const Concepts = ({ component }) => {
+  const [showDescriptions, setShowDescriptions] = React.useState(false)
+
+  const conceptsForComponent = _.sortBy(
+    _.toPairs(openUIConceptsByComponent[component]),
+    ([openUIName, concepts]) => -concepts.length,
+  )
+  console.log(conceptsForComponent)
```
nicholasrice, line 14: needed?

gregwhitworth:
@levithomason - how are things coming with this?

gregwhitworth:
@levithomason is there a reason to hold this PR up?

levithomason:
Thanks for the ping.  Will get this released.

Note, there's a billing issue with Netlify that is failing PRs that I've been working on in the background.  We should be able to release without that.

PR #62, opened 2020-03-31
Move research focused components under analysis

gregwhitworth:
Due to seeing some confusion from folks on hackernews, while we hash out some things I think it's important to try and show folks that most are in the research phase. Technically, I think ALL of them are more than anything but I decided to split it on which ones currently have a proposal or not.

In addition to that, the W3C has a notion of Editor's Draft and Working Draft, the Working Draft has an expectation that things have been discussed and reviewed by the community. This currently isn't the case for <select> nor the table component. As such I've set their proposals to editors draft to reflect that.

I understand that this isn't the perfect solution but until we have phases in place to denote it clearly we should try and avoid people from thinking that Research = Proposal because that isn't the case and Analysis kind of implies this.


gregwhitworth:
Adding @chrisdholt as a reviewer since @levithomason is OOF, for feedback.

gregwhitworth:
@levithomason I'm going to merge this in. I welcome any adjustments you want to this but I'd really like to start differentiation in some way and this will be a stand in until we have a way to delineate current progress of the items.

PR #66, opened 2020-04-01
Add YAML front matter to specify page title when rendered to HTML

Verious:
Added YAML front matter to specify page title when rendered to HTML.


levithomason:
Thanks much!

PR #72, opened 2020-04-16
Adding initial editors for select & template

gregwhitworth:
This is a simple change based on our resolution for adding editors.

This is to add editors for select and the spec template of @EisenbergEffect in addition to Levi and Myself.

I need approval from you @levithomason and it would be good to ensure that you want this responsibility @EisenbergEffect :)


gregwhitworth:
I'm going to go ahead and squash this commit

PR #87, opened 2020-05-08
Clean up <select> property/attribute/method tables

### LINKED ISSUE #86: Inconsistent classification of <select> attributes/methods/properties
At https://open-ui.org/components/select some states are listed as both attributes and as methods of `<select>`,   Examples are  `autofocus`, `disabled`, and `form`.  All of these should be either one or the other.

The same page also has separate tables for `<option>` attributes and `<option>` properties.  It doesn't seem to me like this is a useful distinction to make in this context.  Would it be clearer to just enumerate the properties (aka IDL attributes)?  Or is there value in accounting for the content attributes separately?

dandclark:
This PR fixes a few issues with the property/attribute/method tables for `<select>` at https://open-ui.org/components/select.

* Several non-methods were listed as methods for `<select>`.  These are moved to the property table.
* Added some missing methods for `<select>`'.
* `<option>`'s list of attributes was incomplete.   Rather than fix this up and have parallel tables for content attributes and properties, I've changed it so that only properties are listed.  All of the relevant content attributes are reflected in IDL attributes aka properties, so there does not seem to be value in maintaining separate tables for each.
* Make the anchor tags on the table headers consistent.

Closes #86


dandclark:
Deployment failures appear to have come from https://github.com/WICG/open-ui/pull/60.
Taking a look.

gregwhitworth:
Actually, I just noticed that there are deploy failures. @levithomason you had mentioned there were builds breaking - does this failure look the same to you before we dig in further?

gregwhitworth:
> Deployment failures appear to have come from #60.
> Taking a look.

Thanks Dan

dandclark:
It's definitely the same failure visible in the [logs](https://github.com/WICG/open-ui/pull/60/checks) from #60.

5:40:07 PM: error Building static HTML failed for path "/components/select.research"
5:40:07 PM:    99 | ) =>
5:40:07 PM:   100 |   _.map(
5:40:07 PM: > 101 |     _.get(conceptsByComponent, [componentName, conceptName]).filter(
5:40:07 PM:       |                                                              ^
5:40:07 PM:   102 |       concept => concept.name === conceptName && concept.openUIName === conceptOpenUIName,
5:40:07 PM:   103 |     ),
5:40:07 PM:   104 |     'sourceName',
5:40:07 PM: 
5:40:07 PM:   WebpackError: TypeError: Cannot read property 'filter' of undefined
5:40:07 PM:   
5:40:07 PM:   - index.js:101 getSourcesWithComponentConcept
5:40:07 PM:     src/sources/index.js:101:62
5:40:07 PM:   
5:40:07 PM:   - concept-coverage.js:14 ConceptCoverage
5:40:07 PM:     src/components/concept-coverage.js:14:54
5:40:07 PM:   

I added some logging on my end and `Select` `minCharsBeforeSearch` is the componentName/conceptName causing the failure.  I don't yet understand enough about what this is supposed to be doing to know what the fix is, but if I figure it out I'll create another PR.

gregwhitworth:
Ok, I'll merge this in for now since it's the same failure. @levithomason any help you can provide is appreciated

PR #88, opened 2020-05-08
Fix build break from wrong param passed to <ConceptCoverage>

dandclark:
This PR fixes a build break caused by an instance of `<ConceptCoverage>` in concepts.js where the conceptOpenUIName was being passed to the conceptName parameter.  This resulted in a failed lookup when we try to search the concept framework-specific names for the openUIName.  The fix is to pass the uniqueName to the parameter instead.

This was triggered by Semantic's Select's minCharacters which has an openUIName that is not present anywhere else, with the result that we went down the buggy codepath.

### DIFF
--- a/research/src/components/concepts.js
+++ b/research/src/components/concepts.js
@@ -43,13 +43,13 @@ const Concepts = ({ component }) => {
                 <ConceptCoverage
                   component={component}
                   openUIConceptName={conceptOpenUIName}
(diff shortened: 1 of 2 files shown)


dandclark:
cc @gregwhitworth, this fixes the build break that #87 hit.

gregwhitworth:
Woohoo the deploys are working - you rock @dandclark, cc: @levithomason as FYI

Issue #91, opened 2020-05-20
Define file input anatomy


gregwhitworth:
There was a request for this to the CSSWG here: https://github.com/w3c/csswg-drafts/issues/5049

They are currently focused on the psuedo element for button but it would be good to have a full anatomy breakdown on a proposal page backed by our research.

gregwhitworth:
In the PR I added the following:

Looking through the different parts across the different component libraries:

- File Picker
- File Selector

I think both of these are better than chooser but I do think all of them should be appended with button. @Roy-Orbison then made the valid point regarding other standards referencing this control:

> The majority of standards documents for HTML and JavaScript refer to such controls as file upload, and the action performed as selection, and have done so for many years. What benefit is there to CSS authors in deviating from those conventions?

Because of this I'd like to recommend that we use `file-selector-button`. Thoughts?

gregwhitworth:
We discussed this in today's telecon, the minutes are here.
Here is the resolution:

> RESOLUTION: Use the name file-selector-button for the file input button part

@mfreed7 had a good point that we should ensure that this naming aligns with the anatomy we're actively investigating for `<select>`.

andrico1234:
Closing this as the [File proposal](https://open-ui.org/components/file) has been added to the Open UI site

PR #92, opened 2020-05-20
Add an initial draft for file inputs.

emilio:
See https://github.com/w3c/csswg-drafts/issues/5049.


research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,165 @@
...
+Custom inputs on the web are generally achieved by hiding the original input.
+Some approach it by hiding it in a way that the user can still hit (like
+`opacity: 0`), others use JS to forward the click from some other UI element to
+the underlying file input.
+
+- [Bootstrap](https://getbootstrap.com/docs/4.0/components/forms/#file-browser)
+  uses the former approach, by generating a "Browse" button with generated
+  content.
+
+- [Carbon Design System](https://getbootstrap.com/docs/4.0/components/forms/#file-browser)
```
chrisdholt, line 58: This appears to be the Bootstrap link as well

emilio: Whoopsies, fixed. Thanks!

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,165 @@
...
+#### DOM Structure <a href="#dom-structure" id="dom-structure"></a>
+
+```html
+<button tabindex="-1">Browse...</button>
+<!-- Or appropriately-localized string -->
+<label><!-- implementation and state-dependent --></label>
+```
+
+It is important that the button is not tab-navigable, as the input is focused
+atomically, see below.
```
Roy-Orbison, line 96: Side note: EdgeHTML & Trident need two tabs, one for the whole input and one for the button within. Both trigger the same action.

marcoscaceres:
Hi @emilio, as this is a substantive contribution, could you [please join the wicg](https://www.w3.org/community/wicg/)? thanks in advance!

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,165 @@
...
+#### Diagram <a href="#diagram" id="diagram"></a>
+
+TODO
+
+#### DOM Structure <a href="#dom-structure" id="dom-structure"></a>
+
+```html
+<button tabindex="-1">Browse...</button>
+<!-- Or appropriately-localized string -->
+<label><!-- implementation and state-dependent --></label>
```
gregwhitworth, line 92: The label and button have slots correct? Also based on the name that you have for part these should be referenced with part attributes.

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,165 @@
...
+#### CSS Parts <a href="#css-parts" id="css-parts"></a>
+
+| Part Name             | Description                                                                                                                     |
+| --------------------- | ------------------------------------------------------------------------------------------------------------------------------- |
+| `file-chooser-button` | Enables styling the `&lt;button>` or `&lt;input type=button>` part without requiring it to be completely replaced via its slot. |
+
+<p class="question">
+We should consider exposing the `&lt;label>`, maybe?
+
+Maybe even with a `&lt;slot>` in it so that people could replace it with their own
```
gregwhitworth, line 107: Yep, see my comment above. We don't want to simply expose the anatomies for adjusting them but also for replacing content as they may not want the word `Browse` but an icon. I'd recommend dropping in a few additional images to show this under usecases.

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,165 @@
...
+## Accessibility <a href="#accessibility" id="accessibility"></a>
+
+File Inputs should be focusable via a keyboard navigation.
+
+#### Keyboard Navigation and Focus <a href="#keyboard-navigation" id="keyboard-navigation"></a>
+
+File input is focused atomically.
+
+<p class="question">
+  Seems like ideally we could delegate focus to the button, or something something... Gecko used to
```
gregwhitworth, line 135: Let's file a separate issue to discuss this one, leave it as an open question.

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,165 @@
...
+
+---
+
+### API <a href="#api" id="api"></a>
+
+See [the Resources section](#resources) for the specification and the MDN reference.
+
+#### Properties and Attributes <a href="#properties-attributes" id="properties-attributes"></a>
+
+See [the Resources section](#resources) for the specification and the MDN reference.
```
gregwhitworth, line 69: I'd prefer for this to be filled in. That said - I'd also love to chat with the MDN team to see if they have an API endpoint we can possibly pull this in from on build.

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,165 @@
...
+```
+
+It is important that the button is not tab-navigable, as the input is focused
+atomically, see below.
+
+#### CSS Parts <a href="#css-parts" id="css-parts"></a>
+
+| Part Name             | Description                                                                                                                     |
+| --------------------- | ------------------------------------------------------------------------------------------------------------------------------- |
+| `file-chooser-button` | Enables styling the `&lt;button>` or `&lt;input type=button>` part without requiring it to be completely replaced via its slot. |
```
gregwhitworth, line 102: Looking through the different parts across the different component libraries:

* File Picker
* File Selector

I think both of these are better than chooser but I do think all of them should be appended with button

Roy-Orbison: I agree with the `-button` suffix as the selector's purpose essentially dictates it. However, for consistency's sake, the verb really should be a choice between "upload" and "select"/"selector". The majority of standards documents for HTML and JavaScript refer to such controls as file upload, and the action performed as selection, and have done so for many years. What benefit is there to CSS authors in deviating from those conventions?

There is merit to each:

* the button triggers a file selection system but operating it doesn't guarantee a file upload attempt by the user; and
* emphasising upload as the action is important because successful selection of any file immediately makes the file's name and contents available to running script, unlike the olden days when it was only a state of intent to upload upon form submission.

gregwhitworth:
@emilio any update on the changes requested? Let me know if you need a hand in any way.

For bug linking, this will resolve #91

research/src/pages/file.proposal.mdx
gregwhitworth: @Roy-Orbison that seems like a valid argument for selector then. One aspect that Open UI is trying to ensure is that it uses nomenclature used by web developers building controls/components today as well as industry standards or other operating system paradigms, etc. So based on this I think `file-selector-button` seems like a good fit. @emilio what do you think?

emilio:
No, just got a bit swamped with other work, will get to this probably tomorrow or early next week.

gregwhitworth:
@emilio Thanks. As we discussed today on the call the button part should be named `file-selector-button`. https://github.com/WICG/open-ui/issues/91#issuecomment-639160328

research/src/pages/file.proposal.mdx
emilio: That seems... unfortunate, fwiw.

research/src/pages/file.proposal.mdx
emilio: I think they should have a _wrapping_ slot, so that you can replace the whole thing, not just the node inside it.

research/src/pages/file.proposal.mdx
emilio: @chrisdavidmills do you know if MDN has some API we could use here?

(I'd rather punt on this until we have the setup figured out, but let me know if you want to go through the whole thing... Not all `<input>` APIs apply to `type=file` I guess.)

research/src/pages/file.proposal.mdx
emilio: Filed #107, referenced from here..

research/src/pages/file.proposal.mdx
emilio: Added slots in both.

research/src/pages/file.proposal.mdx
gregwhitworth: I'm ok punting on it for now but can you file an issue for it then?

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,178 @@
...
+#### DOM Structure <a href="#dom-structure" id="dom-structure"></a>
+
+```html
+<button tabindex="-1" part="file-selector-button">
+  <slot name="button">
+    <!-- Localized "Browse..." string -->
+  </slot>
+</button>
+<label part="file-selector-label">
+  <slot name="label">
```
gregwhitworth, line 107: Let's have the slot names be the same as the part for consistency.

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,178 @@
...
+| `file-selector-button` | Enables styling the `&lt;button>` part without requiring it to be completely replaced via its slot. |
+| `file-selector-label`  | Enables styling the `&lt;label>` part without requiring it to be completely replaced via its slot.  |
+
+---
+
+## Behavior <a href="#behavior" id="behavior"></a>
+
+### States and Interactions <a href="#states-interactions" id="states-interactions"></a>
+
+- As a result of keyboard (either Enter or Spacebar without modifier) or mouse
```
gregwhitworth, line 128: Not to block on this PR, but please see the table that we just landed for `<select>` which begins to get very specific about what specific events are tied to what parts and the necessary changes. https://open-ui.org/components/select#events-1

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,178 @@
...
+  / touch interaction (click), an implementation-dependent dialog will show
+  up. This dialog cannot be accessed by the website, but its contents and
+  appearance may depend on things like the `accept` attribute.
+
+- Dragging an file into the component may also be used to select a particular
```
gregwhitworth, line 133: nit `an` -> `a`

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+---
+
+## Behavior <a href="#behavior" id="behavior"></a>
+
+### States and Interactions <a href="#states-interactions" id="states-interactions"></a>
+
+- As a result of keyboard (either Enter or Spacebar without modifier) or mouse
+  / touch interaction (click), an implementation-dependent dialog will show
+  up. This dialog cannot be accessed by the website, but its contents and
```
gregwhitworth, line 130: I think we'll want to get more specific than this. Initially we'll want the following:

* What happens when a user selects a file (eg: the value of file is updated)
* What happens when a user presses cancel, or similar button if offered by the dialog?

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,178 @@
...
+  appearance may depend on things like the `accept` attribute.
+
+- Dragging an file into the component may also be used to select a particular
+  file.
+
+## Accessibility <a href="#accessibility" id="accessibility"></a>
+
+File Inputs should be focusable via a keyboard navigation.
```
gregwhitworth, line 138: You hit on this just following so this line feels redundant let's just delete it.

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,178 @@
...
+
+</p>
+
+#### Form Input <a href="#form-input" id="form-input"></a>
+
+This component integrates naturally with form submission as described in [the HTML spec](<https://html.spec.whatwg.org/#file-upload-state-(type=file)>).
+
+### Globalization <a href="#globalization" id="globalization"></a>
+
+The component reacts naturally to directionality. The initial label content and
```
gregwhitworth, line 159: I'm not sure what you mean by it reacts naturally to directionality. Since Open UI does not get into styling of the controls technically this may not be true for a given implementation of a file control eg: they may not use logical properties and thus the positioning of parts may not follow directionality.

I'm not sure if there is a need for either part to follow writing mode direction but I could be wrong. I haven't investigated this much. If you don't have time to dig into this, possibly open an issue and either clarify what you mean by directionality or remove it.

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,178 @@
...
+the button contents should be localized.
+
+### Security <a href="#security" id="security"></a>
+
+The component very intentionally doesn't expose to the outside world anything
```
gregwhitworth, line 164: This should be written in a normative form. Possibly something like:

> The implementation should ensure that the component does not provide access to content on the user's file system unless the user has selected it.

gregwhitworth: Also, there may be normative text in WHATWG that is similar to this and you may just want to link to that

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,178 @@
...
+---
+
+## Resources <a href="#resources" id="resources"></a>
+
+- [MDN Reference for `<input type=file>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file)
+- [HTML Specification](<https://html.spec.whatwg.org/#file-upload-state-(type=file)>)
+
+---
+
+## Next Steps <a href="#next-steps" id="next-steps"></a>
```
gregwhitworth, line 176: We'll have issues opened and as we continue to work through the spec we'll find more next steps so let's remove this section

research/src/pages/file.proposal.mdx
```diff
@@ -0,0 +1,178 @@
...
+TODO
+
+#### Diagram <a href="#diagram" id="diagram"></a>
+
+TODO
+
+#### DOM Structure <a href="#dom-structure" id="dom-structure"></a>
+
+```html
+<button tabindex="-1" part="file-selector-button">
```
gregwhitworth, line 101: @dandclark and I have an explainer that we're working on in conjunction with a few other folks that we hope to publish shortly. Can you wrap this with `<file>`. I am fine if you want to open a seperate issue to discuss this but we desire to light these anatomies up and of course currently we can't have slots in a replaced element. As such we'll need to discuss introducing a new element of `<file>` to allow for this. If you don't want to put that in there open an issue and we won't block the PR on that.

research/src/pages/file.proposal.mdx
chrisdavidmills: We have a separate reference doc for each input type — see [`<input type="file">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file), and a [section here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement#Properties_file) that lists properties of the `HTMLInputElement` interface that apply only to file types.

But we don't have an API that would expose such info automatically, as such. I'd be interested to hear more info about what you'd want here.

gregwhitworth: @chrisdavidmills to avoid discussing this in the PR is there a place we can have this discussion. Possibly on an MDN github issue?

chrisdavidmills: @gregwhitworth yup — best place is an issue on https://github.com/mdn/sprints/issues/

BillGoldstein:
Why a label element rather than an output element?
Considering the input type=file may already have a label for="" and/or be wrapped in a label element.

gregwhitworth:
@BillGoldstein that's a valid point and it won't change the name computation in any way and does semantically represent what is actually occurring in a more meaningful way. @emilio can you change that as well?

una:
Thank you for this PR! Closing in favor of #140

PR #94, opened 2020-05-26
Slider a11y verification

kolaps33:


kolaps33:
wrongly created PR by mistake

PR #97, opened 2020-05-27
Added definitions to contribute

### LINKED ISSUE #81: Definition of a "control"?
Would it be useful to include a more-or-less formal definition of a "control" somewhere in the docs?  The [select](https://open-ui.org/components/select) Editor's Draft uses the term but I'm not sure whether there's actually a formal definition.  If nothing else this could help clarify the Charter regarding which kinds of widgets fall under the purview of OpenUI.

One way to attempt a definition would be to describe what a control *does*, e.g.:

A control is an HTMLElement with these properties:
1. Maintains a value that is reflected in the `value` property.  Fires `input` and `change` events corresponding to changes in the control's `value`.* 
1. Has a common set of APIs for validation of its `value` (`validity`, `validationMessage`, `willValidate`, `checkValidity()`, `reportValidity()`, the `invalid` event, and others that are specific to particular control types). 
1. Participates in `<form>`s and form submission. 
1. Has some user interface that allows a user to interact with the control to view and/or change its value.*

\* Partial exception: `<input type=”button”>` and `<input type=”submit”>` don’t fire `input`/`change` events and their value cannot be changed by user interaction, only viewed (as button text).

gregwhitworth:
I've updated the contribution document again. This time adding a terrible subnav at the top of the document but will hopefully allow people to jump down a bit quicker. Added in some anchors for sharing. Also added in some definitions for use across documents. The definition of a control is the important aspect as that is what is outlined in each of our specs as top level sections.

Also forced Home to be the first nav item.

This resolves #81

### DIFF
--- a/research/src/pages/contribute.mdx
+++ b/research/src/pages/contribute.mdx
@@ -72,7 +72,7 @@ A design system does not need to meet all of these criteria but it should meet m
 The design systems we've launched with we believe are representative of these criteria.
 We may not accept PRs contributing design systems that are lacking by these criteria.
 
-### Adding a design system <a href="#add-a-design-system" id="add-a-design-systems"></a>
+### Adding a design system <a href="#add-a-design-system" id="add-a-design-system"></a>
 
 Create a JSON5 file in `/sources` for the design system, like `/antd.json5`.
 Add the `$schema` key pointing to our `design-system.schema.json5` schema and complete the required fields.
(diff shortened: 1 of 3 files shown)


research/src/pages/contribute.mdx
```diff
@@ -222,16 +263,28 @@ to be filed again the WHATWG HTML specification.
...
+## Component <a href="#definition-component" id="definition-component"></a>
+
+The dictionary defines a component as, "a part or element of a larger whole" and this is
+true for components in a design system's component library. Components normally have a meaning,
+although this meaning may not be presented to the user (such as a utility component). A component,
+as the definition states can be a part of another component which are known as **composite components**.
+
+## Control <a href="#definition-control" id="definition-control"></a>
+
+A control is a subset of component that allows some form of user interaction. Controls have a defined model,
```
dandclark, line 288: Do we want things like `<meter>` and `<progress>` to fall under the purview of this project?  Since the user can't manipulate them I'm not sure that they fit here unless we count reading the value to be "some form of user interaction"...but by that interpretation a `<div>` would match too because a user can read the contents.

That said I don't have a better suggestion at the moment, and it's probably not critical to bikeshed this part of the definition right now since IMO the more important part is the stuff about the defined model, parts, slots, states, and controller code.

marcoscaceres:
@gregwhitworth, you might need to rejoin the WICG (as either independent or... )... we look forward to having you back 🤗🤗🤗

research/src/pages/contribute.mdx
gregwhitworth: I did think about those, such as output or details/summary and I find those to be a component. So maybe the model and state should be moved up to component and that user interaction bit can stay on control?

dandclark: Hmm, yeah I think that would be a good way to break it down.

gregwhitworth: Alrighty - I took a stab at adjusting it a bit. Give it another glance and let me know your thoughts.

PR #99, opened 2020-05-29
Updated anatomy with named outer slots

### LINKED ISSUE #80: Slotting custom content into controls using new element types vs built-in <slot>s
The [Editor's draft for `<select>`](https://open-ui.org/components/select) includes an [anatomy](https://open-ui.org/components/select#anatomy) breaking the control down into distinct parts and slots where a page author could swap in their own HTML to customize the control.

As currently written, the page author would use the `<button>` and new `<listbox>` element to slot their custom content into the `<select>`, e.g.
```html
<select>
    <button>My custom button with custom styles</button>
    <listbox> <!-- potentially styled by the page author -->
        <option>option 1</option>
        <option>option 2</option>
    </listbox>
</select>
```

I'd like to consider an alternative approach where the `<select>` parts are replaced by referencing named `<slot>`s.  For example:
```html
<select>
    <div slot="select-button">My custom button with custom styles</div>
    <div slot="select-listbox"> <!-- potentially styled by the page author -->
        <option>option 1</option>
        <option>option 2</option>
    </div>
</select>
```

Advantages to using elements like `<button>` and `<listbox>` for customizing control UI parts:
* Potentially more intuitive for developers.  With built-in elements it's obvious that the author is doing something standard, whereas this might not be so clear with named `<slot>`s.
* No concerns with `<slot>` name collisions.  I'm not sure how much of a real concern this is though, since the scope of the `<slot>` names is only under a given `<select>` element.

Advantages to using `<slot>`:
* Path to standardization and implementation is shorter without the need to introduce new element types (or update behavior for existing elements like `<button>`).
* There's already well-defined support for fallback content in `<slot>`s for handling cases where the page author wants to replace some parts of the control but not others.  This could be more complex with the new element approach, particularly if we would be relying on parser fixup to fill in elements not supplied by the page author.
* More scalable as we roll this out to more control types.  E.g. for `<input type="slider"/>` it could be easier to standardize more named slots versus introducing `<thumb>`/`<slider>` elements, etc.

I've fleshed out this idea further and provided more examples at https://jsfiddle.net/dandclark_msft/f8vmk6np/2/.

Thoughts?

gregwhitworth:
This PR does the following and will resolve #80 

I have added tentative names for the slots that are outside of the defined parts. While working on the checkbox @nicholasrice brought up a valid question of why we would want to allow the author to completely replace an entire part. This is to ensure that we aren't locking in current anatomies and behaviors completely. This would allow authors to replace the light DOM with their own variation of part. For example:

```
<select>
    <div slot="button-wrapper">
       <p>I'm able to put whatever I want in here, no limitations</p>
       <div part="button">This is the button that will trigger the open state of the <select>
    </div>
    <div part="list-box">
        ... options ...
    </div>
</select>
```

This will allow our controller code that knows to look for the button part and attach necessary event listeners while allowing the author to replace only the part they truly care about. @dandclark if you can take a glance at the anatomy itself and naming I'd appreciate it. Here is the anatomy in image form:

image

### DIFF
--- a/research/src/components/anatomy.css
+++ b/research/src/components/anatomy.css
@@ -62,6 +62,10 @@
   content: "slot";
 }
 
+.component-anatomy slot:not([name=""])::before {
+  content:  attr(name) " [slot]";
+}
+
 .component-anatomy host {
   border: 5px solid black;
 }
--- a/research/src/components/select-anatomy.js
+++ b/research/src/components/select-anatomy.js
@@ -22,11 +22,7 @@ const SelectAnatomy = () => {
               <slot>
                 <div class="element">
                   <div class="anatomy-label" name="optgroup" data-slot></div>
-                  <slot>
-                    <part name="option" data-slot>
-                      <slot>Content</slot>
-                    </part>
-                  </slot>
+                  <part name="option" data-slot></part>
                 </div>
               </slot>
             </part>
--- a/research/src/components/select-anatomy.js
+++ b/research/src/components/select-anatomy.js
@@ -1,3 +1,5 @@
+/* eslint-disable */
+
 import React from 'react'
 import './anatomy.css'
 
@@ -8,26 +10,24 @@ const SelectAnatomy = () => {
       <label for="show-slots"> Show slots</label>
       <div class="component-anatomy">
         <host name="select" data-slot>
-          <slot>
-            <slot data-slot>
-              <part name="button" data-slot>
-                <slot>Currently selected option</slot>
-              </part>
-            </slot>
-            <slot data-slot>
-              <part name="listbox" data-slot>
-                <slot>
-                  <div class="element">
-                    <div class="anatomy-label" name="optgroup" data-slot></div>
-                    <slot>
-                      <part name="option" data-slot>
-                        <slot>Content</slot>
-                      </part>
-                    </slot>
-                  </div>
-                </slot>
-              </part>
-            </slot>
+          <slot name="button-wrapper" data-slot>
+            <part name="button" data-slot>
+              <slot>Currently selected option</slot>
+            </part>
+          </slot>
+          <slot name="listbox-wrapper" data-slot>
+            <part name="listbox" data-slot>
+              <slot>
+                <div class="element">
+                  <div class="anatomy-label" name="optgroup" data-slot></div>
+                  <slot>
+                    <part name="option" data-slot>
+                      <slot>Content</slot>
+                    </part>
+                  </slot>
+                </div>
+              </slot>
+            </part>
           </slot>
         </host>
       </div>
--- a/research/src/components/select-anatomy.js
+++ b/research/src/components/select-anatomy.js
@@ -12,7 +12,7 @@ const SelectAnatomy = () => {
         <host name="select" data-slot>
           <slot name="button-wrapper" data-slot>
             <part name="button" data-slot>
-              <slot>Currently selected option</slot>
+              <slot name="selected-value">Currently selected value</slot>
             </part>
           </slot>
           <slot name="listbox-wrapper" data-slot>
--- a/research/src/components/select-anatomy.js
+++ b/research/src/components/select-anatomy.js
@@ -10,12 +10,14 @@ const SelectAnatomy = () => {
       <label for="show-slots"> Show slots</label>
       <div class="component-anatomy">
         <host name="select" data-slot>
-          <slot name="button-wrapper" data-slot>
+          <slot name="button-container" data-slot>
             <part name="button" data-slot>
-              <slot name="selected-value">Currently selected value</slot>
+              <slot name="selected-value">
+                <part name="selected-value">Currently selected value</part>
+              </slot>
             </part>
           </slot>
-          <slot name="listbox-wrapper" data-slot>
+          <slot name="listbox-container" data-slot>
             <part name="listbox" data-slot>
               <slot>
                 <div class="element">
--- a/research/src/components/specimens.js
+++ b/research/src/components/specimens.js
@@ -1,4 +1,3 @@
-import _ from 'lodash'
 import React from 'react'
 import { getImagesForComponentConcept } from '../sources'
 import Image from './image'
@@ -8,7 +7,7 @@ const Specimens = ({ component, conceptName, showDescriptions }) => {
 
   return (
     <div style={{ display: 'flex', flexWrap: 'wrap', border: '1px solid #ccc' }}>
-      {images.map(image => {
+      {images.map((image) => {
         const hasOverrideName = image.name !== image.openUIName
 
         return (
--- a/research/src/styles/spec.css
+++ b/research/src/styles/spec.css
@@ -13,13 +13,3 @@
     margin-right: .5em;
     font-weight: bold;
 }
-
-.link {
-    display: inline-block;
-    font-family: sans-serif;
-    color: #ccc;
-}
-
-.link::before {
-    content: "#";
-}


dandclark:
Looking at the diagram, I'm not sure it makes sense to have the unnamed slots inside the button, the optgroup, and the option.  An unnamed slot, at least in terms of the HTML `<slot>` element, receives the first-level children of the shadow host that are not explicitly assigned to a slot.  So if there is more than one unnamed slot, all the ones after the first one are not addressable and might as well not be there.

If we want to allow users to slot content into the children of the button without replacing the button itself, we could make that one a named slot.  I'm not sure if this is a common enough scenario that we'd want to add complexity to allow it though.  It brings up questions like, what happens to the slotted content when the user chooses a different value in the dropdown?  Does the platform just throw away the user content to update the button content to the new chosen value?  Or does the user content always win?  It could be simpler to make the user always replace the full button if they want to change it.

The spot where an unnamed slot is appropriate is the one directly under the listbox, so stuff like this can continue to work:
```html
<select>
  <option> thing</option>
  <option> other thing</option>
</select>
```
Where the options get placed in the unnamed slot under the listbox.  But if there are other unnamed slots this breaks, e.g. in the anatomy proposed above the options would get slotted as children of the native button part.

dandclark:
Bikeshedding names; I don't necessarily think these are better but just throwing them out there as alternatives to consider.
- `button-container` and `listbox-container`
- If we do have named slots to replace the contents (see my prev comment), `button-outer` and `button-inner`?  Or `button-entire` and `button-contents`?
- Simply `button` and `listbox`?

My concern with `button-wrapper` etc. is that to me it implies that replacing it is just replacing a wrapper to swap in styles or something, but not replacing the button itself.  But that might just be my own reading of it; I'd be curious as to how web developers actually tend to name these things.

We can always just run with what you've got and open an Issue to consider a naming scheme that will be broadly applicable.

gregwhitworth:
> An unnamed slot, at least in terms of the HTML <slot> element, receives the first-level children of the shadow host that are not explicitly assigned to a slot. So if there is more than one unnamed slot, all the ones after the first one are not addressable and might as well not be there.

Ugh, I'd really love to avoid naming every slot that exists but yes a `content` based slot makes sense from a naming perspective. With regards to your question regarding allowing them to change out their content and how that interacts with the controller code I think those are valid questions to ask and we shouldn't dismiss them because they're hairy questions. 

Wouldn't these questions still need to be answered even if they completely change the button as we'll need a location that represents where the controller code will be putting the currently selected value. Because the button will be where the event listener for changing the open state will be tied, not necessarily where the selected value will be displayed (if at all).

dandclark:
> Wouldn't these questions still need to be answered even if they completely change the button as we'll need a location that represents where the controller code will be putting the currently selected value. Because the button will be where the event listener for changing the open state will be tied, not necessarily where the selected value will be displayed (if at all).

You're right, regardless of the slot layout we have to consider these questions.  One possible answer is that if the whole button is replaced, no controller code is applied for changing the button contents to the currently selected value.  The user would to supply their own code if they want that behavior, or leave the native button in place.

One reason this might be desirable is if the user wants to have something in addition to the raw text of the selected value in their button.  E.g. this multi-select from https://atlaskit.atlassian.com/packages/core/select that has styled boxes around the values, and close buttons for each.

image

It would be counterproductive if the developer building this has to fight with native controller code to get the exact content that they want in the button.

gregwhitworth:
> One possible answer is that if the whole button is replaced, no controller code is applied for changing the button contents to the currently selected value.

Not sure how I feel about this because entire premise of having "known" parts was so that the controller code wouldn't need to be re-written. Leaving out the more complex multi-select, just take a split button:

```
<select>
     <div slot="button-container">
          <div part="value-text">Currently selected value</div>
          <button part="button"><svg ...></svg></button>
     </div>
        .....
</select>
```

Here I haven't changed the model at all solely the view and the controller code should still work. Granted I've shown here that we may need a part called `value-text` or similar to allow the controller code to know where to put the value(s). I think this is something that we should really carefully consider. @chrisdholt @levithomason @EisenbergEffect what do you all do? Do you all simply lock in the `<select>` and extensibility isn't possible as we're actively discussing or do you all allow a user of your component library to modify parts?

gregwhitworth:
> However, make not mistake, when the developer did this level of swaping, it was on them to do more manually wire up between the controller code and their new view slot replacement.

This is what I was trying to determine. @dandclark I think this is a good indication that this will allow us to spec that `<T>-container` slots will require controller code to be written by the author. This however does make me feel like we need a slot for the value as that is the most common use case and I don't want every user that wants to simply adjust where the button and value are to have to re-write all of that controller code. @dandclark thoughts?

dandclark:
Clarifying question:
Say we have two slots for the button:
- Slot surrounding the button that allows the dev to replace the whole thing.  In this case, they're responsible for writing their own controller code to update the value displayed in the button.
- Slot inside the button that allows the dev to replace the value.  Native controller code still runs to update the value displayed in the button when the user picks an `<option>`.

In the second option, wouldn't the native controller code just overwrite whatever the user put in the slot?  Or would the point of using that slot just be to provide an initial value, like an `<input>` placeholder text, that will get overwritten by the native controller code when the user picks a value?

gregwhitworth:
> In the second option, wouldn't the native controller code just overwrite whatever the user put in the slot?

Yes, that is precisely my expectation. I think what this ultimately boils down to is more meta, what makes a select. We can have a reasonable default with a definition of how it all is tied together and even how they can prevent default on that specific piece. But I should be able to accomplish the same model with the same slots and parts. The replacement of the entire part requiring new code is fine as long as we then separate the slot for the value and part that is listening for the events to modify the popup state. So because of this we'll need a slot for `select-value`? Let me take a quick stab at updating the anatomy and post a screenshot of it.

EisenbergEffect:
Not sure what is possible with the underlying C++, but I don't think you can nest shadow dom slots. So, when working out the anatomy for the button and value, to stay shadow dom spec compliant, I think you will need to layer the html with the value sitting on top of the button.

gregwhitworth:
Ok, so I've adjusted it like so:

```
<slot name="button-wrapper" data-slot>
       <part name="button" data-slot>
          <slot name="selected-value">Currently selected value</slot>
      </part>
</slot>
```

The event for adjusting the `open` state will be attached to the button part which then you could do something like this as a consumer to create a split button since we named this slot:

```
<select>
      <div part="button" style="display: flex;">
              <div slot="selected-value">Option One</div>
              <div><svg>... arrow down svg ..</svg></div>
      </div>
      <div part="list-box">
           <option>Option One</option>
       </div>
</select>

<script>
       // We only want the arrow to be propagating click events to create a split button
      // style UI/UX
      let value = this.shadow.querySelector('select slot[selected-value]')
      value.addEventListener('click', (e) => { e.stopPropagation() })
</script>
```

This will allow the adjustment of the content and button independent, allowing the controller code to function accordingly. Now we'll need to define how the content gets placed into the slot (eg: `textContent`, `value`, etc). Because technically you could allow an author to put an input in to the slot and have the content appear but that will vary based on what the author inserts into the slot.

dandclark:
> Not sure what is possible with the underlying C++, but I don't think you can nest shadow dom slots. So, when working out the anatomy for the button and value, to stay shadow dom spec compliant, I think you will need to layer the html with the value sitting on top of the button.

@EisenbergEffect I might be misunderstanding what you're saying here but nested `<slot>` elements work as I would expect them to in HTML.  Here's some examples I put together a couple weeks ago that shows how they can be nested to allow flexible replacement of the default content: https://jsfiddle.net/dandclark_msft/f8vmk6np/15/.  This runs fine in Chrome/Edge.

@gregwhitworth  So the expected behavior in your example is that clicking on the arrow icon would *not* open/close the popup, but clicking anywhere else in the popup would open/close?

I'm not sure the example quite works as written.  The `<div part="button" style="display: flex;">` and `<div part="list-box">` won't be slotted into the right parts of the shadow DOM without specifying the `slot` attribute.

The line `let value = this.shadow.querySelector('select slot[selected-value]')` begs the question of whether the shadow DOM should be `mode: 'open'`.  I think it probably should not be.  Leaving it open allows developers to take dependencies on the exact content of the browser's default HTML.  We're standardizing the slots and their relative positions to each other, but not the exact HTML content they use to build the default APIs, so this could be a problem.

If the shadow DOM is closed, this scenario can still be achieved by referencing the slotted light DOM elements directly instead of going through the shadow root.

EisenbergEffect:
@dandclark Awesome! For whatever reason I thought I had tried that in the past and it didn't work. This is excellent. Thanks for teaching me something new :)

gregwhitworth:
> @gregwhitworth So the expected behavior in your example is that clicking on the arrow icon would not open/close the popup, but clicking anywhere else in the popup would open/close?

The inverse, only clicking the arrow should propagate the click event.

> The line let value = this.shadow.querySelector('select slot[selected-value]') begs the question of whether the shadow DOM should be mode: 'open'. I think it probably should not be. 

Yeah, I snagged this from my own web component which does have it open. But yes I'd expect that the author would use their light DOM content that is slotted into `selected-value`.

So to hone in on my key question there, if we provide a slot for `selected-value` that should enable the controller code to continue to function for the `button` part and the `selected-value` which will allow common scenarios that authors desire to do that will re-use the default `<select>` model. Are you in agreement with that anatomy?

I'd like to note that the example that you provided with the input and tag type items is actually a control known as a combobox, not a select. Many component libraries now combine them, which is fine but my expectation all along was to keep the model of a `<select>` the same as it is today and introduce a new element called `<combobox>` which is effectively what you get with an input tied to a datalist.

dandclark:
> The inverse, only clicking the arrow should propagate the click event.

Oops, yes, I see now.

> So to hone in on my key question there, if we provide a slot for `selected-value` that should enable the controller code to continue to function for the `button` part and the `selected-value` which will allow common scenarios that authors desire to do that will re-use the default `<select>` model. Are you in agreement with that anatomy?

For the reasons stated above I don't think the example in your last comment quite works as intended -- the button and the listbox are not slotted into the shadow DOM -- and I'm having a hard time coming up with code that would quite work for this scenario with that anatomy.  The problem is that if I try something like 

```html
<select>
      <div slot="button" part="button" style="display: flex;">
              <div slot="selected-value">Option One</div>
              <div><svg>... arrow down svg ..</svg></div>
      </div>
     <!-- lisbox/options go here --> 
</select>
```

Then the nested `<slot name="selected-value">` in the initial anatomy has been swapped out by the user-provided `<div slot="button">` and its subtree.  So if the controller code is doing something like the equivalent of 
`window.addEventListener("change", () => {nativeShadowRootOfTheSelect.querySelector("slot[name=selected-value]").innerText = "the new value"}` then this no longer works because there is no such <slot> element; it's been swapped out by the user content.  Furthermore the slot attribute on `<div slot="selected-value">` is confusing because that div isn't actually getting slotted in anywhere, as the `<slot>` it would have gone into has already been replaced when the `<div slot="button">` replaced the `<slot name="button">`'s subtree.

The way I'd been thinking about the overall controls customization approach is something like this:
- slots are used by the developer to swap their custom markup into useful places in the control shadow tree.  Controller code doesn't do anything interesting with slots; they just behave like plain `<slot>` elements.
- Parts are what the controller code hooks up to.

Thinking about it that way I'd amend the anatomy to this:

```html
<slot name="button-wrapper" data-slot>
       <part name="button" data-slot>
          <slot name="selected-value"><part name="selected-value">Currently selected value</part></slot>
      </part>
</slot>
```

where the controller code would do something roughly equivalent to 
`window.addEventListener("change", () => {nativeShadowRootOfTheSelect.querySelector("part[name=selected-value]").innerText = "the new value"}`.

Then, your example would look like this:
```html
<select>
      <div slot="button" part="button" style="display: flex;">
              <div part="selected-value">Option One</div>
              <div><svg>... arrow down svg ..</svg></div>
      </div>
     <!-- listbox/options go here --> 
</select>
```

In this case I've replaced the whole button, but my `<div part="selected-value">` will have its value updated automatically by the controller code.
Alternatively I could do this if I wanted to swap out just the text of the default button and manage it myself without the the controller code touching it (since it's not marked as a part):
```html
<select>
      <div slot="selected-value" style="display: flex;">
              My custom text
      </div>
     <!-- listbox/options go here --> 
</select>
```

Is this making sense?  If not I can try to code some of this up as samples like in https://jsfiddle.net/dandclark_msft/f8vmk6np/15/.

> I'd like to note that the example that you provided with the input and tag type items is actually a control known as a combobox, not a select. Many component libraries now combine them, which is fine but my expectation all along was to keep the model of a `<select>` the same as it is today and introduce a new element called `<combobox>` which is effectively what you get with an input tied to a datalist.

Thanks for clarifying, that sounds good to me.

gregwhitworth:
> slots are used by the developer to swap their custom markup into useful places in the control shadow tree. Controller code doesn't do anything interesting with slots; they just behave like plain <slot> elements.
> Parts are what the controller code hooks up to.

Agree 100%. And this was my initial mistake in having QSA focus on the slot which parts will allow us to avoid, again I think the confusion I caused was pulling in content from my other WC. We're on the same page. That anatomy looks good to me, I'll update the PR in the morning. Ultimately I want to get this in a decently solid shape so that we can start building it out and ensuring that we can build the usecases. Because if the anatomy doesn't allow it then we'll be voiding some of our key principles. I also think it will allow us to get devs to start trying to break it.

gregwhitworth:
Ok, I've updated the anatomy for the button and also changed the slot names from `wrapper` to `container`. Here is an image of the parts/slots that you'll see on the spec to make it a bit easier to discuss:

image

The actual DOM can be seen in the `select-anatomy.js` file.

gregwhitworth:
Thanks @dandclark - here are my responses

1. This is trying to represent what actually does occur, the `<option>`(s) do end up inside of the `<optgroup>` and the content provided inside of an `<option>` is likewise placed inside of its content box, so the slots are there. Ultimately I guess this comes down to if we want to represent the content box that is automatically created for most HTML elements. EG: `<div><slot></div>`. I honestly don't have a strong opinion here.
2. That's what I intended for the [structure](https://open-ui.org/components/select#structure-1) section but am fine incorporating it into the anatomy box although I already think it's a bit busy, especially for `<select>`.

For now, to land this PR I'll remove those slots as I'm not married to them.

dandclark:
> 2. That's what I intended for the [structure](https://open-ui.org/components/select#structure-1) section but am fine incorporating it into the anatomy box although I already think it's a bit busy, especially for `<select>`.

I totally missed the structure section, sorry -- I think that's great as-is and given that it's right next to the anatomy box i don't think it needs to be incorporated into it.

PR #103, opened 2020-06-04
Initial recommendations for part events

gregwhitworth:
This PR lands specific recommendations of events for `<select>` parts. This effectively should represent the "controller" code if we utilize the MVC approach for controls. One aspect that I didn't get in there that @dandclark and I discussed a little bit is blur. My initial thought is to put a blur event on the `<select>` that will change the `open` state. Please let me know your thoughts and if I'm missing any events.


research/src/pages/select/select.proposal.mdx
```diff
@@ -116,11 +116,60 @@ children:
...
+
+| Event    | Behavior                                                                                         | Impacts          |
+| -------- | ------------------------------------------------------------------------------------------------ | ---------------- |
+| `change` | Updates the textContent property of the selected-value slot with the value property text content | textContent prop |
+
+### part button
+
+| Event          | Behavior                           | Impacts                                                                  |
+| -------------- | ---------------------------------- | ------------------------------------------------------------------------ |
+| `click`        | Toggles the open state on the host | [open](#open-state) state                                                |
```
chancestrickland, line 131: This may have already been discussed elsewhere, but should the open state not be toggled on `pointerdown`? I believe different OS's native counterparts handle this differently.

nicholasrice: Confirmed OSX v10.15.4 opens on `pointerdown`

research/src/pages/select/select.proposal.mdx
```diff
@@ -116,11 +116,60 @@ children:
...
+| `keyup(space)` | Toggles aria-expanded              | [aria-expanded](#https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
+| `keyup(enter)` | Toggles the open state on the host | [open](#open-state) state                                                |
+| `keyup(enter)` | Toggles aria-expanded              | [aria-expanded](#https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
+
+### part listbox
+
+| Event             | Behavior                                                                                                               | Impacts                                                                  |
+| ----------------- | ---------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------ |
+| `keyup(down key)` | Moves focus to the next `<option>` in the listbox and sets the                                                         | focus                                                                    |
```
nicholasrice, line 144: Similar vein to the above, but today OSX moves focus on *keydown*.

research/src/pages/select/select.research.behavior.mdx
```diff
@@ -263,53 +263,26 @@ Currently this is testing the built-in browser based `<select>` control
...
-- [possible focus trapping proposal following popup investigation]
+| Library                       | Behavior                                         |
+| ----------------------------- | ------------------------------------------------ |
+| Chrome/Edge/Firefox (Windows) | Selects the currently highlighted option         |
+| Safari (Catalina)             | Retains the previously selected option           |
+| Evergreen Combobox            | Removes all selected options to default settings |
+| Fluent UI                     | Retains the previously selected option           |
+| Ant Design                    | Retains the previously selected option           |
+| Atlas Kit                     | Retains the previously selected option           |
+| Carbon                        | Selects the currently highlighted option         |  | Fluent UI | Retains the previously selected option |
```
nicholasrice, line 284: nit: formatting

research/src/pages/select/select.research.behavior.mdx
```diff
@@ -263,53 +263,26 @@ Currently this is testing the built-in browser based `<select>` control
   <option value="ZW">Zimbabwe</option>
 </select>
 
-## States:
+## Pressing escape
 
-**Open** - this is when the popup is open and the state is placed on the select.
+While adding in the events I noticed an interesting user experience that I
```
nicholasrice, line 268: With the above spec, where arrow up/down *sets focus but does not select* I'd expect escape to revert to the previous selection, because no selection operation occurred prior to escaping the modal.

FWIW [ARIA](https://www.w3.org/TR/wai-aria-practices/#listbox_kbd_interaction) describes the selection behavior as optional in single-select listboxes.

gregwhitworth: > FWIW ARIA describes the selection behavior as optional in single-select listboxes.

And I'd like to make it so that it isn't optional, I'll add this to the issue and will probably loop in the aria folks.

gregwhitworth: Issue is here: https://github.com/WICG/open-ui/issues/102

research/src/pages/select/select.proposal.mdx
gregwhitworth: is this for all `keyup` events I listed or solely focus movement? It does seem on Windows it to be `keydown` for moving focus. That said, on selection with pointer it is on `pointerdown` but the state of open is not changed until `pointerup`.

chancestrickland:
When handling collisions, MacOS uses an arrow indicator that scrolls the list on `mouseover` vs. a traditional scrollbar. Not sure if that's worth to considering as a part, but this is the current experience with a regular `select` element.
image

gregwhitworth:
> Not sure if that's worth to considering as a part

To determine the parts we're primarily trying to gather the most common paradigms. This is an interesting question however and I think this actually raises the question of defining scrollers in general, not just this specific scenario. Scrollers have parts and to some extent have been defined. It does seem like something we should consider but not block this PR on.

research/src/pages/select/select.proposal.mdx
dandclark: One important difference with `keydown` vs `keyup` is that holding down the arrow key and having it spin quickly through the options will only work with `keydown.`

nicholasrice: Yea this seems to be for all keyboard events. I can't test the selection keys because selection is happening on focus, but the reset of them are happening on keydown and not up.

gregwhitworth: Thanks, for now then I'll move them to `keydown` until someone can prove otherwise.

research/src/pages/select/select.proposal.mdx
```diff
@@ -46,7 +46,7 @@ then select the appropriate one for you.
...
 | `validity`                                                                                              | `ValidityState`         |                                       | Represents the current validity state of the control                                                                                                     |     |
-| `value`                                                                                                 | `string`                | `""`                                  | Returns the `value` property of the first selected `<option>` if there is one, otherwise an empty string                                                 |     |
+| `value` <a id="value"></a>                                                                              | `string`                | `""`                                  | Returns the `value` property of the first selected `<option>` if there is one, otherwise an empty string                                                 |     |
```
dandclark, line 49: What's the purpose of this `<a>`?  I didn't see any visible change on the deploy preview.

research/src/pages/select/select.proposal.mdx
```diff
@@ -116,11 +116,55 @@ children:
...
+| `keydown(enter)`      | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` on the `<select>`         | [open](#open-state) state                                                |
+| `keydown(enter)`      | If the `<select>` does **not** have the `multiple` attribute then toggle the aria-expanded attribute on the `<select>` | [aria-expanded](#https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
+| `keydown(space)`      | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` on the `<select>`         | [open](#open-state) state                                                |
+| `keydown(space)`      | If the `<select>` does **not** have the `multiple` attribute then toggle the aria-expanded attribute on the `<select>` | [aria-expanded](#https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
+| `keydown(escape)`     | Toggles the `open` state of the `<select>`                                                                             | [open](#open-state) state                                                |
```
dandclark, line 152: There should be a row for `keydown(escape)` to toggle `aria-expanded` too, right?

A refactoring that we could consider (not necessarily part of this change) could be another table that ties the value of certain states to the value of certain `aria` semantics.  E.g. declaring that the value of the `open` state is always reflected in the value of `aria-expanded` would allow us to delete a lot of these rows.

research/src/pages/select/select.proposal.mdx
```diff
@@ -116,11 +116,55 @@ children:
...
+| `keydown(space)`      | Changes the `selected` state of the current `<option>` and updates the `<select>` value property                       | [selected](#selected) prop <br/> [value](#value) prop                    |
+| `keydown(enter)`      | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` on the `<select>`         | [open](#open-state) state                                                |
+| `keydown(enter)`      | If the `<select>` does **not** have the `multiple` attribute then toggle the aria-expanded attribute on the `<select>` | [aria-expanded](#https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
+| `keydown(space)`      | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` on the `<select>`         | [open](#open-state) state                                                |
+| `keydown(space)`      | If the `<select>` does **not** have the `multiple` attribute then toggle the aria-expanded attribute on the `<select>` | [aria-expanded](#https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
```
dandclark, line 151: These aria-expanded links are broken by the '#' in the beginning (`#https://www.w3.org...`).  The browser tries to navigate it as a bookmark on the page.

research/src/pages/select/select.proposal.mdx
```diff
@@ -116,11 +116,55 @@ children:
...
+| `keydown(up key)`\*   | Moves focus to the previous `<option>` in the listbox                                                                  | focus                                                                    |
+| `keydown(enter)`      | Changes the `selected` state of the current `<option>` and updates the `<select>` value property                       | [selected](#selected) prop <br/> [value](#value) prop                    |
+| `keydown(space)`      | Changes the `selected` state of the current `<option>` and updates the `<select>` value property                       | [selected](#selected) prop <br/> [value](#value) prop                    |
+| `keydown(enter)`      | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` on the `<select>`         | [open](#open-state) state                                                |
+| `keydown(enter)`      | If the `<select>` does **not** have the `multiple` attribute then toggle the aria-expanded attribute on the `<select>` | [aria-expanded](#https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
```
dandclark, line 149: I'm thinking that `aria-expanded` should be set on the `button` part, not the `<select>` itself.  I think that should generally be the case with these `aria-*` semantics.  It's the parts in the shadow, not the host element, that are getting rendered, so they are the ones that need the ARIA fixup.

research/src/pages/select/select.proposal.mdx
```diff
@@ -116,11 +116,55 @@ children:
...
+
+### part option
+
+| Event       | Behavior                                                                                                       | Impacts                                               |
+| ----------- | -------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------- |
+| `click`     | Changes the `selected` state of the current `<option>` and updates the `<select>` value property               | [selected](#selected) prop <br/> [value](#value) prop |
+| `click`     | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` on the `<select>` | [open](#open-state) state                             |
+| `pointerup` | Changes the `selected` state of the current `<option>` and updates the `<select>` value property               | [selected](#selected) prop <br/> [value](#value) prop |
+| `pointerup` | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` on the `<select>` | [open](#open-state) state                             |
```
dandclark, line 161: We should probably have rows for `aria-expanded` here too, unless we just tie `open` to `aria-expanded` per my suggestion in another comment.

research/src/pages/select/select.proposal.mdx
```diff
@@ -116,11 +116,55 @@ children:
...
+| -------- | ------------------------------------------------------------------------------------------------ | ---------------- |
+| `change` | Updates the textContent property of the selected-value slot with the value property text content | textContent prop |
+
+### part button
+
+| Event            | Behavior                           | Impacts                                                                  |
+| ---------------- | ---------------------------------- | ------------------------------------------------------------------------ |
+| `click`          | Toggles the open state on the host | [open](#open-state) state                                                |
+| `click`          | Toggles aria-expanded              | [aria-expanded](#https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
+| `pointerup`      | Toggles the open state on the host | [open](#open-state) state                                                |
```
dandclark, line 133: Having both `click` and `pointerup` in the table is confusing me a bit for two reasons:
1) If I mouse-click on an element both `click` and `pointerup` are fired.  This suggests to me that both the `click` and the `pointerup` rows of the table are "fired", meaning that everything toggles twice and nothing really happens (I might just be taking this document more literally than is intended, but it would be nice to clarify this somehow).
2) `pointerup` doesn't seem like the right choice because of the following behavior: If I click and hold the mouse button down somewhere else on the page, drag the mouse over the button part, and then release the mouse, I'll get a `pointerup` -- as opposed to `click` which also requires the mouse-down to have been on the same element.  This seems wrong.  The native `<select>` in Chrome on Windows and Mac opens on mousedown, and `<input type="button">` is usually used with `click`.  It seems to me like we should stick with one of those unless there is some compelling reason I missed.

research/src/pages/select/select.proposal.mdx
gregwhitworth: it's to provide an anchor for linking to, for example: `[selected](#selected) prop <br/> [value](#value) prop`

dandclark: Ah, I didn't see the referencing anchor, thanks.

research/src/pages/select/select.proposal.mdx
```diff
@@ -126,39 +126,36 @@ children:
...
+| `keydown(up key)`   | Moves focus to the previous `<option>` in the listbox                                                                       | focus                                                                   |
+| `keydown(enter)`    | Changes the `selected` state of the current `<option>` and updates the `<select>` value property                            | [selected](#selected) prop <br/> [value](#value) prop                   |
+| `keydown(space)`    | Changes the `selected` state of the current `<option>` and updates the `<select>` value property                            | [selected](#selected) prop <br/> [value](#value) prop                   |
+| `keydown(enter)`    | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` on the `<select>`              | [open](#open-state) state                                               |
+| `keydown(enter)`    | If the `<select>` does **not** have the `multiple` attribute then toggle the `aria-expanded` attribute on the `<select>`    | [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
```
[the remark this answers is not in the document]
gregwhitworth: GAH! I thought I caught all those little changes. Thanks for the eagle eye @dandclark

PR #108, opened 2020-06-09
image: Don't use the non-standard zoom property to shrink images.

emilio:
Use the standard srcset instead.


gregwhitworth:
What's the reason for the change, I mean it has no impact on the output since all of these are only pulling in a single source.

PR #124, opened 2020-07-09
Updating Bootstrap and Primer documentation to include their use of Flex

sopranopillow:
Updates Bootstrap and Primer to show the Flex component.

### DIFF
--- a/research/src/sources/primer.json5
+++ b/research/src/sources/primer.json5
@@ -97,6 +97,10 @@
           "description": "Use .hidden-text-expander to indicate and toggle hidden text."
         }
       ]
-    }
+    },
+    {
+      "name": "Flex",
+      "url": "https://primer.style/components/Flex"
+    },
   ]
 }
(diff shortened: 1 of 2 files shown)


research/src/sources/primer.json5
```diff
@@ -97,6 +97,10 @@
           "description": "Use .hidden-text-expander to indicate and toggle hidden text."
         }
       ]
-    }
+    },
+    {
+      "name": "Flex",
```
khmakoto, line 102: Is there a definition you could add here like the one `Button` has above?

sopranopillow: I can add something like:
`Flex provides a more efficient way to lay out, align and distribute space among items in a container, even when their size is unknown and/or dynamic` based on: https://css-tricks.com/snippets/css/a-guide-to-flexbox/ under the background section.

khmakoto:
It would probably be good to update the title of this PR to something like `Updating Bootstrap and Primer documentation to include their use of Flex`.

PR #125, opened 2020-07-10
Update sync split-button example with changes in Explainer doc

dandclark:
Update select implementation so that the controller code is applied to the button and listbox parts directly, rather than using the `<slot>`s.

Refactor so that the same code adds controller functionality to both the default parts and to the user-replaced parts, which should make it easier to keep these in-sync.

In the split-button example:
* Remove the preventDefault that was keeping mouse clicks from working on the left half of the button, and achieve this functionality instead by moving `part='button' to the arrow element.
* Add slot/part to listbox.
* Fix typo in CSS class.


components/src/select/demos/bootstrap-split/index.html
```diff
@@ -108,25 +108,18 @@ <h1>Select - Open UI Bootstrap Split Button Styles</h1>
        support for the option parts.</p>
 
         <oui-select>
-            <div slot="button-container" part="button">
+            <div slot="button-container">
               <div class="select-label">Primary</div>
-              <div class="arrow" aria-label="toggle dropdown"></div>
+              <div class="arrow" part="button"></div>
```
gregwhitworth, line 113: By moving this here - this results in the `part="button"` receiving focus which breaks the focus ring solution of the `slot=button-container` working as expected. This is where things can get interesting because you'll want focus to land on the container and this *will* invoke the popup which applied the focus styles. So it the container behaves with keyboard/focus as part="button" but the text is not clickable. This is why I had the slot being the part="button". You can see the example dropdown here: https://getbootstrap.com/docs/4.0/components/dropdowns/

It's about halfway down and blue with the work primary. I think you'll need to add the script code back and move part back to the slot.

dandclark: I pushed a change to give the `<div class="select-label">Primary</div>` a tabindex, and updated the CSS for the focus outline.  The behavior should now match the behavior at https://getbootstrap.com/docs/4.0/components/dropdowns/#split-button-dropdowns:
1. The "Primary" and the down-arrow halves of the button are both individually focusable by click and by tabbing to them.
1. Each half has their own focus outline; there is never a focus outline drawn around the entire button.
1. Invoking the "Primary" half does nothing (presumably in a real use case the dev would  hook up event handlers to invoke the Primary action).

Note this is distinct from the "The best part is you can do this with any button variant, too" examples just above the https://getbootstrap.com/docs/4.0/components/dropdowns/#split-button-dropdowns section.  Those look similar (in that they have down-arrows icons drawn on them) but they aren't split-buttons, so the whole button is focusable as a single item.

PR #128, opened 2020-07-17
Fixing component name matrix regex to correctly group items that end with 'e' or 'er'

khmakoto:
Components whose names ended with "e" or "er" where not grouped together with their plural counterpants because of an error in the regex. This PR fixes that issue.

__Before:__

image

image

__After:__

image

image


gregwhitworth:
Great change!

levithomason:
🙇‍♂️ Thanks @khmakoto

PR #130, opened 2020-07-17
fix(image): fix property typo

assuncaocharles:
property `srcset` throws console error since the expected property is `srcSet`

Screenshot 2020-07-17 at 09 35 01


assuncaocharles:
@gregwhitworth just pinging you so you can have a look in this one also :) 

Thank you

PR #132, opened 2020-07-17
fix(layout): conditionally frontmatter access

assuncaocharles:
Page 404 there's no frontmatter what is breaking the build in master and the access to the custom 404 page.


assuncaocharles:
Hey @gregwhitworth 

Can we merge this one to have CI passing in the other PR's?

PR #135, opened 2020-07-20
Clarified vision statement

gregwhitworth:
Based on our discussion, I have created a PR that tries to provide clarity for Open UI.


gregwhitworth:
I'm going to land this since this keeps the spirt of the current vision but is simplified down and Levi has provided feedback offline.

PR #138, opened 2020-07-21
Create section defining 'light dismiss' behavior of `<select>` listbox part

### LINKED ISSUE #137: Make <select> change to closed state when focus leaves listbox
Today I set out to make what I thought would be a simple improvement: update the `<select>` anatomy such that the `listbox` part has a `blur` event handler that switches the `open` state to `false`.  This achieves the classic `<select>` behavior where clicking outside the listbox (or anything else that causes focus to jump elsewhere on the page) causes it the listbox to be collapsed.

It turns out that `blur` doesn't quite work; it doesn't bubble, so when focus goes from an option in the listbox to something else, the listbox won't get the event.  Note: the listbox itself normally isn't focusable, only its contents.

An better alternative is `focusout`.  This one fires when anything in the subtree of the element loses focus.  This is a bit too noisy by default, since it will fire when focus switches from one `<option>` to another, but we can scope it thusly (`relatedTarget` for focusout is the element that is receiving focus):
```js
if (!event.relatedTarget || // event.relatedTarget is null if something non-focusable was clicked and focus is going back to the <body>
    !listbox.contains(event.relatedTarget)) { // Close the <select> if the focus target isn't in the listbox's subtree.
// set <select> `open` state to false
}
```
This mostly works except for the case when something non-focusable is clicked inside the listbox.  In the context of `focusout` this is indistinguishable from when something non-focusable is clicked outside the listbox (`relatedTarget == null` either way).  So, it will cause the `<select>` to close, which seems wrong.  Non-focusable things would include `<optgroup>` headers and dividers like the one in [Bootstrap buttons](https://getbootstrap.com/docs/4.0/components/dropdowns/#split-button-dropdowns) (the spacer between Option Three and Option 4):

image

For the native `<select>`, clicking an `<optgroup>` caption does not cause the `<select>` to close in at least Chrome/Edge/Firefox.  This matches my expectation.  With Bootstrap buttons, clicking the divider actually *does* cause it to collapse, which is not what I would expect and might actually be due to the implementation difficulties that I'm outlining.

So, how to denote this behavior in the `<select>` anatomy?  I was hoping that we could continue to express behavior in terms of JS event handlers, but I'm not sure that is possible here without invoking hacks such as using e.g. `mouseover`/`mouseout` or watching `click` event bubbling in the listbox to alter the behavior of the `focusout` handler somehow such that it won't trigger a close for clicks inside the popup.

Thoughts? @gregwhitworth @melanierichards Native light-dismiss for the listbox popup should make all this "just work", but there is still the question of how to describe this behavior in the `<select>` anatomy tables if this is not straightforwardly possible to implement in terms of JS events.

Or maybe it's fine to just collapse the `<select>` when something non-focusable inside the listbox is clicked.  It doesn't seem to be a blocker for Bootstrap.

Or maybe (hopefully) I'm missing a more straightforward way to achieve this behavior. :)

dandclark:
Create a new section to define "light dismiss" behavior for the `<select>` listbox part, and reference the new section from the Open State Section.

Ideally we would describe this behavior in terms of a `blur` or `focusout` event, but it turns out that both of these are not adequate for the task; see #137.  To achieve this behavior today would require a combination of multiple JS events to prevent the listbox from collapsing when non-focusable content inside the listbox is clicked.  In the future we could consider fleshing out the light dismiss section to clearly specify this combination, or reference a new platform primitive if one is introduced to make this easier.

Closes #137 .

### DIFF
--- a/research/src/pages/select/select.proposal.mdx
+++ b/research/src/pages/select/select.proposal.mdx
@@ -201,12 +201,17 @@ and is therefore considered to be invalid.
   - Selecting one, or more (if `multiple` is true), options by clicking or hitting the `enter` or `space` key
 
 ### Light dismiss
+
 The listbox part has "light dismiss", behavior, defined as being dismissed (removing
 the `open` state of the `<select>`, in this case) by either of the following things:
+
 - The user presses the escape key.
-- Focus moves outside of the subtree of the listbox (because of either user interaction or script),
-  *except* in the case where focus was switched to the `<body>` element due to a user invoking a
-  non-focusable element inside the subtree of the listbox.
+- A focus change occurs (because of either user interaction or script), where the focus
+  target is outside of the subtree of the listbox. This includes the case where the user
+  invokes a non-focusable element, which causes focus to switch to the `<body>`.
+  - There is one exception to this: if a user invokes a non-focusable element
+    in the subtree of the listbox, focus still moves to the `<body>`, but "light
+    dismiss" does not occur.
 
 ### Keyboard traversal
 
--- a/research/src/pages/select/select.proposal.mdx
+++ b/research/src/pages/select/select.proposal.mdx
@@ -195,11 +195,19 @@ and is therefore considered to be invalid.
 - The currently selected `<option>`, or the first if one isn't selected, should be
   visible within the listbox. This may require moving the list to accomodate listbox positioning
   and available space.
-- The user can dismiss the listbox and remove the state of `open` from the `<select>` in any of the
+- The user can dismiss the listbox and remove the state of `open` from the `<select>` in either of the
   following manners:
-  - The user clicks outside of the listbox
-  - Presses the escape key
-  - Selects one, or more (if `multiple` is true), by clicking or hitting `enter` or `space` key
+  - Triggering any of the listbox's <a href="#light-dismiss">light dismiss</a> behaviors
+  - Selecting one, or more (if `multiple` is true), options by clicking or hitting the `enter` or `space` key
+
+### Light dismiss <a class="link" href="#light-dismiss" id="light-dismiss"></a>
+The listbox part has "light dismiss", behavior, defined as being dismissed (removing
+the `open` state of the `<select>`, in this case) by either of the following things:
+- The user presses the escape key.
+- Focus moves outside of the subtree of the element with <a href="#light-dismiss">light dismiss</a>
+  (because of either user interaction or script), *except* in the case where focus was switched to the
+  `<body>` element due to a user invoking a non-focusable element inside the subtree of the element
+  with <a href="#light-dismiss">light dismiss</a>.
 
 ### Keyboard traversal
 
--- a/research/src/pages/select/select.proposal.mdx
+++ b/research/src/pages/select/select.proposal.mdx
@@ -200,7 +200,7 @@ and is therefore considered to be invalid.
   - Triggering any of the listbox's <a href="#light-dismiss">light dismiss</a> behaviors
   - Selecting one, or more (if `multiple` is true), options by clicking or hitting the `enter` or `space` key
 
-### Light dismiss <a class="link" href="#light-dismiss" id="light-dismiss"></a>
+### Light dismiss
 The listbox part has "light dismiss", behavior, defined as being dismissed (removing
 the `open` state of the `<select>`, in this case) by either of the following things:
 - The user presses the escape key.
--- a/research/src/pages/select/select.proposal.mdx
+++ b/research/src/pages/select/select.proposal.mdx
@@ -204,10 +204,9 @@ and is therefore considered to be invalid.
 The listbox part has "light dismiss", behavior, defined as being dismissed (removing
 the `open` state of the `<select>`, in this case) by either of the following things:
 - The user presses the escape key.
-- Focus moves outside of the subtree of the element with <a href="#light-dismiss">light dismiss</a>
-  (because of either user interaction or script), *except* in the case where focus was switched to the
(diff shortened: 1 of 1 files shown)


dandclark:
cc @melanierichards

research/src/pages/select/select.proposal.mdx
```diff
@@ -195,11 +195,19 @@ and is therefore considered to be invalid.
...
-- The user can dismiss the listbox and remove the state of `open` from the `<select>` in any of the
+- The user can dismiss the listbox and remove the state of `open` from the `<select>` in either of the
   following manners:
-  - The user clicks outside of the listbox
-  - Presses the escape key
-  - Selects one, or more (if `multiple` is true), by clicking or hitting `enter` or `space` key
+  - Triggering any of the listbox's <a href="#light-dismiss">light dismiss</a> behaviors
+  - Selecting one, or more (if `multiple` is true), options by clicking or hitting the `enter` or `space` key
+
+### Light dismiss <a class="link" href="#light-dismiss" id="light-dismiss"></a>
```
melanierichards, line 203: Can we add an accessible name to the anchor link? Project convention seems to be an `aria-label="Anchor"`. I think we could iterate on that so that all the links have unique identifiers, but that's out of scope for this PR :)

research/src/pages/select/select.proposal.mdx
```diff
@@ -195,11 +195,19 @@ and is therefore considered to be invalid.
...
+The listbox part has "light dismiss", behavior, defined as being dismissed (removing
+the `open` state of the `<select>`, in this case) by either of the following things:
+- The user presses the escape key.
+- Focus moves outside of the subtree of the element with <a href="#light-dismiss">light dismiss</a>
```
melanierichards, line 207: Is saying "with a light dismiss" here somewhat circular logic? Perhaps we could either nix this phrase from the description (just "Focus moves outside of the subtree of the element") or be more specific about what kind of user input behaviors cause a light dismiss.

research/src/pages/select/select.proposal.mdx
dandclark: That link is actually not visible on the page (since it's empty); it's just there to be a target for # anchor navigation.  The 
The image link to the left of the header text is apparently generated by the build, with `aria-label="Anchor"` included:
image

However it turns out that I don't actually need that empty link at all.  The `<h3>` gets the right ID if I remove it, so anchor navigation to the section still works correctly.

image

research/src/pages/select/select.proposal.mdx
dandclark: The attempted meaning was that "the element" refers to "the element that has the property of being light-dismiss" aka the listbox, but this just makes things murkier so I'll just say "listbox".

research/src/pages/select/select.proposal.mdx
dandclark: (so I removed it).

research/src/pages/select/select.proposal.mdx
```diff
@@ -195,11 +195,18 @@ and is therefore considered to be invalid.
...
-  - Presses the escape key
-  - Selects one, or more (if `multiple` is true), by clicking or hitting `enter` or `space` key
+  - Triggering any of the listbox's <a href="#light-dismiss">light dismiss</a> behaviors
+  - Selecting one, or more (if `multiple` is true), options by clicking or hitting the `enter` or `space` key
+
+### Light dismiss
+The listbox part has "light dismiss", behavior, defined as being dismissed (removing
+the `open` state of the `<select>`, in this case) by either of the following things:
+- The user presses the escape key.
+- Focus moves outside of the subtree of the listbox (because of either user interaction or script),
```
una, line 207: If the user clicks on a non-selectable element outside of the subtree, would this switch the focus? I'm not sure it would in the browser, but it should also be a means of dismissal

dandclark: Focus would switch to the `<body>` element in this case.  I've updated it to call out both cases separately:
- User clicks in non-focusable thing outside the listbox: focus moves to `<body>` and light dismiss occurs.
- User clicks on non-focusable thing inside the listbox: focus moves to `<body>`, but light dismiss does not occur.

gregwhitworth:
@dandclark I am not seeing any difference in this PR and #141 - what am I missing here?

dandclark:
This PR is trying to define light dismiss in the `<select>` anatomy.

#141 was for adding light-dismiss behavior to the demos at https://epic-cray-0896e3.netlify.app/src/select/ (minus the issue of clicking non-focusable content inside a popup).

research/src/pages/select/select.proposal.mdx
```diff
@@ -195,11 +195,23 @@ and is therefore considered to be invalid.
...
   visible within the listbox. This may require moving the list to accomodate listbox positioning
   and available space.
-- The user can dismiss the listbox and remove the state of `open` from the `<select>` in any of the
+- The user can dismiss the listbox and remove the state of `open` from the `<select>` in either of the
   following manners:
-  - The user clicks outside of the listbox
-  - Presses the escape key
-  - Selects one, or more (if `multiple` is true), by clicking or hitting `enter` or `space` key
+  - Triggering any of the listbox's <a href="#light-dismiss">light dismiss</a> behaviors
+  - Selecting one, or more (if `multiple` is true), options by clicking or hitting the `enter` or `space` key
```
gregwhitworth, line 201: Let's not block this PR on this - but based on the resolution of #171 we may need to update this

gregwhitworth (approved):

PR #141, opened 2020-07-23
Add light dismiss to listbox controller

dandclark:
Per the discussion in #137, implementing light dismiss for the `<select>` listbox is not trivial because relying on `focusout` causes the listbox to be collapsed when non-focusable content inside a listbox is clicked.  However, for our current set of examples this behavior doesn't come up except in for the split-button example, where there is a non-focusable divider.  It turns out that this closes the listbox in the real [Bootstrap split-button](https://getbootstrap.com/docs/4.3/components/dropdowns/#split-button) though, so it seems fine to match that behavior.

Aside from that issue, adding light dismiss makes the examples behave a lot more like "real" `<select>`s.

Note that this doesn't affect the filter example because it implements its own dropdown and prevents event propagation, so it would need to implement its own light dismiss.


gregwhitworth:
this will be discussed in this weeks telecon, I'll finish review following that call

PR #156, opened 2020-08-14
Support home/end keys to jump to first/last option in listbox

### LINKED ISSUE #154: Should the Home and End keys jump to the first and last <option>?
We should consider making the Home/End keys jump to the beginning/end of the `<select>`'s options list.
Works in Chromium, Firefox, and Safari `<select>`s.  Doesn't work in [bootstrap](https://getbootstrap.com/docs/4.3/components/dropdowns/#split-button) or [ant design](https://ant.design/components/select/), but perhaps it should.

dandclark:
For `<select>` `listbox`, make the home/end keys move focus to the first/last option, matching the current behavior of native browser `<select>`s.  This is enumerated as an optional best practice in https://w3c.github.io/aria-practices/#listbox_kbd_interaction.

Closes #154


dandclark:
Not much discussion on this yet, but I created the PR in case we want to do this.

chrisdholt:
@dandclark approving because I'm a firm believer we should have this :). From an Open UI standpoint, I think optionally enumerated behavior by the W3C is typically good to put into the spec: https://w3c.github.io/aria-practices/#listbox_kbd_interaction. From an implementation standpoint, I think design systems and frameworks can choose to implement this, but I think with this specifically we should recommend.

dandclark:
@chrisdholt Thanks for that link!  I was just digging through the ARIA docs trying to find something like this.  Totally agree that we should have this.

@gregwhitworth, I'm not sure if this would really be controversial enough to call for discussion in an Open-UI telecon, or if this can just land.

gregwhitworth:
> I think design systems and frameworks can choose to implement this, but I think with this specifically we should recommend.

This reminds me that we really should define what Open UI spec compliance is but I won't take this PR on that tangent. I need to open a meta issue around this. @chrisdholt

dandclark:
> Can you open an issue to discuss how to handle that divergence

Filed https://github.com/WICG/open-ui/issues/159

Issue #104, opened 2020-06-04
The scope of the select control


gregwhitworth:
In discussing the `<select>` with some people there is a desire to allow for an input to allow filtering of the available options. Some component libraries bundle this in with their select and others refer to this as a combo box. I'm going to do some research as I think it's important to ensure that the anatomy we're defining allows this extensibility if we believe it to be an expected part of the `<select>`.

chancestrickland:
Some important potential distinctions to consider while researching:

- In a native MacOS select, when the menu opens the popover is positioned directly over the top of the currently selected option, which obscures the trigger button. You likely don't want this with a combobox, as the input retains focus and the popover changes as the user continues to interact with the component (unsure if positioning should be left to spec or a consumer to customize, but the positioning will impact the spec for events in the MacOS example)
- The popover may position itself based on collisions. With a select this is a bit more straight forward as options don't typically appear/disappear while it's open. With a combobox it does, and collisions/positioning may change as a result. In both cases collisions may change if the user scrolls while the popover is open.
- Not sure how we've landed on focus locking inside an open select menu if there are nested tabbable elements. Some implementations lock, some close the menu on tab and continue to the next element after the trigger. In a combobox you can't really lock focus inside the popover because focus needs to stay in the input while the user types and interacts with the list.

gregwhitworth:
@chancestrickland thanks but that's not necessarily the purpose of this issue. Those seem like behavior considerations and most of the ones you noted are actually style based solutions which Open UI does not try to go after. This discussion is for the semantic meaning of the `<select>` versus a `combobox` both in written form and in practice. The reason for this was born from a discussion with @mfreed7, @una and other webdevs that a common desire from the `<select>` and why they re-build them is to have an input that allows filtering of the available options.

So my desired outcome from this issue is to determine:

1. Whether Open UI recommends that there should be two separate controls for `select` and `combobox`, why or why not.
2. How we plan to accomplish the search need within `select` if we decide that they should both exist but filtering should be expected offering on a `select`.

chancestrickland:
Makes sense to me for positioning, though is focus management not more of a behavioral thing we should be concerned with? And if it is, would that make a difference as to whether or not the two are separate controls or is there room for that kind of variance in a single spec?

gregwhitworth:
@chancestrickland Focus will be, I'm saying this specific issue isn't for investigating those. We have landed on some focus issues such as #74 and #82 and if you feel there others please feel free to file them in separate issues.

[...]

PR #157, opened 2020-08-17
Add typeahead algorithm for <select>

### LINKED ISSUE #145: Should <select> anatomy include type-to-search?
I've seen "typeahead' used to mean a few slightly different things, so in this issue I'm specifically talking about this behavior/functionality:

- When the `<select>` is focused (either open or closed), if the user types characters, these are collected in a buffer and selection jumps to the next `<option>` following the currently selected `<option>` that matches the text in the buffer (there may be some subtleties here that I am eliding for the purposes of this discussion).
- After some timeout with no more characters (short, maybe 0.5s) the buffer is cleared.
- The select has a button, not a textfield, so the buffer contents are not visible except as they are reflected in the changing selection.
- The `<option>`s are not filtered; typing makes the selection jump around but does not hide non-matching `<option>`s.

This behavior is present on the default `<select>` in desktop Chromium, Firefox, and Safari (there are some subtle differences but they are mostly the same).  I tried it on a few design systems ([Ant Design](https://ant.design/components/select/), [Carbon](https://www.carbondesignsystem.com/components/select/code/), [Atlas](https://atlaskit.atlassian.com/packages/core/select)) and the only one I saw that had this functionality was Carbon but that's because they build on the native `<select>`.

We should consider including this behavior in the `<select>` anatomy.  If we don't do so, implementers would probably include it in their controller code for [Custom Select](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ControlUICustomization/explainer.md) anyway, because not to do so would mean that custom selects would less accessible and less generally convenient than native `<select>`s.  This behavior is a big time-saver for typical `<select>` use cases like choosing one's country or state from a long list of `<option>`s. 

It's more complicated than the other `<select>` "controller code" we have defined so far, but I think it's worth including.  The fact that it's  not trivial means that it's less likely that developers rolling their own implementations would include this functionality, and I think that the lack of it in several major component frameworks is proof of this.

dandclark:
Add typeahead behavior to the `<select>` anatomy.
This behavior allows a user to choose items in the listbox by typing characters that match the first letters of a given `<option>`.  After a short timeout, the search "buffer" is cleared, and the user can begin a new search.

This is distinct from the "filtering" approach associated with comboboxes that have a textbox and listbox, which we haven't yet included in the anatomy.  See discussion in #104 and  #145.

Closes #145.

### DIFF
--- a/research/src/pages/select/select.proposal.mdx
+++ b/research/src/pages/select/select.proposal.mdx
@@ -137,17 +137,18 @@ children:
 
 ### part listbox
 
-| Event               | Behavior                                                                                                               | Impacts                                                                 |
-| ------------------- | ---------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------- |
-| `keydown(down key)` | Moves focus to the next `<option>` in the `listbox`                                                                    | focus                                                                   |
-| `keydown(up key)`   | Moves focus to the previous `<option>` in the `listbox`                                                                | focus                                                                   |
-| `keydown(enter)`    | Changes the `selected` state of the current `<option>` and updates the `<select>` value property                       | [selected](#selected) prop <br/> [value](#value) prop                   |
-| `keydown(space)`    | Changes the `selected` state of the current `<option>` and updates the `<select>` value property                       | [selected](#selected) prop <br/> [value](#value) prop                   |
-| `keydown(enter)`    | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` of the `<select>`         | [open](#open-state) state                                               |
-| `keydown(enter)`    | If the `<select>` does **not** have the `multiple` attribute then toggle the `aria-expanded` attribute of the `button` | [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
-| `keydown(space)`    | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` of the `<select>`         | [open](#open-state) state                                               |
-| `keydown(space)`    | If the `<select>` does **not** have the `multiple` attribute then toggle the `aria-expanded` attribute of the `button` | [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
-| `keydown(escape)`   | Toggles the `open` state of the `<select>`                                                                             | [open](#open-state) state                                               |
+|                                 Event                                  |                                                        Behavior                                                        |                                 Impacts                                 |
+| ---------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------- |
+| `keydown(down key)`                                                    | Moves focus to the next `<option>` in the `listbox`                                                                    | focus                                                                   |
+| `keydown(up key)`                                                      | Moves focus to the previous `<option>` in the `listbox`                                                                | focus                                                                   |
+| `keydown(enter)`                                                       | Changes the `selected` state of the current `<option>` and updates the `<select>` value property                       | [selected](#selected) prop <br/> [value](#value) prop                   |
+| `keydown(space)`                                                       | Changes the `selected` state of the current `<option>` and updates the `<select>` value property                       | [selected](#selected) prop <br/> [value](#value) prop                   |
+| `keydown(enter)`                                                       | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` of the `<select>`         | [open](#open-state) state                                               |
+| `keydown(enter)`                                                       | If the `<select>` does **not** have the `multiple` attribute then toggle the `aria-expanded` attribute of the `button` | [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
+| `keydown(space)`                                                       | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` of the `<select>`         | [open](#open-state) state                                               |
+| `keydown(space)`                                                       | If the `<select>` does **not** have the `multiple` attribute then toggle the `aria-expanded` attribute of the `button` | [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
+| `keydown(escape)`                                                      | Toggles the `open` state of the `<select>`                                                                             | [open](#open-state) state                                               |
+| `keydown(any single character that no other listbox event listens to)` | Run the [typeahead steps](#typeahead-steps) given `<select>` and `event.key`                                           | focus                                                                   |
 
 ### part option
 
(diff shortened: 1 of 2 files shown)


dandclark:
For the review, I'm interested in feedback not just on the algorithm but also on how I've presented it.
This is more complex than other behaviors described in the anatomy so far, so I didn't have a precedent
to work with.  I ended up writing it in a similar style to parts of the HTML spec.  Additionally I used JS
functions to describe the array manipulation in the typeahead steps.  I could alternatively do that part
in prose; not sure what is clearer.

research/src/pages/select/select.proposal.mdx
```diff
@@ -201,7 +202,47 @@ and is therefore considered to be invalid.
...
+     &lt;option&gt;, then presumably they wouldn't still be typing search text.
+   </p>
+1. Let `newly focused option` be the result of [finding the first matching
+   option](#find-the-first-matching-option) given `buffer` and `options for search`.
+1. If `newly focused option` is not null, move focus to `newly focused option`.
+
+#### To <a class="link" href="#find-the-first-matching-option" id="find-the-first-matching-option">find the first matching option</a> given a character buffer `buffer` and a list of options `options`, run these steps:
+
+1. For each `<option>` `option` in `options`:
+   1. If `buffer` is a case-insensitive prefix of `option.label`, then then return `option`.
```
gregwhitworth, line 234: Remove extra then

research/src/pages/select/select.proposal.mdx
```diff
@@ -201,7 +202,47 @@ and is therefore considered to be invalid.
...
+1. Return null.
+
+#### The <a class="link" href="#typeahead-buffer" id="typeahead-buffer">`typeahead buffer`</a> of a `<select>`
+
+A buffer containing the characters typed by the user for typeahead. Initially an empty string.
+
+#### <a class="link" href="#typeahead-search-timeout" id="typeahead-search-timeout">Typeahead search timeout</a>
+
+This is the maximum time that may elapse between keystrokes without the typeahead
```
gregwhitworth, line 243: Given that we want spaces to work within the typeahead should we remove that listener at the start of the algo and then add it back in during this section?

research/src/pages/select/select.proposal.mdx
```diff
@@ -201,7 +202,47 @@ and is therefore considered to be invalid.
...
+   `Array.prototype.slice.call(options, focusedIndex + 1, options.length).concat(Array.prototype.slice.call(options, 0, focusedIndex + 1))`.
+   Otherwise, let `options for search` be the array of options given by
+   `Array.prototype.slice.call(options, focusedIndex, options.length).concat(Array.prototype.slice.call(options, 0, focusedIndex))`.
+   <p class="note">
+     The point is that if we're in the middle of a search, we should keep trying to match on the
+     currently selected &lt;option&gt;. If this is a new search, we should start searching at the
+     &lt;option&gt; after the currently selected &lt;option&gt;. If the user wanted the current
+     &lt;option&gt;, then presumably they wouldn't still be typing search text.
+   </p>
+1. Let `newly focused option` be the result of [finding the first matching
```
gregwhitworth, line 227: Do we actually want focus to move? How do we allow for keyboard entry to continue if focus has left the input and moved to the option?

dandclark: I think it should work the same way as arrow up/down, which also move focus among the options.  When thinking about the [listbox events](https://open-ui.org/components/select#part-listbox), I assume that they bubble up to the listbox when an option is focused.  Otherwise arrow up/down, Enter/Escape, etc. wouldn't work when an option is focused.

research/src/pages/select/select.proposal.mdx
```diff
@@ -201,7 +202,47 @@ and is therefore considered to be invalid.
...
+
+1. Let `start new search` be false.
+1. If longer than the <a href="#typeahead-search-timeout">typeahead search timeout</a>
+   has elapsed since the previous invocation of the [typeahead steps](#typeahead-steps)
+   for `listbox`, then set `start new search` to true and set `buffer` to an empty string.
+1. Let `buffer` be `<select>`'s [`typeahead buffer`](#typeahead-buffer)
+1. Append `typed character` to `buffer`.
+1. Let `options` be the `<select>`'s list of options.
+1. Let `focusedIndex` be the index of `<select>`'s currently focused option.
+1. If `start new search` is true, then let `options for search` be the array of options given by
```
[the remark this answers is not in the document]
dandclark: This reorders the array for the search such that the items before the currently selected option (`Eu` in your example) come after the options following the currently selected one.  The focus can still jump backwards in the list to `Eu` (but if were other matches in the list after the currently selected option, it would jump to those first).

I think the way I tried to write this with JS functions is not at all clear though, so I've rewritten this using prose.  See changes here and in "find the first matching option".  Hopefully this is clearer now.

Of course I'm open to discussion on whether this is the final behavior we want to end up with.  At this point I'm just trying to match what I observe in Firefox/Chromium, which seems like reasonable behavior.  I've been using this to play with some of the interesting corner cases: https://jsfiddle.net/dandclark_msft/rx8tnps3/9/

research/src/pages/select/select.proposal.mdx
```diff
@@ -137,17 +137,18 @@ children:
...
+| `keydown(enter)`       | If the `<select>` does **not** have the `multiple` attribute then toggle the `aria-expanded` attribute on the `button part` | [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
+| `keydown(space)`       | If the `<select>` does **not** have the `multiple` attribute then toggle the state of `open` on the `<select>`              | [open](#open-state) state                                               |
+| `keydown(space)`       | If the `<select>` does **not** have the `multiple` attribute then toggle the `aria-expanded` attribute on the `button part` | [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) attr |
+| `keydown(escape)`      | Toggles the `open` state of the `<select>`                                                                                  | [open](#open-state) state                                               |
+| `keydown(a-z,A-Z,0-9)` | Run the [typeahead steps](#typeahead-steps) given `<select>` and `event.keyCode`                                            | focus                                                                   |
```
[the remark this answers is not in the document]
dandclark: Thanks - fixed in current version.

research/src/pages/select/select.proposal.mdx
dandclark: 'Space' will also trigger the 'typeahead steps' (after I addressed your feedback that the character for the event should not just be alphanumeric), so I don't think it should require extra handling with respect to the timeout.

research/src/pages/select/select.proposal.mdx
gregwhitworth: @dandclark You're correct, I guess I'm curious about how we pass back the entered content into the text input to ensure that the experience is better. I know that currently this isn't how they work but it is what seems to be a general agreement of a better experience when offering type ahead. I think for now we can resolve on this but this will probably need to be addressed if we agree to change the anatomy to a regular input.

research/src/pages/select/select.proposal.mdx
gregwhitworth: I'm confused how this will be the case given that the space event will still be tied to the listbox to open it unless I'm misunderstanding how the listeners are working and effectively the space event will be blocked while the method is running. Even if that is the case though, we should be clear in what can be leveraged in type ahead that even though space won't trigger the typeahead method it should be usable within the typeahead algo. Looking at the algo it doesn't seem to limit any chars but space is one that selects an option. I'm not sure how to reconcile those two

dandclark: Sorry, I misunderstood, I see what you're saying now.

I'd actually overlooked that we had `keydown(space)` handlers in `listbox`.  For Edge/Chrome and Firefox, 'space' doesn't do anything when the popup is opened.  Did we diverge from that behavior deliberately?  I'm tempted to say that we should converge with the current browser behavior and remove the `keydown(space)` handlers from the `listbox` (the ones on `button` would still be there, so that 'space' could still be used to open the popup). No functionality would be lost because it 'space' currently has the same behavior as 'Enter' for listbox.  Looking at the design systems listed at https://open-ui.org/components/select.research#names, (looking at button/listbox selects only, not the comboboxes with text input whose typeahead will work differently) I found these results:
Space doesn't do anything: [Ant Design](https://ant.design/components/select/), [Evergreen](https://evergreen.segment.com/components/select/), [Lightning](https://www.lightningdesignsystem.com/components/select/) (this one has typeahead), [Semantic UI](https://semantic-ui.com/modules/dropdown.html#selection), [Bootstrap](https://getbootstrap.com/docs/4.3/components/dropdowns/) (space toggles when on the button, but not when focus tabs into the listbox)
Space closes the popup, no typeahead: [Atlassian](https://atlassian.design/components/select/examples#popup-select), [Material](https://material-components.github.io/material-components-web-catalog/#/component/select), [Fluent UI](https://developer.microsoft.com/en-us/fluentui#/controls/web/dropdown)
Space closes the popup, has typeahead: [Carbon](https://www.carbondesignsystem.com/components/select/code/)
Carbon is funny because in their [live demo](https://www.carbondesignsystem.com/components/select/code/#live-demo) you can see that closing on space breaks their typahead.  Try selecting "Gray 100" via typeahead -- it's painful.

If we want to keep the behavior where 'space' closes the listbox and/or selects an option, I only see two possibilities:
1. Don't allow 'space' in typeahead.  This means that multi-word examples can't be easily selected via typahead, as in the Carbon example.
1. Treat 'space' as part of the typeahead if a search is underway, but make it select an option/close the popup otherwise.  This seems even more confusing and opaque to users than the current typeahead behavior in browsers.

Neither of these are great so I'm in favor of removing the `keyboard(space)` handlers from the listbox and relying on Enter for that functionality.

dandclark:
(Force-pushed to resolve merge conflicts).

research/src/pages/select/select.proposal.mdx
gregwhitworth: ah, interesting - good catch. I'm so used to space invoking controls that I must have assumed it would select and change state.

gregwhitworth: I filed an issue for this here: https://github.com/WICG/open-ui/issues/171

gregwhitworth (approved):

PR #165, opened 2020-08-24
feat(breadcrumb): add proposal

assuncaocharles:
Adds `Breadcrumb` proposal


assuncaocharles:
Hey @gregwhitworth 

Have a look in this one when you can :)

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+#### Structure <a class="link" href="#structure" id="structure"></a>
+
+- `<breadcrumb>` - The root element that contains the list of items and dividers **[required]**
+- `<breadcrumb-item>` - The element representing a place in the path to the current position on the application **[required]**
+- `<breadcrumb-divider>` - The divider between the `<breadcrumb-item />`
+
+#### Accessibility <a class="link" href="#accessibility" id="accessibility"></a>
+
+It should be wrapped by a navigation landmark `nav` conatining `aria-label` or `aria-labelledby`, the `<breadcrumb-item />` representing the current page should contain `aria-current="page"` ( if the current page isn't a link the attribut is optional )
```
pompomon, line 56:
```suggestion
It should be wrapped by a navigation landmark `nav` containing `aria-label` or `aria-labelledby`, the `<breadcrumb-item />` representing the current page should contain `aria-current="page"` ( if the current page isn't a link the attribute is optional )
```
typo ;)

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+#### Structure <a class="link" href="#structure" id="structure"></a>
+
+- `<breadcrumb>` - The root element that contains the list of items and dividers **[required]**
+- `<breadcrumb-item>` - The element representing a place in the path to the current position on the application **[required]**
+- `<breadcrumb-divider>` - The divider between the `<breadcrumb-item />`
+
+#### Accessibility <a class="link" href="#accessibility" id="accessibility"></a>
+
+It should be wrapped by a navigation landmark `nav` containing `aria-label` or `aria-labelledby`, the `<breadcrumb-item />` representing the current page should contain `aria-current="page"` ( if the current page isn't a link the attribute is optional )
```
gregwhitworth, line 56: Can we get very specific about the parts. eg:

part=breadcrumb:
* role = nav

I'm also curious on the name computation here. Is it common for the breadcrumb to be what you focus on or solely the first link? If you land on the container and it is focusable, how does it compute the name?

gregwhitworth (changes requested):

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,62 @@
...
+- [Carbon](https://www.carbondesignsystem.com/components/breadcrumb/code/)
+- [Fabric](https://developer.microsoft.com/en-us/fluentui#/controls/web/breadcrumb)
+- [Lightning](https://www.lightningdesignsystem.com/components/breadcrumbs/)
+- [Semantic UI](https://semantic-ui.com/collections/breadcrumb.html)
+
+## &lt;breadcrumb&gt; Properties <a href="#breadcrumb-properties" id="breadcrumb-properties"></a>
+
+| Attribute Name | Type             | Default Value | Description                                                                      |     |
+| -------------- | ---------------- | ------------- | -------------------------------------------------------------------------------- | --- |
+| `items`        | `HTMLCollection` |               | A `HTMLCollection` of items to represent the current path within the application |     |
```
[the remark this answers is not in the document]
assuncaocharles: Gotcha your point, my idea to add it here would be to allow `child` and `shorthand` apis, but I can agree with you and remove it from attributes.

research/src/pages/breadcrumb.proposal.mdx
assuncaocharles: The commons is for the container not be focusable and you always land in the first link itself

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,62 @@
...
+| ------------- | -------- | ------------- | ---------------------------------------------------- |
+| `content`     | `custom` | `/`           | The content to be rendered as boundary for the items |
+
+### Anatomy <a class="link" href="#anatomy" id="anatomy"></a>
+
+#### Diagram <a class="link" href="#diagram" id="diagram"></a>
+
+<BreadcrumbAnatomy />
+
+#### Structure <a class="link" href="#structure" id="structure"></a>
```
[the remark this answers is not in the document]
assuncaocharles: Updated, have a second look please

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+- `<breadcrumb>` - The root element that contains the list of items and dividers, it's composed by `nav` and `ol` elements  **[required]**
+- `<breadcrumb-item>` - The element representing a place in the path to the current position on the application, it's a focusable `li` that can contain text or anchors **[required]**
+- `<breadcrumb-divider>` - The divider between the `<breadcrumb-item />`, it should be a non-focusable li
+
+Proposed HTML:
```
gregwhitworth, line 41: I don't think this is feasible, as effectively we're saying that a breadcrumb item is the `<li>` correct and the breadcrumb is the `nav`? I think it would be great if you and I can jump on a call to discuss this. DM me on twitter, I'm happy to setup a relatively early meeting to make this happen.

gregwhitworth:
@assuncaocharles I left a comment, ultimately I think it'd be cool to jump on a call for 30 minutes to work through some of these and get on the same page. To help land this. DM me on Twitter or email me so we can set this up.

research/src/pages/breadcrumb.proposal.mdx
chrisdholt: I'd also expect that the anchor is the focusable portion here rather than the list item - I don't see why we'd have a non-focusable element receive focus as opposed to the anchor within listed below. Similarly, I'm not sure we'd want the divider/separator enumerated as a list item itself.

@gregwhitworth we're implementing Breadcrumb currently in FAST, it would be great to get visibility on this PR and some of the potential issues for the broader Open UI community.

assuncaocharles: Hey @gregwhitworth 

This is the pattern proposed by aria 

```html
<nav aria-label="Breadcrumb" class="breadcrumb">
  <ol>
    <li>
      <a href="../../">
        WAI-ARIA Authoring Practices 1.1
      </a>
    </li>
....
```

chrisdholt: I'd push back on this a bit - this is an example by ARIA, but it's not "the way" IMO.

assuncaocharles: `The following example demonstrates the breadcrumb design pattern.` Looks pretty heavy on the definition way for me. Also it's pretty much what was implemented by the libraries that has minimum a11y support for breadcrumbs.

chrisdholt: I don't want to get too far down the semantic rabbit hole, but I think that semantics are important - especially when it comes to the spec. The *spec* is the definition. The example, as you've noted above - _demostrates_ the design pattern. The spec itself points to that as an example. It is *one implementation* and I think we should be careful to conflate the example with the specification.

The spec on this is fairly simply as of now and so I'll paste below:
> - Breadcrumb trail is contained within a navigation landmark region.
> - The landmark region is labelled via aria-label or aria-labelledby.
> - The link to the current page has aria-current set to page. If the element representing the current page is not a link, aria-current is optional.

From an accessibility standpoint, the above is the minimum bar and the example is, as you mentioned, a very common implementation of the pattern following the spec.

Given the above, the important bit here is the role and hierarchy. While I agree we should aim to use native elements wherever possible, I believe that is too prescriptive an approach for us to take.

chrisdholt: I don't think there's anything particularly wrong with the DOM structure, but I would add that using the roles model is equally as valid here. There may be external influences, pre-existing implementations for teams, etc where using that exact structure may be prohibitive. I'm all good with focusing on the roles (nav, list, listitem) as the recommendation with an example of that using what you proposed above. That's really all I'm advocating for. The overall ARIA hierarchy vs. the implementation of that. Hopefully that clarifies what I mean.

assuncaocharles: That do makes sense for me, but I think that what you see in the proposal or 

```html
<div role="navigation">
    <div role="list>
        <div role="listitem"></div>
    </div>
</div>
```

are equivalent and it's correct replacement. It's defined as "Proposed HTML" such as aria but not "The only correct way". Using the elements as in the example is the most generic way to represent the intention so it can be achieve as the same for other developers implementing it. In Fluent for example we are doing it only using divs and roles... But as a generic way I guess having the correct elements gives better the message.

chrisdholt: Yep, same page. :thumbsup:

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,60 @@
...
+### Anatomy <a class="link" href="#anatomy" id="anatomy"></a>
+
+#### Diagram <a class="link" href="#diagram" id="diagram"></a>
+
+<BreadcrumbAnatomy />
+
+#### Structure <a class="link" href="#structure" id="structure"></a>
+
+- `<breadcrumb>` - The root element that contains the list of items and dividers, it's composed by `nav` and `ol` elements  **[required]**
+- `<breadcrumb-item>` - The element representing a place in the path to the current position on the application, it's a `li` that can contain text or anchors **[required]**
```
gregwhitworth, line 38: why are you limiting the `<li>` to only containing text or anchors? I don't think we should prescribe the content in this scenario. I think you should delete everything after position in the application

gregwhitworth: also - should be `in the application` not on

research/src/pages/breadcrumb.proposal.mdx
gregwhitworth: > @gregwhitworth we're implementing Breadcrumb currently in FAST, it would be great to get visibility on this PR and some of the potential issues for the broader Open UI community.

Feel free to ping on discord and raise the issues as separate issues (that you believe you need resolution on) to discuss on the call by adding them to the agenda by `agenda+` label

research/src/pages/breadcrumb.proposal.mdx
gregwhitworth: > It should be wrapped

Can you define what `it` is

Also, can you add the part attribute to your items so that your HTML maps to your anatomy?

gregwhitworth (changes requested):

research/src/pages/breadcrumb.proposal.mdx
assuncaocharles: Done

research/src/components/breadcrumb-anatomy.js
```diff
@@ -0,0 +1,27 @@
...
+import './anatomy.css'
+import './breadcrumb-anatomy.css'
+
+const BreadcrumbAnatomy = () => {
+  return (
+    <div class="component-anatomy-wrapper">
+      <input type="checkbox" id="show-slots" />
+      <label for="show-slots"> Show slots</label>
+      <div class="component-anatomy">
+        <host name="breadcrumb">
```
gregwhitworth, line 13: As we discussed on the call - here is the anatomy I'm proposing so that we can limit the explanations later since you're not actually proposing new components nor elements.

```
<host name="bread-crumb">
     <slot name="breadcrumb-container">
        <ol>
            <li part="breadcrumb-item"></li>
            <li part="breadcrumb-divider" aria-hidden="true"></li>
        </ol>
     </slot>
</host>
```

cc: @chrisdholt

assuncaocharles: That's a little confusing, has the anatomy for should just represent where we would place the "known" components, like the `Table` anatomy that shows where we place `table-row`, `table-header`, `table-pagination` etc... If I naively apply the proposed I get this rendered:

Screen Shot 2020-09-23 at 05 01 28

Trying to find a common ground I am applying this:

```jsx
        <host name="breadcrumb" role="nave">
          <slot name="breadcrumb-container">
            <part name="ol">
              <part name="li">item</part>
              <part name="li" aria-hidden="true">
                divider
              </part>
            </part>
          </slot>
        </host>
```

Screen Shot 2020-09-23 at 05 02 28

assuncaocharles: Also edited the structure section descriptions

gregwhitworth: @assuncaocharles that's precisley what I meant by the HTML I wrote. Only modification is to have this:


`<part name="li [part=item]">` and the same for divider. You can leave the content of item and divider there

research/src/components/breadcrumb-anatomy.js
```diff
@@ -0,0 +1,28 @@
...
+const BreadcrumbAnatomy = () => {
+  return (
+    <div class="component-anatomy-wrapper">
+      <input type="checkbox" id="show-slots" />
+      <label for="show-slots"> Show slots</label>
+      <div class="component-anatomy">
+        <host name="breadcrumb" role="nav">
+          <slot name="breadcrumb-container">
+            <part name="ol">
+              <part name="li">item</part>
```
gregwhitworth, line 16: as noted in the other comment add into the name `[part=breadcrumb-item]` to both the `ol` and `li`

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,61 @@
...
+    <li><a>Item 1</a></li>
+    <li aria-hidden="true">/</li>
+    <li><a>Item 2</a></li>
+  </ol>
+</nav>
+```
+
+#### Accessibility <a class="link" href="#accessibility" id="accessibility"></a>
+
+The list of Breadcrumb Items should be wrapped by a navigation landmark `nav` containing `aria-label` or `aria-labelledby`, the `<breadcrumb-item />` representing the current page should contain `aria-current="page"` ( if the current page isn't a link the attribute is optional )
```
gregwhitworth, line 55: Can we now get very explicit. For example:

* `breadcrumb`: has a semantic representation of a `<nav>` element

etc

gregwhitworth (changes requested):
Two last little changes then I'll approve. Once you get an updated anatomy can you please open an issue specific to that anatomy as I'd like to ensure that the group is in agreement before landing the PR. Also, if @chrisdholt or someone else can give this a review new proposals need at least 2 people to sign off. And as noted, I'd like resolution on the proposed anatomy

research/src/components/breadcrumb-anatomy.js
chrisdholt: I've got some concerns on the above and would like to further discuss before this gets approved. I think I'm more closely aligned to an *end result* of what Charles initially had. I brought this to some folks on our team and the following things came up:

- We aren't sure that the separator should be a list item
- I'm not sure I would expect the list item (li) itself to be the actionable bit. 
    - What affordance does a non-sighted user have to know that's actionable?
    - Slotted content for an item makes more sense to me here, but there isn't an "item" concept so that doesn't seem possible. That would allow me to reuse the anchor and get it's native behaviors and semantics, which is valid. Consider that I want to open these links in a new tab, or a new window because it's a form and I don't want to navigate you away from the page.
    - It seems like we are convoluting multiple semantic roles with the current implementation. `role="listitem"` + "role="link".

As it stands now, I'd be a voice of dissent in moving towards this implementation.

assuncaocharles: Hey @chrisdholt 

    1. I think divider should be list item, it's the common across most of the other libs and even aria recommendation. Of course the implementation can change and like in Fabric it can be rendered inside of `Item` for several reasons. 
    2. I don't think the item ( `li` ) should be actionable at all also, it can contain anchor or texts inside.

assuncaocharles: @gregwhitworth did your last update, this is what we have:

Screen Shot 2020-09-24 at 07 53 40

assuncaocharles:
@gregwhitworth solved them all

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,66 @@
...
+</nav>
+```
+
+#### Accessibility <a class="link" href="#accessibility" id="accessibility"></a>
+
+The Breadcrumb is composed by:
+
+- A `nav` equivalent wrapper that should always have `aria-label` or `aria-labelledby`
+- A ordered `list`
+- `li` items and dividers where the `li` represeting the item can contain anchor or other elements and the `li` represeting the divider should have `aria-hidden="true"`
```
chrisdholt, line 59: A few notes here based on feedback in another thread:

1) Should we use the name divider or separator and why?
2) Does the li actually represent the item if it is not actionable or is the item slotted content?
3) If the expectation is that a separator (or divider) slotted in as part of a list item is presentational only - therefore, *always hidden*, why make that the rule? It seems more intuitive to expect that we have a presentational recommendation with this as "allowable". Either way, I think we need a caveat that this can be handled a couple of ways (per Charles' comment elsewhere).

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,66 @@
...
+- If the last `li` represeting item has an anchor for the current page, this anchor should have `aria-current="page"` otherwise this attribute is optional
```
chrisdholt, line 60: We should note that only anchors which include an href should reference `aria-current="page"`. Anchors without an href are treated as text and should not include `aria-current="page"` as certain AT (such as narrator) will announce the anchor (without an href) as a link anyway.

chrisdholt: From the [spec](https://www.w3.org/TR/wai-aria-1.2/#aria-current): 
> The aria-current attribute is used when an element within a set of related elements is visually styled to indicate it is the current item in the set...A page token used to indicate a link within a set of pagination links, where the link is visually styled to represent the currently-displayed page.

Anchor without href as text via UIA: https://w3c.github.io/html-aam/#el-a-no-href

research/src/pages/breadcrumb.proposal.mdx
assuncaocharles: 1. Doesn't look like there's any consensus between libs, but looks like a `space X tabs` discussion. I have no problem with either names. 
2. I believe it's the slot for the item as shown in the anatomy
3. The usage of `aria-hidden` was the common practice I found in the libs that are not using the divider as an pseudo element.

research/src/pages/breadcrumb.proposal.mdx
assuncaocharles: I would guess that using anchor without `href` is a misusage, not sure if it should be mentioned here.

chrisdholt: It’s totally valid and treated as text by the browser per the note above. It would be invalid and cause issues when used with aria-current=“page” as that could incorrectly be announced as a link.

assuncaocharles: Isn't this understood by "anchor for the current page"? How could be anchored to the current page without the `href`?

assuncaocharles:
Any update here @gregwhitworth ?

gregwhitworth:
@assuncaocharles ping to @chrisdholt can you please keep the convo on your discussion going as I'd like those resolved before moving forward on this.

research/src/components/breadcrumb-anatomy.js
gregwhitworth: @chrisdholt can you review please

assuncaocharles:
> resolved

@gregwhitworth Not sure what action you expect from me here or was it just a ping for @chrisdholt ?

chrisdholt:
> > resolved
> 
> @gregwhitworth Not sure what action you expect from me here or was it just a ping for @chrisdholt ?

Just a ping for me. I'm reviewing and will open issues for anything I think deserves a discussion.

research/src/pages/breadcrumb.proposal.mdx
chrisdholt: Per my note above, an anchor with an href is a link, An anchor without an href is considered text. 

I think the current language in the W3C specification describes this way more succinctly than I could and adopting it in place of the above would resolve this issue.

> The link to the current page has aria-current set to page. If the element representing the current page is not a link, aria-current is optional.

research/src/pages/breadcrumb.proposal.mdx
chrisdholt: 1. Issue #200 created to discuss this
2. Leaving for now
3. Issue #201 created to discuss

chrisdholt (approved):
Approved, though the one thing I would really like to see updated here would be to leverage the W3C's note regarding the use of `aria-current="page"`. It conveys the nuance I think we're looking to convey in a very succinct manner.

Link to that suggestion is here: https://github.com/WICG/open-ui/pull/165/files#r507201612

research/src/pages/breadcrumb.proposal.mdx
assuncaocharles: Applied

assuncaocharles:
> Approved, though the one thing I would really like to see updated here would be to leverage the W3C's note regarding the use of `aria-current="page"`. It conveys the nuance I think we're looking to convey in a very succinct manner.
> 
> Link to that suggestion is here: https://github.com/WICG/open-ui/pull/165/files#r507201612

Applied suggested change

assuncaocharles:
So can we merge this one @gregwhitworth ?

assuncaocharles:
Any update here @gregwhitworth ?

gregwhitworth:
@assuncaocharles there was a resolution regarding divider. See this issue: https://github.com/WICG/open-ui/issues/200

assuncaocharles:
@gregwhitworth I removed the `Divider` but I am not sure how to represent it as a `::mark`  in the anatomy

una:
@assuncaocharles I think you just need to represent it as `item::marker`. Does it need a specified part if its a pseudo element?

chrisdholt:
> @assuncaocharles I think you just need to represent it as `item::marker`. Does it need a specified part if its a pseudo element?

I don't think this would necessitate a part if it's a pseudo element.

assuncaocharles:
> @assuncaocharles I think you just need to represent it as `item::marker`. Does it need a specified part if its a pseudo element?

Not sure how to represent like that

gregwhitworth:
> Not sure how to represent like that in the anatomy. How it would be in the anatomy @una @gregwhitworth @chrisdholt ?

Technically the `::marker` psuedo element is there for any `li` so we don't have to represent it. Although I get your reasoning for including it, what you have represented is fine with me although I'd prefer for it to not be side by side, or to make the `psuedo element` size to min-content. This makes the psuedo element appear much larger than that of `item`.

Also, the box itself of the marker comes before the item itself so it should be placed first.

Screen Shot 2020-11-25 at 1 54 11 PM

gregwhitworth (changes requested):
> Not sure how to represent like that in the anatomy. How it would be in the anatomy @una @gregwhitworth @chrisdholt ?

Technically the `::marker` psuedo element is there for any `li` so we don't have to represent it. Although I get your reasoning for including it, what you have represented is fine with me although I'd prefer for it to not be side by side, or to make the `psuedo element` size to min-content. This makes the psuedo element appear much larger than that of `item`.

Also, the box itself of the marker comes before the item itself so it should be placed first.

Screen Shot 2020-11-25 at 1 54 11 PM

assuncaocharles:
> > Not sure how to represent like that in the anatomy. How it would be in the anatomy @una @gregwhitworth @chrisdholt ?
> 
> Technically the `::marker` psuedo element is there for any `li` so we don't have to represent it. Although I get your reasoning for including it, what you have represented is fine with me although I'd prefer for it to not be side by side, or to make the `psuedo element` size to min-content. This makes the psuedo element appear much larger than that of `item`.
> 
> Also, the box itself of the marker comes before the item itself so it should be placed first.
> 
> Screen Shot 2020-11-25 at 1 54 11 PM

@gregwhitworth In the image bellow `::marker` is inside `li` as a pseudo-element would be. 

98810991-dceb3700-23fe-11eb-8fa6-44f10cfbea66

kant2002:
I notice that SAP Fiory provide some interesting concept for Breadcrumb.
image
I think this concept valuable for large projects, where breadcrumbs can be rather large in length

My understanding that this cannot be represented in current breadcrumb proposal. So question, should such advanced cases like this be part of Open UI?

kant2002:
The Carbon Design System from IBM suggest different way for collapse of breadcrumbs

image

from [there](https://www.carbondesignsystem.com/patterns/overflow-content)

una:
It seems like there are open issues around a breadcrumb dropdown functionality. That seems like a (very useful) combination of components, rather than default base functionality. Would you all be okay with merging this PR and taking that discussion into a new issue?

una:
@assuncaocharles did you add `::marker`? I don't see it in the PR. Would love to merge it in

assuncaocharles:
@una i did not, I'm not sure how to express that in the anatomy as I said before

gregwhitworth:
@assuncaocharles this one looks correct to me, granted it wouldn't have the divider part on it: https://github.com/WICG/open-ui/pull/165#issuecomment-734293512

Finally, and this is what I'd like @chrisdholt and @una to weigh in on - do we really need to add `::marker` as it is already a standardized pseudo element of a list-item?

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,66 @@
...
+A `<breadcrumb />` trail consists of a list of links to the parent pages of the current page in hierarchical order. It helps users find their place within a website or web application. Breadcrumbs are often placed horizontally before a page's main content.
+
+## Prior Art/Examples <a href="#prior-art" id="prior-art"></a>
+
+- [Ant Design](https://ant.design/components/breadcrumb/)
+- [Bootstrap](https://getbootstrap.com/docs/4.3/components/breadcrumb/)
+- [Auth0](https://styleguide.auth0.com/components/breadcrumb)
+- [Atlas Kit](https://atlassian.design/components/breadcrumbs/examples)
+- [Carbon](https://www.carbondesignsystem.com/components/breadcrumb/code/)
+- [Fabric](https://developer.microsoft.com/en-us/fluentui#/controls/web/breadcrumb)
```
chrisdholt, line 25:
```suggestion
- [Fabric](https://developer.microsoft.com/en-us/fluentui#/controls/web/breadcrumb)
- [FAST](https://explore.fast.design/components/fast-breadcrumb)
```

chrisdholt:
> @assuncaocharles this one looks correct to me, granted it wouldn't have the divider part on it: [#165 (comment)](https://github.com/WICG/open-ui/pull/165#issuecomment-734293512)
> 
> Finally, and this is what I'd like @chrisdholt and @una to weigh in on - do we really need to add `::marker` as it is already a standardized pseudo element of a list-item?

You can create the same UIA representation using ARIA and while I know that's not completely ideal, it is valid and `::marker` wouldn't work there. I don't know that we need to make that the rule though, more saying that both can be accurate representations - this really comes down to nuance in my mind.

gregwhitworth:
@assuncaocharles can you do a final pass looking at @chrisdholt commit recommendations so that we can do an initial landing of this PR

research/src/pages/breadcrumb.proposal.mdx
andrico1234: I updated this PR to include your suggestion! (a little late to the party 😄)

research/src/pages/breadcrumb.proposal.mdx
andrico1234: Updated this PR to reference the open discussions

andrico1234:
Hi folks (@gregwhitworth @una @chrisdholt @assuncaocharles )

I've updated this PR to include Chris' suggested changes. Are you happy for me to merge this in?

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+import BreadcrumbAnatomy from '../components/breadcrumb-anatomy'
+
+## Overview
+
+Breadcrumbs are navigation land marks used to show a user's location within the site or app.
+
+## Use Cases
+
+A `<breadcrumb />` trail consists of a list of links to the parent pages of the current page in hierarchical order. It helps users find their place within a website or web application. Breadcrumbs are often placed horizontally before a page's main content.
```
scottaohara, line 16:
```suggestion
A `<breadcrumb />` trail consists of a list of links to the parent pages of the current page in hierarchical order. It helps users find their place within a website or web application. Breadcrumbs are often placed horizontally before or as the first component of a page's main content.
```

scottaohara: a breadcrumb is not always a hierarchical ordering of links.  sometimes it represents the history that a user took to reach their current page.  does that matter for this?

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+### Anatomy <a class="link" href="#anatomy" id="anatomy"></a>
+
+#### Diagram <a class="link" href="#diagram" id="diagram"></a>
+
+<BreadcrumbAnatomy />
+
+#### Structure <a class="link" href="#structure" id="structure"></a>
+
+- `<breadcrumb>` - The root element that that has the semantic value of navigation **[required]**
```
scottaohara, line 38:
```suggestion
- `<breadcrumb>` - The root element that has the semantics of a navigation **[required]**
```

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+- `<ordered-list>` - The list containing items and dividers **[required]**
+- `<list-item>` - The wrapper to items or divider content **[required]**
+
+Proposed HTML:
+
+```html
+<nav>
```
scottaohara, line 45: this is just a note, but if this were to be adopted it'd be best if this were given a control type/xml value of 'breadcrumb navigation', or something similar.  exposing as just a `nav` is possible now, and not quantifying this doesn't provide much value from a semantics/accessibility semantics point of view

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+  <ol>
+    <li><a>Item 1</a></li>
+    <li aria-hidden="true">/</li>
```
scottaohara, line 48: would be better represented by a `:marker` but... without that being created, this is "fine"

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+  </ol>
+</nav>
+```
+
+#### Accessibility <a class="link" href="#accessibility" id="accessibility"></a>
+
+The Breadcrumb is composed by:
+
+- A `nav` equivalent wrapper that should always have `aria-label` or `aria-labelledby`
+- A ordered `list`
```
scottaohara, line 59:
```suggestion
- An ordered `list`
```

scottaohara: note that without numeric / ordering list markers, there is presently no meaningful difference between an `ol` or `ul`.  they are both exposed as a `list`.

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+    <li><a>Item 2</a></li>
+  </ol>
+</nav>
+```
+
+#### Accessibility <a class="link" href="#accessibility" id="accessibility"></a>
+
+The Breadcrumb is composed by:
+
+- A `nav` equivalent wrapper that should always have `aria-label` or `aria-labelledby`
```
scottaohara, line 58: this would only be necessary if, when implemented, a `nav` was not given more context to its purpose.

I'd submit that there's no reason to add this, and as a web component an accessible name of `breadcrumb` or an `aria-roledescription` of `breadcrumb` could instead be set by default to not require an author need to specify these

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+- A ordered `list`
+- `li` items and dividers where the `li` represeting the item can contain anchor or other elements and the `li` represeting the divider should have `aria-hidden="true"`
+- The link to the current page has aria-current set to page. If the element representing the current page is not a link, aria-current is optional.
```
scottaohara, line 61:
```suggestion
- The link or non-interactive element to the current page has `aria-current=page`. 
```

scottaohara: setting `aria-current` is not optional.

what should be optional is setting the value to `page` or `location`.

research/src/pages/breadcrumb.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+## Resources <a class="link" href="#resources" id="resources"></a>
+
+- [Breadcrumb design pattern example](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html)
```
scottaohara, line 65: Léonie Watson literally wrote the spec on `aria-current`, so it is worth noting her explainer: https://tink.uk/using-the-aria-current-attribute/

note that here, Léonie specifically calls out that a breadcrumb should use `aria-current` and not `page`.

I've used both in my explainer on this component: https://scottaohara.github.io/a11y_breadcrumbs/

scottaohara (changes requested):
things to update / considerations to make

research/src/pages/breadcrumb.proposal.mdx
andrico1234: Great! I've added both to the resources list

andrico1234:
Thanks for taking the time to leave comments @scottaohara , I've merged in the typo corrections and other suggested changes. 

As for the rest of the comments, I'm not sure how to best approach them. I'll be happy to make notes within the proposal about these unresolved questions, e.g. the correct markup for the spacer, to get this (20 month old 😱) PR merged in. Whoever decides to champion this proposal in the future can resolve the comments.

andrico1234 (approved):

Issue #116, opened 2020-07-02
Add research for <dialog> and related popups


melanierichards:
Currently conducting some research on various flavors of popups, in order to determine how we might make these easier to author with the right accessibility semantics, user input behaviors, styling, etc. Findings from this research should be added to the Open UI project. :)

melanierichards:
@gregwhitworth I can't self-assign, mind assigning me to this issue?

melanierichards:
PR #205 was opened to address this, and was intentionally left very broad. I'd like to split up this PR into "things we think are a good match for `<popup>`" and "other top layer things". Per [MS Edge Explainers #455](https://github.com/MicrosoftEdge/MSEdgeExplainers/issues/455) we should audit various ARIA roles and make sure we've got all the popup-y things. That was part of my initial research, but it's been a bit so I'll go back with a fine-toothed comb and make sure I didn't miss something.

andrico1234:
I'll close this ticket as the popup API spec has recently been merged into the [Open UI site](https://open-ui.org/components/popup.research.explainer) and [this issue](https://github.com/openui/open-ui/issues/18) focuses specifically on a `<dialog>` spec.

@melanierichards  Feel free to reopen if I've gotten too trigger happy with the close button 😄

PR #205, opened 2020-10-21
Add Popup research

melanierichards:
This research PR gathers together various types of popups and related concepts into one collection. It is likely sensible to break these popup types down further into separate Open UI components. Some of these may also not be subclasses of "popup"s, but are their own type of component entirely (tooltip may be such an example).

There is a LOT of detail to dig into here, and we can build out the research much more on each popup subclass. The conversation I'm primarily interested in driving is, "what is a popup, anyway?" It would be helpful to break these concepts down into:

* Popups for which you might reach for a new `<popup>` element, knowing that you might have to add a bit of semantic sugar with ARIA/extra behaviors.
* Popups for which you'd expect the web platform to provide a more specific, unique control.
* Things that are not popups.

We should come to a conclusion on what separates these buckets of components. What truly makes a popup a **popup**, vs some other element that just happens to be rendered in the top layer?

More details are in the PR changes, but the concepts here include:

* Autocomplete (Popup Part)
* Alert
* Combobox (Popup Part)
* Date-Time Picker (Popup Part)
* Dialog
* Menu
* Misc (teaching UI, other arbitrary popups)
* Select (Popup Part)
* Toast
* Tooltip


melanierichards:
Addresses #116

patrickhlauke:
i've had so many discussions with folks around the idea that "if it pops up and contains lots of structured content, and particularly focusable/interactive child elements, and the user is not expected to operate anything inside the popup AND the underlying page at the same time, then it's conceptually more like a dialog that just happens to be styled like a popup".

related https://github.com/twbs/bootstrap/issues/28446

melanierichards:
Closing this stale PR in favor of opening a more scoped interpretation of `popup`.

Issue #184, opened 2020-09-30
Naming convention for proposed elements


una:
RE: #180 

I'd like to discuss about naming conventions for these elements. Take file as an example:

- current HTML: `<input type="file">`
- open-UI proposal: `<file>`

Another option would be `<file-input>` or `<input-file>` as a naming convention. These feel more clear to me and nod to the underlying web components (which we hope to become native). You can also replace "file" with anything and you still have a naming pattern, i.e.:

- `<input-color>`
- `<input-date>`
- `<input-dropdown>`

etc.

kenchris:
People might have made custom elements with those names as they contain a hyphen, so that might cause issues

gregwhitworth:
@kenchris that's a valid point but not entirely sure how that relates to Open UI specifically. If they make their way into the platform itself they may not have a hyphen as only platform based components can be implemented without a dash. I think as it relates to creating a component following Open UI leveraging the web component paradigm makes sense until a standardize HTML element is created.

With regards to the options, I kind of want to get away from the input foo paradigm personally. That's why I initially recommended file as I think that's what should be proposed for the native element. That said, Open UI should probably not take a dependency on that until it is standardized in WHATWG.

**TLDR;** I agree that we should use a web components naming convention unless a standardized element that matches the same semantic meaning can be used in its place. Which specific name for file and inputs in general I don't have a strong pref.

bkardell:
>  I kind of want to get away from the input foo paradigm personally. 

I would say this is a largely uncontroversial position among implementers/plenty of standards people.  The type model in which a thing totally changes its 'ness' is a regrettable choice that is full of problems we are still digging out of, and we'd rather use composition/decoration here. So, I agree.

levithomason:
The ideal eventual HTML element would be `<file>`, etc, where the name of the component is the name of the element.  This is how all component libraries I've seen name and implement their components.  If we need to prefix for now, I'd support `<oui-*>` as a prefix meaning "Open UI", however, I don't think this should impact what browsers might implement someday.

mrmckeb:
If we're not talking about standardisation, and instead about using these specifications to create inputs _today_, we could also prefix with `wicg-` so that it's clear that the input is built to the spec we've proposed. I don't mind `input-`, but I'm not a fan of suffixes (I _feel_ that most developers would prefer prefixes).

[...]

PR #207, opened 2020-10-25
Updated checkbox anatomy and parts based on resolution

gregwhitworth:
This is an update to the checkbox anatomy based on the resolution of https://github.com/WICG/open-ui/issues/199

Additionally, I am adding in the `oui-` namespace for the host to show it's a custom web component and also removed the `part=control` as I don't think there is a need for it especially since it isn't referenced in the rest of the spec.

### DIFF
--- a/research/src/components/checkbox-anatomy.js
+++ b/research/src/components/checkbox-anatomy.js
@@ -4,7 +4,7 @@ import './anatomy.css'
 export default () => {
   return (
     <div className="component-anatomy">
-      <host name="checkbox">
+      <host name="oui-checkbox">
         <part name="label">
           <slot name="label"></slot>
         </part>
--- a/research/src/components/checkbox-anatomy.js
+++ b/research/src/components/checkbox-anatomy.js
@@ -8,13 +8,8 @@ export default () => {
         <part name="label">
           <slot name="label"></slot>
         </part>
-        <part name="checkbox">
-          <part name="checked-indicator">
-            <slot></slot>
-          </part>
-          <part name="indeterminate-indicator">
-            <slot></slot>
-          </part>
+        <part name="indicator">
+          <slot></slot>
         </part>
       </host>
     </div>
--- a/research/src/pages/checkbox.proposal.mdx
+++ b/research/src/pages/checkbox.proposal.mdx
@@ -66,11 +66,10 @@ The `<checkbox>` control is primarily leveraged to indicate a binary state. For 
 
 #### Slots
 
-| Slot Name                 | Description                                                   | Fallback Content                                 |
-| ------------------------- | ------------------------------------------------------------- | ------------------------------------------------ |
-| `label`                   | Add custom markup for the control's label                     | Empty                                            |
-| `checked-indicator`       | Content to indicate the checkbox is in a checked state        | Element with checked and indeterminate indcators |
-| `indeterminate-indicator` | Content to indicate the checkbox is in an indeterminate state | Element with checked and indeterminate indcators |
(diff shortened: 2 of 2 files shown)


research/src/components/checkbox-anatomy.js
```diff
@@ -8,13 +8,8 @@ export default () => {
         <part name="label">
```
kant2002, line 8: Should the line above (`<host name="checkbox">`) have name changed from `checkbox` to `oui-checkbox`?

research/src/pages/checkbox.proposal.mdx
```diff
@@ -98,11 +93,11 @@ The `<checkbox>` control is primarily leveraged to indicate a binary state. For
...
 #### Keyboard
 
-Keyboard users can operate the `<checkbox>` as follows:
+Keyboard users can operate the `<oui-checkbox>` as follows:
 
-- `<checkbox>` is eligible for document focus.
-- `<checkbox>` is eligible for `autofocus` when the `autofocus` attribute is present on the host.
-- `<checkbox>` can be the target of `click` events (when not `disabled`) either through mouse click, user tap, or keyboard `space` key press.
+- `<oui-checkbox>` is eligible for document focus.
+- `<oui-checkbox>` is eligible for `autofocus` when the `autofocus` attribute is present on the host.
```
kant2002, line 99: Small question about terminology. Since I do not actively involved and read everything I can miss something.
I in general understand sentence and it's spirit. But what confusing me is before I have mapping between word `host` and tag `<host>` in the example. And I clearly see that <host> tag indicate just structure of the component, but not the how I would write component. Right now `<oui-checkbox>` too close to actual way how to write component in the HTML, and I do not immidiately associate that HTML with actual anatomy of the component.

gregwhitworth: > And I clearly see that tag indicate just structure of the component, but not the > how I would write component.

Can you expound, I'm not sure how to address your feedback to provide clarity

research/src/components/checkbox-anatomy.js
gregwhitworth: yeah it should, thanks

nicholasrice (approved):

research/src/pages/checkbox.proposal.mdx
```diff
@@ -54,34 +54,29 @@ The `<checkbox>` control is primarily leveraged to indicate a binary state. For
 #### DOM Structure
 
 ```html
-<host>
```
kant2002, line 57: @gregwhitworth in continue to previous comment. I think better to leave `host` since this is was implicitly hint on Shadow DOM/HTML template concepts. And my understanding that skeletons/anatomy is something similar in the spirit. 

Maybe it would be good to expand what is "host" means somewhere else on the website.

research/src/pages/checkbox.proposal.mdx
una: Naming discussion: #184

gregwhitworth: @una I get this if we change `host` to be that of `oui-host`. That said, he is saying that this isn't how he would write the component in HTML which is what is confusing me.

research/src/pages/checkbox.proposal.mdx
gregwhitworth: But the `oui-checkbox` is actually the host. Maybe, to clarify the confusion here is to have the checkbox-anatomy.js have a class of `host` rather than a tag of `<host>` so that the generated content applies?

PR #211, opened 2020-10-27
Add Lion to sources

erikkroes:
I've added the Lion components to the sources.


gregwhitworth (changes requested):
In this current form it doesn't add in any content to the site. Probably the most simplistic variant of this would be that of the `web.json5` which just ensures that it shows up in the name matrix and count on the research pages. I think this should at least get to that level so that the information is consumable.

gregwhitworth (approved):

PR #212, opened 2020-10-27
Made contribute docs more specific about where research template partial data comes from

melanierichards:


research/src/pages/contribute.mdx
```diff
@@ -206,9 +206,27 @@ pathToProposal: /components/test
...
+### Research page content
+
+After this front matter, copy in the content from the [Component Research Template](https://github.com/WICG/open-ui/blob/master/research/src/pages/component-research-template.mdx).
+
+This template includes a few optional partials you can import into the research page.
+Each of these walks through the design system JSON5 files in `/sources` to finds components where the `openUIName` value matches the `Name` value of your research page front matter.
+
+- **Anatomy:** lists out component anatomy by part,
+  if any matching components have defined an anatomy.
+- **Concept Coverage:** lists matching components according to unique names,
```
una, line 218: Not sure that the name "concept coverage" is very clear for this one. Maybe this should be a subsection under concepts, "also known as"?

melanierichards: Thanks for the review Una! I think that's out of the scope for this PR since I'm just documenting what's there w/r/t partials, but I like the idea of renaming this partial. Maybe worth opening a new issue?

una: SGTM

una (approved):

gregwhitworth (approved):

PR #223, opened 2020-11-25
feat(badge): add research

assuncaocharles:
This PR adds `Badge` component research


assuncaocharles:
@gregwhitworth Have a look into this one when you have some time :)

gregwhitworth (approved):
Only feedback that I have is that I'd possibly force or call out that pill is used for the same manning as badge by Evergreen: https://evergreen.segment.com/components/badge-and-pill/

They denote that a pill is only for numbers, not sure why they have different components though. I'd open an issue to resolve on the name of badge.

una:
Thank you for adding this research! One thing I want to also make sure is included here is directionality. For example the [Material-UI](https://material-ui.com/components/badges/) badges show multiple direction options, as well as an empty state (empty dot), which indicates *some* information but not a specific number for the badges. 

Kapture 2020-11-25 at 16 46 34

assuncaocharles:
> Thank you for adding this research! One thing I want to also make sure is included here is directionality. For example the [Material-UI](https://material-ui.com/components/badges/) badges show multiple direction options, as well as an empty state (empty dot), which indicates _some_ information but not a specific number for the badges.
> 
> Kapture 2020-11-25 at 16 46 34

Hi @una 

Added all Material examples :)

assuncaocharles:
> Only feedback that I have is that I'd possibly force or call out that pill is used for the same manning as badge by Evergreen: https://evergreen.segment.com/components/badge-and-pill/
> 
> They denote that a pill is only for numbers, not sure why they have different components though. I'd open an issue to resolve on the name of badge.

In this research I am ignoring they call it `pills` since it matches all example of `count` or `number` and evergreen is the only making this kind of distinction.

kant2002:
Examples from other design systems.

image

Taken from here: https://backpack.github.io/components/badge?platform=native

Extremely fringe case if you like. But leaving here for completeness. Can add to research if needed.

PR #233, opened 2020-12-02
Update tabs.proposal.mdx

### LINKED ISSUE #192: Tabs: Provide better description
The text in the current PR:

> *Tabs* are a set of sections of content that display one panel of content at a time. Each tab panel has an associated tab element, that when activated, displays the panel. The list of tab elements is arranged along one edge of the currently displayed panel.

I would say that from looking across 30 years worth of some different design systems we could be more specific about where they seem to agree, and link too.. 

> *Tabs* are a widget* which provides [well-defined meaning and interaction](https://www.w3.org/TR/wai-aria-practices/#tabpanel) allowing one-at-a-time, non-sequential viewing over a  series of thematically grouped, state independent and labelled sections.  They merely increase the users ability to understand and consume these bits of content.  

As far as 'along one edge' - I get it, and I'm not opposed to keeping it but I will just add that these guides once upon a time said said 'at the top'.  That wasn't really debatable.  Then came "well... or on the left".  With mobile devices, on the bottom actually became quite common as well and... I guess some are on the right too?  As hard as it is to imagine, I didn't really image those other things either, and I have seen plenty of responsive tabs that collapse these to inline to something that a lot of designers would call an 'accordion' but actually in every other way are just 'vertical tabs' (accordions in most systems/ARIA can have a multi-open nature like a series of summary/details), including key-handling and roles, really.  Single view with label and keyboard accessibility is also basically what describes a decent '[tabbed carousel](https://www.w3.org/TR/wai-aria-practices/#tabbed-carousel-elements)', I guess - which, also, you know, doesn't probably _look_ like tabs at all in a way... But I guess we need to work to where we draw the lines so I will refer back to this issue because it's hard to untangle these into which section of the document wags which other

* HTML/ARIA term of art.

bkardell:
closes #192 with the proposed text from 192 except s/widget/component as per other discussions

### DIFF
--- a/research/src/pages/tabs.proposal.mdx
+++ b/research/src/pages/tabs.proposal.mdx
@@ -9,8 +9,7 @@ showInMenu: false
 # Tabs Component Specification
 
 ## Overview <a href="#overview" id="overview"></a>
-
-*Tabs* are a set of sections of content that display one panel of content at a time. Each tab panel has an associated tab element, that when activated, displays the panel. The list of tab elements is arranged along one edge of the currently displayed panel.
+*Tabs* are a component which provides [well-defined meaning and interaction](https://www.w3.org/TR/wai-aria-practices/#tabpanel) allowing one-at-a-time, non-sequential viewing over a series of thematically grouped, state independent and labelled sections. They merely increase the users ability to understand and consume these bits of content.
 
 ### Background <a href="#background" id="background"></a>
 
--- a/research/src/pages/tabs.proposal.mdx
+++ b/research/src/pages/tabs.proposal.mdx
@@ -9,7 +9,7 @@ showInMenu: false
 # Tabs Component Specification
 
 ## Overview <a href="#overview" id="overview"></a>
-*Tabs* are a component which provides [well-defined meaning and interaction](https://www.w3.org/TR/wai-aria-practices/#tabpanel) allowing one-at-a-time, non-sequential viewing over a series of thematically grouped, state independent and labelled sections. They merely increase the users ability to understand and consume these bits of content.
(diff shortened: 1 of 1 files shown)


research/src/pages/tabs.proposal.mdx
```diff
@@ -9,8 +9,7 @@ showInMenu: false
 # Tabs Component Specification
 
 ## Overview <a href="#overview" id="overview"></a>
-
-*Tabs* are a set of sections of content that display one panel of content at a time. Each tab panel has an associated tab element, that when activated, displays the panel. The list of tab elements is arranged along one edge of the currently displayed panel.
+*Tabs* are a component which provides [well-defined meaning and interaction](https://www.w3.org/TR/wai-aria-practices/#tabpanel) allowing one-at-a-time, non-sequential viewing over a series of thematically grouped, state independent and labelled sections. They merely increase the users ability to understand and consume these bits of content.
```
una, line 12: "They merely increase the users ability to understand and consume these bits of content." <-- I prefer not to use terms like "merely". This sentence doesn't seem to help define tabs.

una:
Why not use the ARIA definition [linked](https://www.w3.org/TR/wai-aria-practices/#tabpanel) in this one?

"Tabs are a set of layered sections of content, known as tab panels, that display one panel of content at a time. Each tab panel has an associated tab element, that when activated, displays the panel. The list of tab elements is arranged along one edge of the currently displayed panel, most commonly the top edge."

bkardell:
> Why not use the ARIA definition [linked](https://www.w3.org/TR/wai-aria-practices/#tabpanel) in this one?
> 
> "Tabs are a set of layered sections of content, known as tab panels, that display one panel of content at a time. Each tab panel has an associated tab element, that when activated, displays the panel. The list of tab elements is arranged along one edge of the currently displayed panel, most commonly the top edge."

Edits made post our discussion.

una (approved):

Issue #170, opened 2020-08-28
Website doesn't display well on mobile esp contribute page


matt-auckland:
First off, love the idea behind this site I think I'll be making use of it from now on.

As for the issue, what I expect is the sidebar to be hidden on mobile, and shown via some kind of navbar menu dropdown/button.

Another thing is on pages with code blocks (for example: https://open-ui.org/contribute ) things overflow massively.

Edit: I'm just assuming it's the codeblocks and not something else, I haven't thoroughly investigated

![Screenshot_20200828-121919.jpg](https://user-images.githubusercontent.com/12672050/91507363-bbfb6580-e928-11ea-8fc1-529cb5757670.jpg)

aardrian:
Came here to report the same thing. On Android with Chrome/Firefox and on iOS, I get as the screen shot above.

Frustrating to use, but more importantly impacts credibility of the project.

aardrian:
I think it is the `, shrink-to-fit=no` in the viewport `<meta>`, but not in a position to test.

gregwhitworth:
Thanks for filing folks - yeah I've likewise noticed this but haven't had a chance to tackle it yet so thanks for raising it.

brianespinosa:
We should probably make sure this gets addressed overall with #29 too if there is a new design coming.

gregwhitworth:
@brianespinosa completely agree that said I was more thinking along the lines of a stop gap type approach, less so a full end to end

that said I appreciate you linking them - @ststimac ^ fyi

aardrian:
I cannot figure out where to edit the inline CSS, and I am not a Gatsby/etc. user, but if you want a stop-gap fix sans a PR, here is a suggestion...

For this wrapper div:
```
<div style="display:flex;padding:0 1rem;margin:0 auto;max-width:1200px;">
```

Add `flex-wrap:wrap`:
```
<div style="display:flex;padding:0 1rem;margin:0 auto;max-width:1200px;flex-wrap: wrap;">
```

For its child nav:
```
<nav style="margin-right:2em;">
```

Add `flex: 1 1 15em` (or whatever you want its default width to be, which you may want in pixels given you are using pixels on the `max-width`):
```
<nav style="margin-right:2em;flex: 1 1 15em;">
```

For the wrapper's child div / sibling to the nav:
```
<div style="flex:1;">
```

Give it the ability to shrink and grow and give it a maximum width:
```
[...]

Issue #29, opened 2019-11-20
Create better landing page


gregwhitworth:
It would be good for the landing page to do a good job of presenting the problem we're hoping to solve in a clean, simple and straightforward manner. Additionally help describe the relationship between the design systems and the web platform. We should make it easy to jump to the components. Additionally, it would be good to add the logos of companies that are involved in this project.

gregwhitworth:
Initially assigning to Chris since I think he was the one that said he could tackle it.

levithomason:
Any update @chrisdholt?  Really looking forward to this  😄 I'm also game for helping out if needed.

gregwhitworth:
Gonna ping again on this @levithomason @chrisdholt, what's a solid ETA we can put this in? Let's say just an initial design comp? Does mid-May, say the 20th work?

ststimac:
@gregwhitworth I'd be happy to pick this up if this isn't in progress yet.

gregwhitworth:
@ststimac - Goodness I'd love it! We have discussed it numerous times being a need but unfortunately it kept getting bumped. So that would be wonderful if you can create some mockups. Let me know if you want to sync about primary content on the main page.

ststimac:
@gregwhitworth yup we should sync! I've got some time tomorrow or Thursday morning then I'm taking a few days off. So if next week would be better that's fine too!

gregwhitworth:
@ststimac hey what's the update on this?

Also, @levithomason and I were chatting and as you've seen regarding the Figma comps there are more than one demo that are beginning to reference Open UI. Right now the specs are heavily leaning towards developers (whether FE, component authors, or browser implementers) but it would be good for the landing page to call attention to the various benefits Open UI brings to each.

* Developer: "Some fancy words about how it's the definitive solution blah blah"
* Designer: "Leverage design tool kits to design alongside your developers with the same expectation of states, terminology, bla blah"

This could change the overall design of the specification itself to be more designer friendly rather than a bunch of events it could focus on the anatomy, use cases, and links to figma/sketch, etc. It would be good to be able to toggle this on the spec page as well.

ststimac:
@gregwhitworth I sent you and @levithomason comps yesterday. Did they not come through?

gregwhitworth:
@ststimac it got sucked in by a rule :) Thanks

gregwhitworth:
[...]

PR #234, opened 2020-12-03
Move site nav into a menu on smaller screens

andrewiggins:
On screens up to 640px wide, the site navigation is moved into a menu that is toggled on click. For this PR, I wanted to keep it very simple to serve as a stop-gap (as suggested in #170) until the proposed re-design in #29 happens.

openui-mobile-menu

Some current limitations I'll tackle in future PRs
* The select & checkbox proposal pages as well as the component name matrix page have tables that overflow the page on small screens. I'll investigate making this mobile friendly later.
* The mobile menu isn't interactive until our JS loads which is currently a 1.2 MB (😱) minified gzipped file. I imagine on a phone with a spotty connection it may take some time to load. In another PR I'll investigate reducing this bundle size so the mobile menu can become interactive faster.

I'll add some additional context to specific changes in the comments below.

Related: #170

### DIFF
--- a/research/src/components/global.css
+++ b/research/src/components/global.css
@@ -14,10 +14,6 @@
 }
 
 @media (max-width: 640px) {
-  .header-nav {
-    display: none;
-  }
-
   .header-menu-btn[type='button'] {
     display: block;
   }
@@ -35,6 +31,27 @@
 @media (max-width: 640px) {
   .page-wrapper {
     flex-direction: column;
+  }
+}
+
+/*      Community Links    */
+.community-links {
+  display: block;
+}
+
+.community-links.mobile {
+  display: none;
+  justify-content: space-around;
+  margin-bottom: 0.5rem;
+}
+
+@media (max-width: 640px) {
+  .community-links {
+    display: none;
+  }
+
+  .community-links.mobile {
+    display: flex;
   }
 }
 
--- a/research/src/components/global.css
+++ b/research/src/components/global.css
@@ -1,3 +1,67 @@
+/*       Header     */
+.header-menu-btn[type='button'] {
+  display: none;
+  padding: 0.25em 0.5em;
+  appearance: none;
+  border: 0;
+  background: none;
+  color: white;
+  cursor: pointer;
+}
+
+.header-menu-btn[type='button']:hover {
+  background-color: #666;
+}
+
+@media (max-width: 640px) {
+  .header-nav {
+    display: none;
+  }
+
+  .header-menu-btn[type='button'] {
+    display: block;
+  }
+}
+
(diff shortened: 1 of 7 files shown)


research/src/components/global.css
```diff
@@ -1,3 +1,67 @@
...
+  background: none;
+  color: white;
+  cursor: pointer;
+}
+
+.header-menu-btn[type='button']:hover {
+  background-color: #666;
+}
+
+@media (max-width: 640px) {
```
andrewiggins, line 16: I chose 640px semi-randomly (I just used the various device emulations devtools has to see what looked good on that sampling of devices). I imagine people here have done much more research on this so feedback here would be appreciated!

research/src/components/header.js
```diff
@@ -27,19 +28,33 @@ const Header = ({ siteTitle, githubURL }) => (
         <Logo siteTitle={siteTitle} />
       </span>
 
-      <a href={githubURL} target="_blank" rel="noreferrer noopener" style={{ color: 'inherit' }}>
-        <FontAwesomeIcon style={{ marginRight: '0.2em' }} icon={faGithub} /> GitHub
-      </a>
+      <div className="header-nav">
```
andrewiggins, line 31: Oh! I just realized I forgot to add these links into the mobile nav! Lemme fix that.

research/src/components/layout.js
```diff
@@ -56,12 +56,11 @@ const components = {
...
 
 const Layout = ({ children, pageContext }) => {
-  const { frontmatter } = pageContext || {}
+  const [opened, setOpen] = React.useState(false)
+  const onToggleMenu = React.useCallback(() => setOpen((opened) => !opened), [setOpen])
 
-  const ContentWrapper =
-    frontmatter && frontmatter.path && frontmatter.path.startsWith('/components/')
-      ? ComponentLayout
-      : ({ children }) => <>{children}</>
```
andrewiggins, line 64: This line was effectively creating a new component on each render of Layout, causing React to rebuild the page each time the menu was toggled. Fixed it by just rendering children directly instead of wrapping them in a function.

andrewiggins (commented):
If you have a chance, checkout the new nav on your phone so we can get some real device testing!

https://deploy-preview-234--open-ui.netlify.app/

research/src/components/header.js
andrewiggins: Added.

gregwhitworth (approved):

gregwhitworth:
Thanks for this @andrewiggins - I filed the following to track other mobile site improvements. Feel free to recommend any others you've seen: https://github.com/WICG/open-ui/issues/239

PR #236, opened 2020-12-04
Clarify that open state can be set to get default-opened behavior

### LINKED ISSUE #229: [SELECT] Open select by default
I notice that Uber UI has interesting property on the SELECT component - `startOpen` which displays dropdown in opened state. Again, this seems to be fringe, but at least should be considered/discussed.

Referece: https://baseweb.design/components/select/

dandclark:
Per the [resolution](https://www.w3.org/2020/12/03-openui-minutes.html#resolution01) from the 3 December 2020 meeting, clarify that the `open` state will be author-visible and author-changeable by both a JS property and an HTML content attribute.  This resolves #229 since that allows authors to make a `<select>` open by default.

Additionally:
- Fix the formatting/wording of the "Default State" section, which formerly had the misleading top-level bullet "The state of the `<select>` is set to `open`".
- Update the heading of the table in the "`<select>` Properties" section to agree with the section name.

Longer term we should consider formalizing which of the properties are reflected as HTML content attributes.  Right now there isn't really anything specifying that normatively.

### DIFF
--- a/research/src/pages/select/select.proposal.mdx
+++ b/research/src/pages/select/select.proposal.mdx
@@ -28,7 +28,7 @@ then select the appropriate one for you.
 
 ## &lt;select&gt; Properties <a href="#select-properties" id="select-properties"></a>
 
-| Attribute Name                                                                                          | Type                    | Default Value                         | Description                                                                                                                                              |     |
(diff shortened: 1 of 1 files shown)


gregwhitworth (approved):

gregwhitworth:
We should loop in relevant other platform folks given the discussion at TPAC regarding the open attribute.

PR #245, opened 2021-01-12
chore(Checkbox): update spec to consider mixed value

assuncaocharles:
`indeterminate` is not an HTML property. No browser currently supports indeterminate as an attribute. It must be set via JavaScript. See Indeterminate state checkboxes for details.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate
 

`Checkbox` has only three states

`Checkbox` can be checked, unchecked and indeterminate. `checked={false}` & `intederminate={false}` will not produce a forth state that might be confusing for users.
 
https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type%3Dcheckbox)

`aria-checked` also has three states
`mixed`  is proposed because it's also implemented in WAI ARIA spec.
 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role#associated_wai-aria_roles_states_and_properties

### DIFF
--- a/research/src/pages/checkbox.proposal.mdx
+++ b/research/src/pages/checkbox.proposal.mdx
@@ -85,12 +85,12 @@ The `<checkbox>` control is primarily leveraged to indicate a binary state. For 
 
 #### States and Interactions
 
-| State Group     | States         | Initial State | Description                                                                                                                                 | Interaction/Transition                       |
-| --------------- | -------------- | ------------- | ------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------- |
-| `checked`       | `true`/`false` | `false`       | When `true`, associates the control with the parent form data and displays the control with a checked appearance if not also indeterminate. | User click. Controlled programmatically.     |
-| `indeterminate` | `true`/`false` | `false`       | When `true`, displays the control with an indeterminate appearance.                                                                         | No interaction. Controlled programmatically. |
-| `disabled`      | `true`/`false` | `false`       | When `true`, disables the control, preventing user interaction and displaying the control with a disabled appearance.                       | No interaction. Controlled programmatically. |
-| `readonly`      | `true`/`false` | `false`       | When `true`, the control's checked value cannot be changed by user interaction.                                                             | No interaction. Controlled programmatically. |
+| State Group     | States                 | Initial State | Description                                                                                                                                                           | Interaction/Transition                       |
+| --------------- | ---------------------- | ------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------- |
+| `checked`       | `true`/`false`/`mixed` | `false`       | When `true`, associates the control with the parent form data and displays the control with a checked appearance. When `mixed` appears with indeterminate appearance. | User click. Controlled programmatically.     |
+| `indeterminate` | `true`/`false`         | `false`       | When `true`, displays the control with an indeterminate appearance.                                                                                                   | No interaction. Controlled programmatically. |
+| `disabled`      | `true`/`false`         | `false`       | When `true`, disables the control, preventing user interaction and displaying the control with a disabled appearance.                                                 | No interaction. Controlled programmatically. |
+| `readonly`      | `true`/`false`         | `false`       | When `true`, the control's checked value cannot be changed by user interaction.                                                                                       | No interaction. Controlled programmatically. |
 
 #### Keyboard
 
--- a/research/src/pages/checkbox.proposal.mdx
+++ b/research/src/pages/checkbox.proposal.mdx
@@ -30,7 +30,6 @@ The `<checkbox>` control is primarily leveraged to indicate a binary state. For 
 | Attribute Name   | Type     | Default Value | Description                                                                                           |
 | ---------------- | -------- | ------------- | ----------------------------------------------------------------------------------------------------- |
 | `checked`        | `bool`   | `false`       | Controls whether the checkbox is checked or unchecked.                                                |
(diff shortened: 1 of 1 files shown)


research/src/pages/checkbox.proposal.mdx
```diff
@@ -30,7 +30,6 @@ The `<checkbox>` control is primarily leveraged to indicate a binary state. For
 | Attribute Name   | Type     | Default Value | Description                                                                                           |
 | ---------------- | -------- | ------------- | ----------------------------------------------------------------------------------------------------- |
 | `checked`        | `bool`   | `false`       | Controls whether the checkbox is checked or unchecked.                                                |
```
layershifter, line 32: Should be this also updated?

assuncaocharles: What should be the type?

assuncaocharles:
@gregwhitworth Can you have a look into this?

research/src/pages/checkbox.proposal.mdx
```diff
@@ -85,12 +84,12 @@ The `<checkbox>` control is primarily leveraged to indicate a binary state. For
...
-| `readonly`      | `true`/`false` | `false`       | When `true`, the control's checked value cannot be changed by user interaction.                                                             | No interaction. Controlled programmatically. |
+| State Group     | States                 | Initial State | Description                                                                                                                                                           | Interaction/Transition                       |
+| --------------- | ---------------------- | ------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------- |
+| `checked`       | `true`/`false`/`mixed` | `false`       | When `true`, associates the control with the parent form data and displays the control with a checked appearance. When `mixed` appears with indeterminate appearance. | User click. Controlled programmatically.     |
```
chrisdholt, line 89: I think this may depend on implementation - for checkboxes which follow the ARIA pattern, I think mixed could absolutely be a valid value because we're mapping to `aria-checked`. For a checkbox which is based on an HTML input with type checkbox, I don't think mixed would be valid here, rather the indeterminate IDL attribute would be set. Both are valid ways to build a checkbox, but I think we need to articulate that difference somehow if we're adding "mixed" into the...mix?

assuncaocharles: Sorry, it's a bit unclear from your reply what you're proposing. Are you against this change or do you have an another proposal?

chrisdholt: I think how this ends up being built fundamentally impacts whether or not mixed is a valid value. I'm totally fine denoting mixed as a potential value but it's only true if you're building an ARIA checkbox. Mixed would not be valid for the `checked` attribute of an HTML input with type of checkbox. With that, I don't think the change as it stands allows for someone to understand that nuance. I'd like to see that displayed somewhere here. It could be an asterisk next to mixed which calls this out below the table or some other form - I really don't have a preference for exactly how it's conveyed, I just think it's important to call out.

assuncaocharles: Thing is that indeterminate is not also a valid HTML attribute

chrisdholt: Correct - it is an IDL attribute though (JavaScript property): https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox#indeterminate

A bit more about IDL attributes: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#content_versus_idl_attributes

In the case this were being built on an HTML using an input with type checkbox, I would still expect it to have the `indeterminate` IDL attribute / JavaScript property set via JS. Mixed would not be valid in that case because an HTML checkbox doesn't use or need `aria-checked`.

levithomason: Bump on this as it's growing stale. Do we have any recommended solutions?

chrisdholt: @gregwhitworth I'm curious if there's ever been traction around an attribute which reflects the indeterminate IDL attribute. We currently only include indeterminate as IDL in our web components, but we've had feedback from several folks that they'd like an attribute for various reasons. Tangentially related here, but I primarily ask because if it has come up and been soundly rejected we can put it to bed - if not, perhaps it's worth a discussion in Open UI?

gregwhitworth: @chrisdholt I'm unsure which is why I recommend we fork this discussion to an issue and we can research it there. Given the discussion around `open` in `popup` there isn't a desire from browser vendors to have the attributes be actually reflective. So at most we're discussing an attribute for initial state not whether it is/isn't current indeterminate at the attribute level. This is why we would need to define the behavior of indeterminate in more detail and it has enough valid usecases that we should.

chrisdholt: Sounds good - totally in favor. I do think there's probably a couple of issues to spin off from this...maybe these aren't completely accurate, but they reflect a couple concerns (IMO unrelated to one another):

1. ARIA checkboxes are different than `input[type="checkbox"]`
2. Should there be a method of setting initial state for indeterminate via an attribute

gregwhitworth: @chrisdholt yes I think both of those are good. Mind spinning them up?

chrisdholt: Can do - likely will be later tonight or in the AM.

andrico1234:
I'm closing this issue due to inactivity, the result of [these tickets](https://github.com/openui/open-ui/pull/245#discussion_r618678424) should inform the approach we take toward a mix/indeterminate attribute

PR #249, opened 2021-01-14
Add initial reasearch for Menu

ling1726:
This PR Adds initial reasearch for the Menu control for the following major design systems:

* Ant design
* Lightning
* Evergreen
* Atlassian Atlaskit
* Bootstrap
* Fluent UI
* Stardust
* Material design
* Spectrum

I have also renamed existing `Menu` usages to `Menu List` since this is more appropriate to the type of control. The term `Menu` can refer to a different types of controls

The initial research identifies these patterns of interest:
* Basic - i.e. Menu in its most basic form
* Section- ability to divide a Menu into sections
* Nested submenus - ability to contain nested submenus
* Collapsible - ability to add collapsible sections to Menus
* Selection - ability so track selection state of a menu item

This PR should serve as a solid foundation for Menu research and sets the stage for future research into this control


research/src/sources/fabric.json5
```diff
@@ -490,6 +490,24 @@
     {
       "name": "Themes",
       "url": "https://developer.microsoft.com/en-us/fabric#/controls/web/themes"
-    }
+    },
+    {
+      name: 'Menu List',
+      url: 'https://semantic-ui.com/collections/menu.html',
```
chrisdholt, line 496: Do we not have a Fluent UI link for this yet? As an additional note, there are now multiple web implementations of Fluent UI including a web component instance. How do we determine what we pull research from? Considering that the core foundation element of the Fluent UI web component is in FAST (https://fast.design) one way to ensure both are captured in the research would be to include FAST as part of this PR rather than having multiple implementations of Fluent UI shown.

ling1726: That's an unfortunate typo. I've fixed the link.

You make a fair point about FAST as an upcoming flavour of fluent. I've added the control and concepts of the FAST menu component to this PR

research/src/sources/atlassian.json5
```diff
@@ -1,634 +1,656 @@
 {
-  "$schema": "../schemas/design-system.schema.json5",
-  "lastUpdated": "2019-08-22",
-  "name": "Atlaskit",
-  "description": "Atlassian's official UI library, built according to the Atlassian Design Guidelines.",
-  "url": "https://atlaskit.atlassian.com",
-  "by": "Atlassian",
-  "components": [
+  $schema: '../schemas/design-system.schema.json5',
```
khmakoto, line 2: It seems like there was an auto-update on this file, do you think you can check if you can remove that so we can quickly identify the actual changes?

research/src/sources/semantic.json5
```diff
@@ -1,698 +1,720 @@
 {
-  "$schema": "../schemas/design-system.schema.json5",
-  "lastUpdated": "2019-08-22",
-  "name": "Semantic UI",
-  "description": "Semantic is a development framework that helps create beautiful, responsive layouts using human-friendly HTML.",
-  "url": "https://semantic-ui.com",
-  "by": "Semantic Org",
-  "components": [
+  $schema: '../schemas/design-system.schema.json5',
```
khmakoto, line 2: Same with this file here.

khmakoto (commented):
You need to add a research page for the `Menu` under `research/pages`.

Also, it seems like now there are duplicate entries for `Menu` and `Menu List` in some files that might be worth looking into (you can check where there are duplicates in the component naming matrix).

gregwhitworth:
friendly ping @ling1726 - thanks for the reviews @chrisdholt and @khmakoto

research/src/sources/atlassian.json5
ling1726: Ah completely missed this thanks, I reverted and reapplied

research/src/sources/semantic.json5
ling1726: Ah completely missed this thanks, I reverted and reapplied

ling1726:
> You need to add a research page for the `Menu` under `research/pages`.
> 
> Also, it seems like now there are duplicate entries for `Menu` and `Menu List` in some files that might be worth looking into (you can check where there are duplicates in the component naming matrix).

@khmakoto good catch on the `Menu`, none of those entries contained any concepts or research. I've modified those usages to `Menu List` as I proposed in this PR since it is more relevant to the type of UI control. Also added extra design systems too

ling1726:
@gregwhitworth  Sorry missed @khmakoto's comments in my notifications, I've addressed all the existing issues and added more design systems. I've updated PR to describe the extra changes

yoavweiss:
@ling1726 - can you join the WICG to appease the IPR bots?

ling1726:
> @ling1726 - can you join the WICG to appease the IPR bots?

@yoavweiss I got a notification about being nominated to join the Open UI community group after requesting on the WICG portal... I'm not sure if I've missed a step here ?

yoavweiss:
@ling1726 What happens when you click this [link](https://www.w3.org/2004/01/pp-impl/80485/join)?

ling1726:
> @ling1726 What happens when you click this [link](https://www.w3.org/2004/01/pp-impl/80485/join)?

@yoavweiss  I've sent out another request (which I've done before) who should I reach out to for approval ?

yoavweiss:
@dontcallmedom - can you assist? I don't see @ling1726 in the CG...

dontcallmedom:
I'm guessing based on @ling1726 last message that their participation in the WICG needs to be approved by their AC Rep

khmakoto (approved):

ling1726:
@gregwhitworth it seems everything is running green now, what do I need to do to get some feedback and traction for PR completion ?

gregwhitworth:
@ling1726 much appreciated. @una and I do PR reviews at least once a week. I'd also recommend dropping a link into the general discord channel asking for review. @chrisdholt is another consistent reviewer and already weighed in so a friendly ping to him wouldn't hurt. At any rate, it will be reviewed this week.

una:
Thank you for this PR! A few thoughts:

1. I think `Menu` is a better descriptor than `Menu List`, which could imply a list-type of Menu items. Some of these examples are options in an interaction. Of the 10 Design Systems mentioned, none of them call this component a "Menu List". Most are "Menu"/"Menus"/"ContextMenu" etc.
2. This is interesting because I think a lot of these could fall under the `Select` definition, specifically the [Material Menu](https://material.io/components/menus/web#design-amp-api-documentation) and the [Bootstrap](https://getbootstrap.com/docs/4.0/components/dropdowns/) "Menu", which lives under "Dropup variation"

ling1726:
> Thank you for this PR! A few thoughts:
> 
> 1. I think `Menu` is a better descriptor than `Menu List`, which could imply a list-type of Menu items. Some of these examples are options in an interaction. Of the 10 Design Systems mentioned, none of them call this component a "Menu List". Most are "Menu"/"Menus"/"ContextMenu" etc.
> 2. This is interesting because I think a lot of these could fall under the `Select` definition, specifically the [Material Menu](https://material.io/components/menus/web#design-amp-api-documentation) and the [Bootstrap](https://getbootstrap.com/docs/4.0/components/dropdowns/) "Menu", which lives under "Dropup variation"

I knew this was going to come up. 

1. I wanted to specify that a `MenuList` would be a vertical list of items that can hold some sort of single or multi selection state. What I have observed is that `Menu` can be a can of worms, Semantic for example supports a horizontal `Menu` by default, while the w3c defines what is known as `Navigation menu bar` which is a combination of horizontal and vertical menus. Also, a lot of the time Menus can either be used standalone or in a popup... I'm not quite sure how to handle the overlap here so I wanted to be specific

2. While `Select` can also be adequate, very rarely do you see the term for the native HTML control actually used. `Dropdown` is also interesting but actually uses a different set of semantics and keyboard behaviour which W3 defines as `listbox`

chrisdholt (commented):
I'd like to see FAST included here as it's represented in the repo already and has support for menu - as a note, we have support for nested items in PR 😄 :

https://www.fast.design/docs/components/menu
https://explore.fast.design/components/fast-menu

On the naming note which @una brought up, `menu` and `menubar` are a pattern and standard we get from the W3C. For FAST, we elected to use menu as it maps to the W3C interaction model and is fairly common across most libraries. For the most part, my preference is to map to existing standards where we can; I think it would be confusing to have something which represents the W3C menu which is named differently.

ling1726:
@una  I have renamed to `Menu` as requested

@chrisdholt  FAST should already be represented in this PR, is something missing ? I haven't included submenu since it is still in PR, I don't think I should wait for this, can always augment research in the future

research/src/sources/fast.json5
```diff
@@ -72,7 +72,17 @@
...
+      "concepts": [
+        {
+          "name": "basic",
+          "image": "fast-menu-basic.png"
+        },
+        {
+          "name": "section",
+          "image": "fast-menu-section.png"
+        }
+      ]
```
chrisdholt, line 85: There's quite a few concepts not included here such as nested menus, start/end slots for icons and more, etc. Can we add those as research so we have a clear accounting of feature support?

chrisdholt (approved):

ling1726:
@gregwhitworth  I've updated this PR with main, could you please review ?

research/src/sources/fast.json5
ling1726: Fast submenus are represented in research in [97c0be4](https://github.com/openui/open-ui/pull/249/commits/97c0be4a5610957f716029c019c86a0ae7657765)

chrisdholt (approved):
#blessed @andrico1234 :)

PR #250, opened 2021-01-19
Add Avatar research page

behowell:
Add the research for the `Avatar` control from the design systems that support it.


yoavweiss:
@behowell - can you join the WICG to appease the IPR bots?

khmakoto:
Shouldn't the OpenUI names for `label` and `badge` be `initials` and `presence` respectively since they are the most used ones out there?

image

image

khmakoto (commented):
Shouldn't the OpenUI names for `label` and `badge` be `initials` and `presence` respectively since they are the most used ones out there?

image

image

chrisdholt:
> Shouldn't the OpenUI names for `label` and `badge` be `initials` and `presence` respectively since they are the most used ones out there?
> 
> image
> 
> image

I'm not sure the right thing is always illustrated by what is most common. In cases where there is strong consensus, I think that's probably a fair assessment. I'm not sure what's currently shown above though would qualify as strong consensus.

khmakoto:
@chrisdholt fair, I'd say 3 is starting to get into consensus territory though (not strong, but consensus), especially if all the other ones only have 1 instance. For badge I feel like it's ok.

behowell:
@khmakoto @chrisdholt Since there wasn't a strong consensus between the names for label/initials, and badge/presence I picked the more generic names that worked for all of the use cases I saw.

For `label`, there was one design system (Ant Design) that allowed other labels besides just initials (such as a full name). The term `label` seemed to encompass all of the different designs without creating artificial limits to only show initials.

For `badge`, it seemed like the more generic name but it wasn't entirely clear. It also aligns with the OpenUI `Badge` component, although there's nothing saying that the two need to match. Does anyone else have an opinion about what the OpenUI name should be?

I also realize that Atlaskit uses both "presence" and "status". I'll add that as well.

khmakoto (approved):

una:
Generally LGTM 👍

una (approved):

behowell:
Thanks @una! Is there anyone else who needs to sign off before this is merged? It looks like I don't have permissions to merge it myself.

PR #261, opened 2021-02-04
Fix link to fast-select component

kaelig:
The URL may have changed? Seems like the correct one is https://explore.fast.design/components/fast-select


chrisdholt (approved):

kaelig:
@chrisdholt I'm unable to merge this myself (probably because I'm an external contributor), can you do it?

andrico1234 (approved):

PR #262, opened 2021-02-04
feat(Pills): add pills proposal

assuncaocharles:
This PR adds `Pills` proposal


una:
Naming is hard. There is already a `pills` section within `Badges`. Another common naming for these is "chips" or "tags". How can we discern the differences between all of these types? Should we choose a single naming or break them out into subgroups? 

I opened #266 to discuss

una:
Discussed on 2/25 that "Pill" is a naming convention that is tied to a specific style (rounded corners). We want to avoid any style-concerns in naming, and focus on semantics. Working on discussing the differences and naming in #266

una:
Resolution to rename these Tags: https://www.w3.org/2021/03/04-openui-minutes.html#r01

assuncaocharles:
@una Should I just update the name for `Tag` then?

PR #263, opened 2021-02-10
feat(Accordion): adds accordion research 

bsunderhus:
This PR adds `Accordion` component research


yoavweiss:
@bsunderhus - can you join the WICG to appease the IPR bots?

research/src/sources/fast.json5
```diff
@@ -8,7 +8,29 @@
...
+          "image": "fast-accordion-open.png"
+        },
+        {
+          "name": "collapse",
+          "description": "No restraining in the amount of collapsed items",
+          "image": "fast-accordion-collapse.gif"
+        },
+        {
+          "name": "accordion",
+          "description": "Only one opened item",
```
chrisdholt, line 30: Technically there both multi and single are supported under the `mode` attribute.

research/src/sources/fast.json5
```diff
@@ -8,7 +8,29 @@
...
+          "image": "fast-accordion-accordion.gif"
+        }
+      ]
```
chrisdholt, line 33: There are a few things here that may be difficult to account for in terms of research which may not be immediately visible if you check out the explorer (explore.fast.design). It may be beneficial to include these as part of the research.

One that should probably be considered from a research perspective is the support for slots for expand or collapse glyphs. Note: Technically these could be a single slot which changes based on state but for the time being we have them separate.

Similarly, there is support for content which precedes or follows the heading content of the accordion (start/end slots) which could include a glyph for example. Might also be good to represent - it was a common scenario we found across design systems and needs to be accounted for from a template perspective so it may be good to include here.

bsunderhus: > One that should probably be considered from a research perspective is the support for slots for expand or collapse glyphs. Note: Technically these could be a single slot which changes based on state but for the time being we have them separate.

That is a good concept that to be honest most of the libraries researched allow to change the icon indicator, I guess I simply let that pass by because I thought it was trivial. Thanks for the heads up! I'll include it right away.

Do you have any suggestion for the name of the concept @chrisdholt? perhaps `icon`,  `glyph` or `indicator`....

bsunderhus: > Similarly, there is support for content which precedes or follows the heading content of the accordion (start/end slots) which could include a glyph for example. Might also be good to represent - it was a common scenario we found across design systems and needs to be accounted for from a template perspective so it may be good to include here.

That is already represented as the `extra` concept, but I guess I could specialize it better with start/end

chrisdholt: We went with `expand-glyph` and `collapse-glyph` but as I mentioned I think an argument could be made for a single slot. Our use of start/end is really for localization sake as it aligns with CSS logical properties - the generic naming of the slot is intended to convey that the content can be "anything" and the slot is more about its position than anything else. Thanks for updating!

una:
For things like "borderless", does that add context to accordion or is it purely presentational?

bsunderhus:
> For things like "borderless", does that add context to accordion or is it purely presentational?

IMO no, I don't even think it's a useful concept. But it's a concept from Ant Design 🤷🏼‍♂️.  It's purely presentational.

bsunderhus:
> @bsunderhus - can you join the WICG to appease the IPR bots?

I'm in contact with my organisation reggarding this @yoavweiss, apparently I can't solve this by myself.

andrico1234:
Hi @bsunderhus , I'm closing this due to inactivity, but feel free to reopen at any time 😄

PR #264, opened 2021-02-12
Add initial research for Flex component

andrefcdias:
This PR adds initial research for the Flex layout component for the following design systems:

- Ant Design
- Atlaskit
- Cosmos
- Bootstrap
- Evergreen
- Fluent UI - Fabric
- Fluent UI - Stardust
- Primer

Given the vast lack of consistency in naming, `Flex` should be taken as a placeholder name for the component.
Some design systems like Atlaskit, FAST or Material Design name their component as `Grid` but also offering a fluent layout system like Bootstrap and Fluent UI give with Flex, creating further discrepancies across the board.


research/src/sources/fast.json5
```diff
@@ -105,6 +105,12 @@
     {
       "name": "Tree view",
       "url": "https://explore.fast.design/components/fast-tree-view"
+    },
+    {
+      "name": "Data Grid",
+      "openUIName": "Flex",
+      "definition": "The data grid component enables authors to display an array of data in a tabular layout. The layout can include a header region where a title can be displayed for each column.",
+      "url": "https://explore.fast.design/components/fast-data-grid"
```
chrisdholt, line 113: This seems out of place - I think the intent of this control is fundamentally different than the other presentational ones listed here.

research/src/sources/w3.json5
```diff
@@ -106,6 +106,12 @@
       "definition": "Interactive Tabular Data and Layout Containers",
       "url": "https://www.w3.org/TR/wai-aria-practices/#grid"
     },
+    {
+      "name": "Layout Grid",
+      "openUIName": "Flex",
+      "definition": "A grid widget is a container that enables users to navigate the information or interactive elements it contains using directional navigation keys, such as arrow keys, Home, and End. As a generic container widget that offers flexible keyboard navigation, it can serve a wide variety of needs. It can be used for purposes as simple as grouping a collection of checkboxes or navigation links or as complex as creating a full-featured spreadsheet application. While the words row and column are used in the names of WAI-ARIA attributes and by assistive technologies when describing and presenting the logical structure of elements with the grid role, using the grid role on an element does not necessarily imply that its visual presentation is tabular.",
+      "url": "https://www.w3.org/TR/wai-aria-practices/examples/grid/LayoutGrids.html"
+    },
```
chrisdholt, line 114: I think this is represented in lines 105-107, though the layout portion is essentially a sub-type of the "Grid" widget pattern from the W3C. I'm not sure we should duplicate with a more specific example.

chrisdholt (commented):
I'm not sure that the W3C example of layout grid (or the FAST Data Grid built to support the "data grid" from that specification) are the same as the research being added above for supporting various page layouts. Most of the grid's represented in the research here seem purely presentational with the intent to serve the layout and responsive story of a given design system. 

Layout and Data Grids from a W3C standpoint have a fundamentally different purpose than what I'm seeing from most of the examples included in the research here. The intent is not just how things lay out and respond on the screen, but includes specific ARIA roles and attributes as well as a defined interaction model. I definitely see value in gathering research for things related to the W3C layout and data grids, I’m just not sure it’s best represented by what I see here.

If the intent here is to capture research for purely presentational “layout” components which the majority of data here represents, I’m not sure there is a ton of value from an Open UI standpoint. I think the platform already largely provides a robust avenue to accomplish presentational layouts with things like Grid and Flexbox. While I think there can be value in creating abstractions from an individual design system standpoint, I’m not sure it makes sense to seek alignment with those or standardize on an abstraction of the platform.

research/src/sources/fast.json5
andrefcdias: You're right, I guess my reasoning here was to gather all layout capabilities of the different design systems and I ended up misreading the purpose of Data Grid. I agree that this is more of a table layout than any other thing. Will make the changes.

research/src/sources/w3.json5
andrefcdias: I don't feel like `Flex` would really represent the W3C's `Grid` widget definition, so perhaps it's best just to remove this connection between them. What do you think?

andrefcdias:
Hey @chrisdholt, thanks for the feedback! I totally agree with you on that, hence the complications I had trying to see where `Flex` fit within the ecosystem. I've also went ahead and removed the connection to `Data Grid`, that was an oversight on my side.
Did I understand correctly that the idea of `Flex` doesn't fit with the Open UI mission given it's mostly an abstraction of CSS flex? If so, mostly out of curiosity, how do you feel about where Material UI and other `Grid` component implementations fit in?

levithomason:
Hey @andrefcdias, thanks much for the contribution. I agree with Chris on the point of specialized grids, like a datagrid, falling out of scope of a "Flex" component. It is also true that this component is not likely to be one to influence the next iteration of browser platforms.  However, Open UI's mission also includes taking a descriptivist approach to documenting all UI systems' components as they have naturally come to be. That includes common layout components.

On that note, Evergreen also has the idea of a `Box` that I think fits well in this domain of "purely layout components". I don't see a page for it on their public site, but, it is referenced a lot on their [Layout Primitives](https://evergreen.segment.com/components/layout-primitives/) page and there is also a dedicated page on their Github for it: https://github.com/segmentio/ui-box.

levithomason:
Images of these components may be tough since they are usually not visibly styled, however, have a look at the contributing docs for [adding images]( https://open-ui.org/contribute#adding-images-to-a-component) and let's see if we can get some added to this PR.

research/src/sources/w3.json5
andrefcdias: I've gone ahead and removed Layout Grid

levithomason (approved):
This looks good to me for our initial iteration. Thanks again @andrefcdias.

PR #270, opened 2021-02-21
Resolve React console errors

mrmckeb:
This PR updates anatomies to use React components, and resolves console errors around components and classnames.

### DIFF
--- a/research/src/components/anatomy-components.js
+++ b/research/src/components/anatomy-components.js
@@ -1,0 +1,28 @@
+import React from 'react'
+import './anatomy.css'
+
+export const AnatomyWrapper = ({ children }) => (
+  <div className="component-anatomy-wrapper">
+    <input type="checkbox" id="show-slots" />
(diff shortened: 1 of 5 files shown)


mrmckeb:
@gregwhitworth I'm not sure if we want to keep `data-slot` in or not. It was only used by `openui-select`, and is meant to indicate that the a slot is available - but we also have the checkbox to show slots, so I'm not sure if this is needed.

gregwhitworth:
Not sure what errors you're seeing - all I ever see is ESLint errors and still see them. Regarding the data-slot I think this was from historical need either in the actual webcomponent I built or for toggling the slot. Either way it can probably be removed.

gregwhitworth (approved):

una (approved):

mrmckeb:
Hey @gregwhitworth, I was seeing the below errors. This example is from `components/checkbox`.
image

PR #272, opened 2021-02-26
add link to minutes

wseltzer:
minutes generated post hoc from irc log


wseltzer:
noticed that rrsagent wasn't on the channel, so generated minutes from the log

gregwhitworth (approved):

Issue #257, opened 2021-02-02
Should Link have a definition proposal or a full proposal?


khmakoto:
An issue was recently raised in #253 about whether the `Link` component should have only a definition proposal instead of a full proposal given the limited utilization it has across component libraries. This issue is made to keep track of further development on this as well as to serve as a forum for discussion on the matter.

una:
I'd like to understand where `Link` needs defined anatomy beyond what exists with the `<a>` tag? My initial reaction is that we might not need a proposal for the definition of `Link`.

una:
Resolved at OpenUI: Link will not have separate Research/Definition

PR #274, opened 2021-03-01
research: add research section on Image component

kubkon:
This PR adds some basic research on the concept of an `Image` component. Since this is my first PR to this repo and on the topic of OpenUI so I would appreciate any and all feedback.


gregwhitworth:
@kubkon thanks for the PR. As it relates specifically to the image component - what's your next steps for the research?

gregwhitworth (commented):
Ultimately, I think this one will need it's own discussion - similar to the link discussion in #257 . I recommend filing an issue and determining what your desired next step for the image component would be.

kubkon:
@gregwhitworth issue submitted! Here's the link https://github.com/WICG/open-ui/issues/278.

kubkon:
Also, if we decide not to go ahead with a dedicated research page, I think it might be worth adding a section to the contributing guide to pay special attention to components that might not qualify for a dedicated page such as image (#278) or link (#257). Instead, I guess we should recommend the researchers to submit a discovery issue to garner some feedback before jumping in to submitting a PR. Lemme know if that makes sense, and if so, I'll be more than happy to follow it through!

gregwhitworth:
> Instead, I guess we should recommend the researchers to submit a discovery issue to garner some feedback before jumping in to submitting a PR. Lemme know if that makes sense, and if so, I'll be more than happy to follow it through!

@kubkon that is a wonderful idea and we should add this as a pre-req step for research as you're right we've had this happen a few times. Granted, at times the research may prove why there is a need for a component to be defined within Open UI but that can be discussed on the issue prior to the actual research being done. Great idea!!!

cc: @una

levithomason:
@gregwhitworth / @una Yes, as Open UI has the responsibility to descriptively document UI as it has naturally evolved in the industry we should then have a research page for any component scope that exists in reputable sources. As there is an `Image` component in many design systems and component libraries, we should then have an associated research page that documents that component.

It may turn out that some component research pages do not reveal immediate changes needed in platforms. That is OK, it just means the component scope is already handled by the platform.

research/src/sources/bootstrap.json5
```diff
@@ -421,6 +421,41 @@
...
       ]
+    },
+    {
+      "name": "Img",
+      "openUIName": "Image",
+      "url": "https://getbootstrap.com/docs/4.0/content/images/",
+      "concepts": [
+        {
+          "name": "class-img-fluid",
+          "openUIName": "fit",
```
levithomason, line 432: Although the HTML class name is `img-fluid` we want to enter the concept name as just `fluid`. This is because the `img-` prefix is not relevant to the concept being applied to image, but is just a prefix for this component name. The concept being applied to the image is `fluid`.

This will apply to all concepts named `class-*`.

research/src/sources/bootstrap.json5
```diff
@@ -421,6 +421,41 @@
...
+          "description": "Makes image responsive",
+          "image": "bootstrap-image-fluid.png"
```
levithomason, line 434: This image also includes a code sample of HTML. Let's trim it down to just the component itself.

research/src/sources/semantic.json5
```diff
@@ -236,7 +236,68 @@
...
+        {
+          "name": "default",
+          "description": "Unless size is specified, image will use its original dimensions filling the container",
+          "image": "semantic-image-default.png"
+        },
+        {
+          "name": "fluid",
+          "openUIName": "fit",
+          "description": "An image can take up the width of its container",
+          "image": "semantic-image-fluid.png"
```
levithomason, line 251: This image includes a frame around the image and an HTML snippet. Let's trim it down to just an image of the component.

research/src/sources/web.json5
```diff
@@ -93,11 +93,18 @@
...
-      "url": "https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element"
-    },
-    {
-      "name": "Img",
-      "url": "https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element"
+      "openUIName": "Image",
+      "url": "https://html.spec.whatwg.org/multipage/embedded-content.html#the-img-element",
+      "concepts": [
+        {
+          "name": "width",
```
levithomason, line 100: The width and height image examples are of API docs. It would be ideal to have component specimens instead. I can see why this would be difficult to capture as there isn't really a great visual for width/height.

If there aren't actual visual specimens for width/height I wonder if we should exclude them? It just seems strange to see API docs in the visual examples area.

image

levithomason (approved):
This is good to merge after addressing any feedback you deem necessary. I don't want to block this PR.

research/src/sources/web.json5
kubkon: Yep, great points, thanks @levithomason! I was on the fence whether to include them myself, so happy to remove them since they don't really add much value!

research/src/sources/bootstrap.json5
kubkon: Makes sense, thanks!

kubkon:
@levithomason comments addressed. Have a look and lemme know what you reckon!

levithomason:
Thanks for sticking with it @kubkon. Let's merge and we can iterate further if needed. Much appreciated.

PR #294, opened 2021-03-19
Port over popup explainer to Open UI for incubation

melanierichards:
The purpose of this PR is to quickly port over the [popup proposal from Microsoft Edge explainers](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/Popup/explainer.md) to Open UI for standards incubation. This is based on a recent group resolution to take up `<popup>` and related APIs as an "official" work item.

This PR creates:

* A very small research index for popup. This will later be supplemented with content from #205 (smaller scope of transient-UI types than that current PR)
* A copy of the original popup explainer, in Open UI research directories.

I'm current hot-linking to images from the original explainer as I think the project needs a way to deal with/host images that are not part of the "Concepts" grids.

### DIFF
--- a/research/src/pages/popup/popup.research.mdx
+++ b/research/src/pages/popup/popup.research.mdx
@@ -1,0 +1,11 @@
+---
+name: Popup
+path: /components/popup.research
+menu: Analysis
+---
+
+This research page will be expanded to detail various different subclasses of "popup" UI.
+
+## Additional research
+
+- [Initial proposal for a new `<popup>` element and related APIs](/components/popup.research.explainer)
--- a/research/src/pages/popup/popup.research.mdx
+++ b/research/src/pages/popup/popup.research.mdx
@@ -4,8 +4,10 @@ path: /components/popup.research
 menu: Analysis
 ---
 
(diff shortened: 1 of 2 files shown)


chrisdholt (approved):
Love it!

research/src/pages/popup/popup.research.mdx
```diff
@@ -0,0 +1,11 @@
...
+name: Popup
+path: /components/popup.research
+menu: Analysis
+---
+
+This research page will be expanded to detail various different subclasses of "popup" UI.
+
+## Additional research
+
+- [Initial proposal for a new `<popup>` element and related APIs](/components/popup.research.explainer)
```
gregwhitworth, line 11: Can we just call this Explainer? Given that the term `proposal` (we should change this) in Open UI currently is used to denote the more normative spec text I'd prefer to not duplicate this name even though it is of course moving towards proposal.

gregwhitworth (approved):
Approved with the feedback to change the link text to just be Explainer. As a slight aside, in speaking with Brian regarding tabs I do like the premise of these landing pages having a short definition of what the component is. If you're able to add a few sentences stolen from your explainer that defines the `<popup>` component's UX that would be great too but I won't block the PR on that.

melanierichards:
Thanks @gregwhitworth, your feedback should be addressed! Was planning to add the summary/description with the research PR but actually this is a good time for that too. :)

PR #295, opened 2021-03-19
Move research into the root directory

gregwhitworth:
When Open UI was first created there was an expectation of platform and research to be somewhat seperated. This turned out not to be the case and ultimately all work has happened in the research subfolder and not all of it is research. Given this evolution I have moved the research content into the root and deleted the platform subfolder. I also updated the readme to point to the sources that reflect our up to date contribution and charter content.

For this PR I will give a heads up to the group on a call prior to landing it since people are used to going to the research subfolder to run the site.


EisenbergEffect:
This PR is quite old now. Do we want to close it or refresh it?

andrico1234:
I'm going to close this, as this work will be done as part of Phase 2 of the site refactor

Issue #314, opened 2021-04-06
[Navigation Architecture] Create a site map page for the Open UI site


melanierichards:
In addition to the IA restructure we should think about providing a site map, which will help provide [multiple ways](https://www.w3.org/WAI/WCAG22/Understanding/multiple-ways.html) of navigating the Open UI site.

PR #340, opened 2021-04-22
[Navigation Architecture] Create a site map page for the Open UI site

andrico1234:
Uses `gatsby-plugin-sitemap` to generate the site's sitemap. 

closes: https://github.com/openui/open-ui/issues/314

The documentation for the plugin is [here](https://www.gatsbyjs.com/plugins/gatsby-plugin-sitemap) incase we want to use some of the non-default parameters. But the defaults seem sensible enough. 

It's worth noting that the 404 page is excluded by default. Are there any other pages we'd want to exclude?

You can test this yourselves by running gatsby build, and then opening the generated `sitemap.xml` in the public folder. 

Screenshot 2021-04-22 at 08 10 54

### DIFF
--- a/research/gatsby-config.js
+++ b/research/gatsby-config.js
@@ -7,6 +7,7 @@ module.exports = {
     description: 'Open UI ',
     author: 'Open UI',
     githubURL: 'https://github.com/WICG/open-ui',
+    siteUrl: 'https://open-ui.org/',
   },
 
   plugins: [
@@ -66,6 +67,8 @@ module.exports = {
     // },
 
     'gatsby-plugin-remove-trailing-slashes',
+
+    `gatsby-plugin-sitemap`,
 
     {
       resolve: `gatsby-plugin-typography`,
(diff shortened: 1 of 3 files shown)


melanierichards:
Hi @andrico1234, apologies for the delayed review. My understanding is that this change outputs only the xml file, right? If so, I'd be fine with us merging this change, but we shouldn't auto-close #314 as ideally we should provide a person-readable version of the site map into addition the machine-readable format. :)

andrico1234:
No problem, did you have something like this in mind?

https://gatsby.ghost.org/sitemap.xml

melanierichards (approved):

melanierichards:
> No problem, did you have something like this in mind?
> 
> https://gatsby.ghost.org/sitemap.xml

Yep! I think ideally it would be great to have the whole human-readable site map in one flat layer / page, but that's what I'm thinking. Just approved this PR, working on the human-readable version in a fresh PR sounds good.

@gregwhitworth @una any review you want to do on this one? Looks like I can merge this but the repo would prefer a chair approval. :)

andrico1234:
Hey all, is there an update on whether this can be merged in?

melanierichards:
Chatted with the chairs offline, this is good to merge. Going ahead now. Thank you Andrico!

PR #349, opened 2021-05-23
Research: Radio Button

leolopes:
This PR includes conceptual research on Radio Buttons / Radio Button Groups, and also some problem discussions that should be addressed on a telecon.

Some discussion can be found here: [https://github.com/WICG/open-ui/issues/251](https://github.com/WICG/open-ui/issues/251)


research/src/pages/radio-button.research.mdx
```diff
@@ -0,0 +1,122 @@
...
+import Image from '../components/image'
+import Anatomy from '../components/anatomy'
+import Concepts from '../components/concepts'
+import ComponentCoverage from '../components/component-coverage'
+
+<ComponentCoverage component="Radio Button" />
+
+## Introduction
+
+_(For a lenghtier discussion, on which this page was based, please refer to the [original radio button issue](https://github.com/WICG/open-ui/issues/251))_
```
una, line 16:
```suggestion
_(For a lengthier discussion, on which this page was based, please refer to the [original radio button issue](https://github.com/WICG/open-ui/issues/251))_
```

research/src/pages/radio-button.research.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+The research showed that different DS’s treat radio buttons in one of two ways, i.e.:
```
una, line 18:
```suggestion
The research showed that different design systems treat radio buttons in one of two ways, i.e.:
```

research/src/pages/radio-button.research.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+- as the main component, hosted on a “**radio button group**” component;
```
una, line 20: I don't understand the difference between these bullets, maybe this can be reworded to be a bit more clear?

research/src/pages/radio-button.research.mdx
```diff
@@ -0,0 +1,122 @@
...
+- as a part of a “**radio button group**” component, which, in fact, is the actual component.
+
+The two approaches seem very similar at first, but the former seems to indicate that single “radio buttons” could exist. The latter, however, shifts the scope up to the “radio button group”, which, although nonexistent as an HTML element, is an important concept for the W3 specification:
```
una, line 23:
```suggestion
The two approaches seem very similar at first, but the former seems to indicate that single “radio buttons” could exist to indicate a binary state (on or off). The latter, however, shifts the scope up to the “radio button group”, which, although nonexistent as an HTML element, is an important concept for the W3 specification:
```

una (commented):
Thank you for the research, I just added a few initial comments!

research/src/pages/radio-button.research.mdx
leolopes: Hi @una !
You see how this is kind of pointless, because binary states can be accomplished by using checkboxes. I tend to think people do this by mistake, so I would reword your suggestion to "(to **probably** indicate a binary state (on off))".
What do you think?

research/src/pages/radio-button.research.mdx
leolopes: I reworded it, maybe it's easier to understand now.

una (approved):

leolopes:
Thanks, @una :)

Issue #251, opened 2021-01-19
Radio


leolopes:
Just opening this issue to tell I'm going to start research work for the "radio" component.
This is the link for the document draft: https://docs.google.com/document/d/1-TyGOSOcyV4VOqNyUDjEQLjrtssI8cd2-BVld9cJtlo/edit?usp=sharing 

Any contributions are welcome :)

leolopes:
Today I started the document by making a list of references to all the concepts that remind of radio buttons in many design systems.

After that I started a spreadsheet inventory to list all the concepts related to each of the components.

The main points that are still not clear to me and maybe should be better defined in this beginning are:

- whether we should work with the concept of Radio Buttons or Radio Groups (or both separately). In my document you can see that design systems are not consistent with this naming convention, in a variety of ways.
- whether I should list every property that every design system declares as "concepts", or maybe only list concepts that seem more relevant.

If you look at the spredsheet, even though I have for now listed Ant Design's concepts alone, the list of concepts is already big and has the added complexity of being divided by "radio buttons" and "radio button groups".

I would like any suggestion you might have for how I should handle it further.

gregwhitworth:
@leolopes moving the discussion around radio from discord to here:

>  I would appreciate your input, because though radios look similar to checkboxes, it seems they are a bit more complex

Can you expound? I would expect there to be an additional section for how radio groups are tied together via the name attribute but most of the additional behaviors should be very similar outside of potentially a containing element to hold the group (assuming this is consistent across UAs/OSes/Design Systems). I have opinions on having the containing element for the group but that should be in a different issue.

chrisdholt:
> @leolopes moving the discussion around radio from discord to here:
> 
> > I would appreciate your input, because though radios look similar to checkboxes, it seems they are a bit more complex
> 
> Can you expound? I would expect there to be an additional section for how radio groups are tied together via the name attribute but most of the additional behaviors should be very similar outside of potentially a containing element to hold the group (assuming this is consistent across UAs/OSes/Design Systems). I have opinions on having the containing element for the group but that should be in a different issue.

There is a bit of nuance here I think depending how this is implemented, specifically as it relates to accessibility and behavior depending on how these are built. While the browser ties inputs with a type of radio together using the name attribute, per the spec, ARIA radios behave different and are tied together by being a child of a `radiogroup`. Depending on the implementation I would expect a bit of difference between the implementation.

leolopes:
[...]

PR #360, opened 2021-06-20
[Radio Button] Proposal (Editor's draft)

leolopes:
Related to: https://github.com/openui/open-ui/issues/251 

This PR is still a draft for the radio button proposal. I managed to create two components, one for the **radio button** and other for the **radio button group**, and the two would be all it takes to create both a rigid and a flexible layout.

The idea expounded on the proposal, but the gist is that we can either create a group with options _inside_ of it, or a group with options _outside_ of it, that are then bound to the group by a new attribute.

By leveraging slot defaults/fallbacks and proper CSS, we can make basically any layout use case.

===

I would like to validate the basic proposal with the community before developing it too much (you can see I didn't provide much information yet on accessibility, behavior, etc). If this is a valid proposal, I'll carry on.

===

Also, I'm unsure whether I used the "slot logic" correctly... In the group structure, you can see that I put a `div` inside the `choices` slot, but maybe the slot should be a direct parent of the radio buttons... If the developer so chooses, then they could fill the slot with a `div` and then the options. I'm not sure and I'd like your opinion as well.


leolopes:
Hi @una, @gregwhitworth and @levithomason, I am afraid this PR can become a little stale, but I wouldn't like to go very far before validating the main ideas addresses on it. What could we do to proceed?
Thanks.

gregwhitworth:
Thanks for filing this @leolopes - I've added @chrisdholt to do a review on this. Also, I may have missed it but did we discuss this anatomy and get resolution on it?

research/src/pages/radio-button/radio-button.proposal.mdx
```diff
@@ -0,0 +1,185 @@
...
+The following structure represents a group with the `choices` slot filled with radio buttons (their internal structures are omitted for brevity).
+
+```html
+<oui-radio-button-group name="group-name">
+  <slot name="label">
+    <label part="label"></label>
+  </slot>
+  <slot name="required-indicator">
+    <div part="required-indicator"></div>
+  </slot>
```
chrisdholt, line 131: This is interesting and something I hadn't thought of for other form elements which include labels and have a required attribute. @gregwhitworth we should look at adding this holistically where applicable.

research/src/pages/radio-button/radio-button.proposal.mdx
```diff
@@ -0,0 +1,185 @@
...
+    </div>
+  </slot>
+</oui-radio-button-group>
+```
+
+#### DOM Structure (Explicit Grouping)
+
+The following structure, in contrast, represents a group with an empty `choices` slot. The choices are placed outside the group, providing layout flexibility. Each choice has a `group` attribute with value equal to the radio button group's name, making them explicitly bound as choices of that group.
+
+There is also the possibility of having both implicit and explicit choices.
```
chrisdholt, line 147: As noted elsewhere, so long as we can produce the correct structure for AT, I love this.

research/src/pages/radio-button/radio-button.proposal.mdx
```diff
@@ -0,0 +1,185 @@
...
+```
+
+#### Slots
+
+| Slot Name            | Description                                                                      | Fallback Content |
+| -------------------- | -------------------------------------------------------------------------------- | ---------------- |
+| `label`              | Add custom markup for the control's label                                        | Empty            |
+| `required-indicator` | Content to indicate the radio button group is required to have a selected option | '\*'             |
+| `choices`            | Slot for placing the radio buttons                                               | Empty            |
+| `error-message`      | Content of any error messages                                                    | Empty            |
```
chrisdholt, line 176: Is the error message UI something that should be handled by default at the browser level or is that something that is better handled via composition?

chrisdholt:
> Thanks for filing this @leolopes - I've added @chrisdholt to do a review on this. Also, I may have missed it but did we discuss this anatomy and get resolution on it?

@leolopes this is great! @gregwhitworth I don't think we discussed and resolved. There are a couple super interesting things I think we should discuss though and get resolution on so we can move this forward!

research/src/pages/radio-button/radio-button.proposal.mdx
leolopes: I won't take credit for this idea because I just took what was already being addressed for checkboxes: https://open-ui.org/components/checkbox.research.concepts 

However, I'm not sure if this is a good idea, if it's going to be only a visual indicator. Maybe if it could receive a custom text that would be read specifically to represent the "requiredness" of the input, then it would make more sense. What do you think?

leolopes:
Hello reviewers!
Can we proceed with this PR or maybe it is too stale to be merged at this point?

research/src/pages/radio-button/radio-button.proposal.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+- [Radio input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio)
+- [WAI-ARIA Radio](https://www.w3.org/TR/wai-aria-practices/#radiobutton)
+- [WHATWG](<https://html.spec.whatwg.org/multipage/input.html#radio-button-state-(type=radio)>)
+
+[Further examples on research](/components/radio-button.research).
+
+## Radio Button Design
+
+This section relates to a single `<oui-radio-button>`. Because it cannot stand on its own, without being part of and either explicit or implicit `<oui-radio-button-group>`, it does not have attributes such `required` or `readonly`, those being attributes of the group. A single `required` option is a group makes the other choices pointless, while a `readonly` option makes itself pointles.
```
scottaohara, line 35: re: the allowance of the required attribute on a single radio - this makes sense in the context of a `radiogroup` where the only children of the group would be expected to be radio buttons, and at the parent level the requiredness of the grouping could be declared. 

However, current radio buttons in HTML have greater flexibility in where they can be placed / associated with each other, and can still be communicated as required with this flexibility.

for example:

```
<label><input type=radio name=r required> Choice 1</label>
Other information or form fields, or both, associated with making choice 1.
...
<label><input type=radio name=r> Choice 2</label>
Other information or form fields, or both, associated with making choice 2.
```

This would need to be thought out more in how to communicate the required nature of radios that don't neatly fit into a grouping (see my next comment for more on that) or a merging of what HTML presently allows (putting the required attribute on a radio button) and what the ARIA radio/radiogroup roles expect (that if you're using radios - particularly when a choice is required - they are part of a radiogroup, is more limiting than HTML)

research/src/pages/radio-button/radio-button.proposal.mdx
scottaohara: The use of the `label` element initially threw me off here as I assumed this was initially a discussion about the radio button, rather than the label (`legend`) for the radiogroup.

In either case though, this seems like something that would be nice to auto-include for people based on the requiredness of the group or individual form control.  If this markup translates to how this would actually render, being outside of the `label` / `legend`, it wouldn't be part of the accessible name of either - which is generally what would be desired as you don't want to have duplicate "required" announcements (if one were to put 'required' text into this slot).  If this indicator would be expected to be nested in the label/legend of the shadow dom of the component, e.g., 
```html
<fieldset role=radiogroup aria-required=true>
  <legend>[ label part ] [ required-indicator part ]</legend>
  ...
  or 
  <label for=x> [ label part ] [ required-indicator part ]</label>
  <input required id=x type=...>
</fieldset>
```
Then generally it _would_ be expected that the required indicator (particularly if it were text) be treated as visual only, so that the "required" state wasn't communicated twice as part of the state of the control AND in the accessible name of the control.

research/src/pages/radio-button/radio-button.proposal.mdx
```diff
@@ -0,0 +1,185 @@
...
+
+The radio button has, however a `disabled` attribute, meaning it's a possible choice, but that is currently unavailable.
+
+If used inside a `<oui-radio-button-group>`, it gets implicitly bound to the group, becoming one of the choices. If declared outside of it, it should contain a `group` attribute equal to the group's `name` attribute.
```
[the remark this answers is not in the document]
leolopes: Hello @scottaohara , how are you?

During conversations about this proposal, a topic came up, in which the `name` attribute might be reserved for other purposes, or at least should be re-thought.

Also, because the proposal changes a lot of other aspects, it is expected that it's not exactly like the current standard.

research/src/pages/radio-button/radio-button.proposal.mdx
leolopes: I am not sure how this could be implemented in terms of accessibility, but maybe, if it is omitted, ATs could simply state that the field is required.

If it is included, however, and it has text attached to it, it could substitute the default AT required indication.

research/src/pages/radio-button/radio-button.proposal.mdx
```diff
@@ -0,0 +1,185 @@
+---
+menu: Proposals
+name: Radio Button (Editor's Draft)
```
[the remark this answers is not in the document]
leolopes: I didn't change anything at first because the rest of the discussion did not advance. But as soon as I get a clear requirement from the reviewers, I will change it.

research/src/pages/radio-button/radio-button.proposal.mdx
leolopes: Does the code you used as an example mean that, by placing `required` in one of the options, the **group** is required?

scottaohara: yes. the implicit grouping becomes required and you cannot submit a form unless one of the radios is checked

research/src/pages/radio-button/radio-button.proposal.mdx
scottaohara: the required attribute / aria-required=true attribute is what is important to expose to AT for the required nature of the grouping.  The indicator would be to visually inform what is then conveyed programmatically. 

>If it is included, however, and it has text attached to it, it could substitute the default AT required indication.

I'm not exactly sure what you're trying to say here.

research/src/pages/radio-button/radio-button.proposal.mdx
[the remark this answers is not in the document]
scottaohara: Understood, as I did see the mention of the attribute `group`.  My mention of `name`'s current behavior was in regards to the fact that this thread mentioned uncertainty in how to associated radios if not children of a `radiogroup`.  Being that `name` makes this possible now seemed appropriate to bring up since there seemed uncertainty.

Per this, I've also filed an issue against ARIA https://github.com/w3c/aria/issues/1721 to allow for a feature with ARIA to make element associations without the need for explicit nesting within a parent group.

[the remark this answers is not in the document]
leolopes: Oh, I might not have been clear, but I understood `name`'s current behavior :)

> Per this, I've also filed an issue against ARIA https://github.com/w3c/aria/issues/1721 to allow for a feature with ARIA to make element associations without the need for explicit nesting within a parent group.

This would be exactly what we need to make it ARIA compliant, thanks!

research/src/pages/radio-button/radio-button.proposal.mdx
leolopes: > If it is included, however, and it has text attached to it, it could substitute the default AT required indication.

Suppose an AT software might stop at a required field and say "Required" as it reads the `required` attribute. This would be what I expect to be default behavior.

Suppose then we have a `required-indicator` element on DOM, and it has a custom text saying "This field is required due to XYZ". Then, instead of simply reading "Required", it could read both or maybe only the text inside the `required-indicator`.

Just an idea.

research/src/pages/radio-button/radio-button.proposal.mdx
leolopes: Good to know, I was not aware of it.

gregwhitworth:
@leolopes friendly ping on this, do you intend on taking this further?

leolopes:
> @leolopes friendly ping on this, do you intend on taking this further?

Hi @gregwhitworth , I do, I just have been overwhelmed by work and I thought the proposal didn't get much traction so I didn't prioritize this. But I'll try to reserve some time this week and fix conflicts and make some of the requested changes.

research/src/pages/radio-button/radio-button.proposal.mdx
leolopes: I will consider this to be resolved since I still abide by my proposal and I didn't hear back afterwards.

research/src/pages/radio-button/radio-button.proposal.mdx
leolopes: I'll consider it resolved

leolopes:
@gregwhitworth , I was trying to resolve the conflicts and now I noticed that the website was rewriten wit Astro in place of Gatsby. I'll have to learn it and make a lot of adjustments so my content will be able to run on the new structure.

leolopes:
There seems to be an error on a file that was not modified by me. I'm not sure how it can be, because I pulled directly from `master`.

I can try to fix the file and push again later.

gregwhitworth:
@andrico1234 can you take a look at this given the error that @leolopes is seeing?

leolopes:
@gregwhitworth and @andrico1234 I found out what was happening: Husky was messing up with a commented section on the popover.research.explainer.mdx file. I had to bypass the linting proccess on my last commit so I wouldn't run into a syntax error again.

I think either the linter should be reconfigured or the commented section should be removed.

Anyway, the PR is ready for review.

site/src/pages/components/radio.research.mdx
```diff
@@ -10,26 +11,26 @@ import Anatomy from '../../components/anatomy'
 import Concepts from '../../components/concepts'
 import ComponentCoverage from '../../components/component-coverage'
 
-<ComponentCoverage component="Radio Button" />
+<ComponentCoverage component="Radio" />
 
 ## Introduction
 
-_(For a lengthier discussion, on which this page was based, please refer to the [original radio button issue](https://github.com/openui/open-ui/issues/251))_
+_(For a lengthier discussion, on which this page was based, please refer to the [original radio issue](https://github.com/WICG/open-ui/issues/251))_
```
gregwhitworth, line 18: I'd remove this or just make this a bit more conversational, but prior to doing this - please see my overall comment regarding the editor's draft and research.

site/src/pages/components/radio.research.mdx
```diff
@@ -10,26 +11,26 @@ import Anatomy from '../../components/anatomy'
...
 
-The research showed that different design systems treat radio buttons in one of two ways, i.e.:
+The research showed that different design systems treat radios in one of two ways, i.e.:
 
-- as a standalone component, that could be hosted on a “**radio button group**” component;
-- as an invariably non-independent part of a “**radio button group**” component (just like `<option>` tags are dependent on `<select>` tags).
+- as a standalone component, that could be hosted on a “**radio group**” component;
```
gregwhitworth, line 22: wouldn't this be _within_ rather that _on_ ?

site/src/pages/components/radio.research.mdx
```diff
@@ -10,26 +11,26 @@ import Anatomy from '../../components/anatomy'
...
+- as an invariably non-independent part of a “**radio group**” component (just like `<option>` tags are dependent on `<select>` tags).
 
-The two approaches seem very similar at first, but the former seems to indicate that single “radio buttons” could exist, to probably indicate a binary state (on or off). The latter, however, shifts the scope up to the “radio button group”, which, although nonexistent as an HTML element, is an important concept for the W3 specification:
+The two approaches seem very similar at first, but the former seems to indicate that single “radios” could exist, to probably indicate a binary state (on or off). The latter, however, shifts the scope up to the “radio group”, which, although nonexistent as an HTML element, is an important concept for the W3 specification:
```
gregwhitworth, line 25: let's remove _probably_ and be a bit more emphatic on your takeaways from the research. If you don't have a strong position then make it an open question.

site/src/pages/components/radio.research.mdx
```diff
@@ -38,35 +39,35 @@ After the research there are still [discussions to be made](#discussion), where
...
 
 ## Concepts
 
-<Concepts client:load component="Radio Button" />
+<Concepts client:load component="Radio" />
 
 ## Discussion
 
-At a point during the research, the scope of this study os shifted from the “radio button” to the “radio button group”, because there is no particular strong use case for a standalone “radio button”, and also because this use case is against the W3 Specification, the actual component to be considered is the “radio button group”, being composed of two or more radio buttons.
+At a point during the research, the scope of this study os shifted from the “radio” to the “radio group”, because there is no particular strong use case for a standalone “radio”, and also because this use case is against the W3 Specification, the actual component to be considered is the “radio group”, being composed of two or more radios.
```
gregwhitworth, line 50: remove _os_ as "the scope of this study shifted" is sufficient

> there is no strong usecase for a radio

I disagree with this stance as I've heard the same statement for checkbox but can make an argument for it

site/src/pages/components/radio.research.mdx
```diff
@@ -38,35 +39,35 @@ After the research there are still [discussions to be made](#discussion), where
...
 
 Also based on this fact and on the current state of the design systems, a more final anatomy was proposed.
```
gregwhitworth, line 52: I don't follow this sentence.

site/src/pages/components/radio.research.mdx
```diff
@@ -38,35 +39,35 @@ After the research there are still [discussions to be made](#discussion), where
...
 
 ### Proposed Anatomy - by Tag Hierarchy
```
gregwhitworth, line 54: Let's remove proposed anatomy and just make it "Anatomy" as every spec in Open UI is effectively a proposal and as such makes this title redundant

site/src/pages/components/radio.research.mdx
```diff
@@ -38,35 +39,35 @@ After the research there are still [discussions to be made](#discussion), where
...
 
-This section shows the proposed visual constituents for the _Radio Button Group_, which is here considered to be the actual component, while the radio buttons are a part of the group.
+This section shows the proposed visual constituents for the _Radio Group_, which is here considered to be the actual component, while the radios are a part of the group.
```
gregwhitworth, line 56: as with the title, remove _proposed_

site/src/pages/components/radio.research.mdx
```diff
@@ -38,35 +39,35 @@ After the research there are still [discussions to be made](#discussion), where
...
 
-- Radio Button Group (the actual component)
+- Radio Group (the actual component)
   - Group Label (functions as a label for the whole group)
   - Required Indicator
-  - Radio Button (at least two required)
+  - Radio (at least two required)
```
gregwhitworth, line 61: while I like this break down, we should have some text that explains why two are required.

site/src/pages/components/radio.research.mdx
```diff
@@ -38,35 +39,35 @@ After the research there are still [discussions to be made](#discussion), where
...
+  - Radio
     - and so on...
   - and so on...
 
 #### The problem
 
-The above anatomy is regular enough to be assembled in a way similar to (but more complex than) a `<select>`. Attributes set on the group would allow the developer to set the whole group as "disabled", "readonly", vertically or horizontally laid out, types of display (like buttons, radio buttons, combobox, etc), and attributes set on the radio buttons would set their specific states, as we do with `<option>` tags. It is important to understand that the function of radio buttons is almost the same as a `<select>` tag, and this is why the comparison makes sense.
+The above anatomy is regular enough to be assembled in a way similar to (but more complex than) a `<select>`. Attributes set on the group would allow the developer to set the whole group as "disabled", "readonly", vertically or horizontally laid out, types of display (like buttons, radios, combobox, etc), and attributes set on the radios would set their specific states, as we do with `<option>` tags. It is important to understand that the function of radios is almost the same as a `<select>` tag, and this is why the comparison makes sense.
```
gregwhitworth, line 70: I wouldn't bring `<select>` at this stage. Possibly consider a section at the end with something like, "Other solutions considered"

gregwhitworth (changes requested):
I haven't reviewed thoroughly the definitions of the items nor their anatomy's but there is a key change that I'd like to request.

I would love for your proposal to follow the shape of the [Popup API](https://open-ui.org/components/popover.research.explainer/) where it makes sense. Some key aspects I'd like:

* Background: Get into why there is a need for both Radio and Radio Group. you cover this generally but I'd love for the problem to be up front in the explainer whereas right now it's spread throughout the research document.
* Problem statement: You may briefly hit on this in the background but enumerate each of the problems that not having these controls currently present to end users. Again, you have this content just move it up here.
* Goals: This is where you'll enumerate the goals which aim to solve the problem statements
* Radio Element: For now let's remove the API section and lead with the anatomy and definitions. If necessary, don't be afraid to expound on key details of various parts
* Radio Group: Same as radio element
* Other solutions considered: 
     * Why not just use `<select>` or `<selectmenu>`?

Thank you again for doing this work; and while the above may seem daunting it's more a shuffle of the content that you have as we've realized as Open UI has matured that we should primarily author explainers which should try and read as a single document even though there may be supporting content; but someone shouldn't have to read that content to fully understand the proposal and the problem that it's solving.

leolopes:
> ...while the above may seem daunting it's more a shuffle of the content...
@gregwhitworth 

That's a bit of work but I can manage it 😉 . Just not sure when I'll get back to it, but not as long as it took before.
Thanks!

gregwhitworth:
This can be re-opened if someone wants to pick it up. It hasn't had progress for over a year now so closing it out.

leolopes:
@gregwhitworth I will eventually get back to it, maybe during vacation time.
I just didn't because I have been too busy and the required rework, though it doesn't change the essence of the proposal, is not trivial.

gregwhitworth:
@leolopes it's alright; you can always re-open this. Life happens and we appreciate any contributions that can be added.

PR #371, opened 2021-07-04
Feature: Add human-readable sitemap

andrico1234:
Simply replace the Gatsby sitemap plugin, with Ghost's advanced sitemap plugin.

[Docs here](https://www.gatsbyjs.com/plugins/gatsby-plugin-advanced-sitemap/?=sitemap)

A neat feature is that we can create individual sitemaps via custom queries, so we can split research and proposals into their own respective sitemaps.

I'll be happy to look into this if you folks feel this'll provide some value! If not, it's something to consider for the future

Related: https://github.com/openui/open-ui/issues/314

https://user-images.githubusercontent.com/17087167/124399572-16690d80-dd14-11eb-800f-9cf2ae698136.mov


andrico1234:
Hi @melanierichards, I'm tagging you as I don't have permissions to add reviewers

andrico1234:
It's failing as a result of node being out of date. 

Open UI's on 12.10.0, and the sitemap plugin requires versions >=12.19.

With the latest maintenance LTS being node 12.22.2, and the current LTS is 14.17.x. Would anyone be against updating our node version to the latest v12?

Have there been issues with upgrading in the past?

melanierichards:
@andrico1234 looks good to me! I think scoping to particular queries would certainly be interesting and useful. If there's a way to filter out the "Images" column in the results that might be good (since it seems extraneous), but non-blocking.

I'll defer to @gregwhitworth regarding Node versions.

andrico1234:
Great stuff! In that case, I can push a change with a later node version (12 or 14), and hopefully the OUI previews don't explode 😄

gregwhitworth:
@andrico1234 In running this locally I get the following error - can you update the necessary configs in order for this all to run as expected in this one PR:

`warn Plugin gatsby-plugin-advanced-sitemap is not compatible with your gatsby version 2.32.12 - It requires gatsby@^3.0.0`

gregwhitworth:
Friendly ping @andrico1234

andrico1234:
That's strange, i'm not having the problem, and it the version of Gatsby we have specified in the package is `2.20.14`. 

Have you got the Gatsby cli installed? if so, what version is it? Mine's 3.x.x, and I'm not having the build issue.

andrico1234:
The simplest solution might be to upgrade the repo's version of Gatsby to 3.0. If we're cool with that, I can do a regression test across the site, to check for breakages.

gregwhitworth (changes requested):
Can you clean up merge conflicts with main (ironically they're probably from you :) )

andrico1234:
@gregwhitworth the merge conflicts have been resolved, and is ready for a re-review

andrico1234:
Note: I believe the huge file change in the `yarn.lock` came from me removing the `gatsby-plugin-favicon` module, I must've missed it in the dependency cleanup PR. It relied on v2 of `gatsby`, and from the looks of it, no other packages do, which would cause a huge change in the yarn.lock. 

Thought I'd share, since it made me double-take at first.

andrico1234:
@gregwhitworth friendly bump

gregwhitworth (approved):

PR #384, opened 2021-08-11
Chore: bump dependencies

andrico1234:
My main motivation to do this was to [unblock the sitemap](https://github.com/openui/open-ui/pull/371), which requires Gatsby v3+

Updates packages to their most recent versions. I omitted some:

# Todo
- [x] regression test across the site

# Notes
- Bad news, gatsby-plugin-favicon [doesn't work](favicon not good: https://github.com/Creatiwity/gatsby-plugin-favicon/issues/71)
- Good news, gatsby-plugin-manifest, which was previously commented out due to a bug, works fine with Gatsby 3, and has favicon functionality
image
- `sharp` was removed because it wasn't used anywhere. The installation notes for `gatsby-plugin-sharp` don't mention needing to install `sharp` at all.
- `theme-ui` and `gatsby-plugin-theme-ui` were both removed as neither were used in the repo.
- husky wasn't updated because there were some major changes to the api in v6, and I wanted to keep the change surface small
- prettier wasn't updated to keep the change surface small
- the two remark plugins weren't updated to v3 as they removed the ability to import via `require`, gatsby doesn't allow import [without a bit of faff](https://github.com/gatsbyjs/gatsby/issues/23705)
- i didn't update emotion because the repo is no longer available, and so I don't know what was changed in the most recent major version 🤷🏽‍♀️
- I updated the version of Node to 14. Nothing broke
- All images were broken until I made the change in `Image.js`. I have no idea why...


research/src/components/image.js
```diff
@@ -3,7 +3,7 @@ import React from 'react'
 const Image = ({ src, alt = src, style, ...rest }) => {
   let imageData
   try {
-    imageData = require(`../images/${src}`)
+    imageData = require(`../images/${src}`).default
```
andrico1234, line 6: 🤷🏽

gregwhitworth (approved):

PR #395, opened 2021-08-30
Behavior design updates

jxlz:
Design: Adding accessibility red line information for selects.
Behavior: Adding state diagrams and example walk throughs to illustrate state changes and events in context


jxlz:
I'd like to bring some template change proposals to the group for discussion using selects as an example.

research/src/pages/select/select.proposal.mdx
```diff
@@ -161,6 +161,18 @@ children:
...
+When specifying a select in design comps, the following accessibility red lines need to be called out: 
+
+_Example Table_
+
+| Red line | Detail | Notes |
+| ---------- | --------------- | ------- |
+| name    | label              | This is usually the visible label of the select  |
+| role    | combobox           |   |
+| value   | "Make a selection" | It is a better experience to not have an empty default value |
+
```
chrisdholt, line 175: It could just be me, but this seems overly prescriptive of the design process and certain points of it. For example, this seems specifically targeted at teams *building* a custom select. If I work in an organization with an existing component library, would I specify the following within the design comps or would I say, "use `foo-select`"? That seems like a matter of opinion to me and not one that necessarily matters to the goals of Open-UI. Perhaps I'm missing the intent here, but this seems like something that should be decided within design and engineering teams and not necessarily here.

research/src/pages/select/select.proposal.mdx
```diff
@@ -183,6 +195,26 @@ and is therefore considered to be invalid.
...
+![States of a select](../../images/statediagram-select.png)
+
+A `change`event fires on AltUpOREnter, as well as when tabbing off the select when the listbox is open.
+
+### Walk Through Scenario Example
+_Example Table_
+
+| State | Action | Resulting red line change | State change / Event fired
+| ---------- | --------------- | ------- | --------- |
+| collapsed    | Alt+Down      |  | expanded |
```
chrisdholt, line 213: I'm seeing references to `ALT + (up/down)` here, which seems like a Windows specific implementation issue that doesn't map to the documented interaction model for Combobox: https://w3c.github.io/aria-practices/#combobox

chrisdholt:
Would an issue be a better place to discuss proposed additions to the `spec` or `proposal` process?

research/src/pages/select/select.proposal.mdx
```diff
@@ -183,6 +195,26 @@ and is therefore considered to be invalid.
 
 ### Interaction &amp; transition of states
 
+[Select Usage State Diagram](https://xstate.js.org/viz/?gist=7965135d58118c828140f6e830ca93a1)
+
+![Usage of select within other UI](../../images/statediagram-usage-select.png)
+
+[Select State Diagram](https://xstate.js.org/viz/?gist=24e03b6b21dcb0d1eb7471d6ff9043f3)
```
chrisdholt, line 202: @dandclark can comment further, but if we are mapping to `combobox` then we should likely follow the combobox interaction model rather than a single instance. I'm not following how this example represents the current state of select in Open UI.

research/src/pages/select/select.proposal.mdx
jxlz: Would a team just say "use foo-select", or would they say "use foo-select with name <foo-select-name> and default value of <foo-select-default-value>"? This is getting at what properties a dev would need in order to implement it properly within a UI.

research/src/pages/select/select.proposal.mdx
jxlz: Good point, I'll remove it since we're talking about web contexts here.

research/src/pages/select/select.proposal.mdx
chrisdholt: Again, this depends on the team, goal, etc. If we have an existing component library, I would spec, “use foo-select”. If I’m building a component library and we want to build to Open UI’s specification, I would point to the Open UI spec perhaps.

As it pertains to the goals of Open UI, I’m not sure I understand what issue this solves. From my perspective, I’m not sure prescribing how a design team should spec falls under those goals. I think there are several valid approaches, all which depend on a teams goals, size, philosophy, etc. If we were to include a “how to spec in design”, I think we’ll end up with guidance being, “it depends on your requirements.”

research/src/pages/select/select.proposal.mdx
jxlz: I can add flows for all of the keyboard interactions so it feels complete.

research/src/pages/select/select.proposal.mdx
```diff
@@ -183,6 +195,26 @@ and is therefore considered to be invalid.
...
+
+![States of a select](../../images/statediagram-select.png)
```
gregwhitworth, line 204: These don't seem to be working

research/src/pages/select/select.proposal.mdx
gregwhitworth: @jxlz W3C/WHATWG specifications do have optional capability through Non-normative text or the use of prose that use terms like, "May" or "Should". Each of those of course convey something different than "MUST" which represents something that IS NOT optional.

gregwhitworth (changes requested):
I really find this section interesting and I think a design of the outlined specifications in a base design system that will mirror the base styles will be invaluable. But I think I'd prefer to lean towards the default styles with supported imagery or even an agreed upon solution (eg: Figma project, etc). This ties in with the general discussion around web component creation, etc.

andrico1234:
@jxlz , friendly PR bump

gregwhitworth:
Given the lack of activity on this I'm closing this PR; it can always be re-opened if there is a desire to move it forward.

PR #431, opened 2021-12-10
feat: SpinButton research

spmonahan:
Adds SpinButton research. There is little consensus on the name for this component so I went with "SpinButton" as the ARIA role name for this type of control.


gregwhitworth:
This has been sitting for over a year so closing this out, it can always be re-opened if you want to take it further.

PR #441, opened 2022-01-14
select: Add accessibility part information

scottaohara:
initial drat of adding accessibility information to the `select` draft spec.

cc @dandclark

### DIFF
--- a/research/src/pages/select/select.proposal.mdx
+++ b/research/src/pages/select/select.proposal.mdx
@@ -117,6 +117,30 @@ children:
   Should this apply to the entire anatomy or solely the &lt;option&gt; and &lt;optgroup&gt; element
 </p>
 
+### Accessibility Implementation <a class="link" id="assistive-tech"></a>
+
+The `<select>` element represents a "combobox" and would be mapped as such across the different accessibility APIs per the [`combobox` role](https://www.w3.org/TR/core-aam-1.1/#role-map-combobox) and the [HTML AAM spec](https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings). The `<listbox>` popup would be mapped per the [`listbox` role](https://www.w3.org/TR/core-aam-1.2/) 
+
+The different parts of the `<select>` element and its popup `<listbox>` must expose the following default accessibility roles, states and properties: 
+
+
+| Part       | Role                   | State/Property Attribute     | Usage                                                                               |
+| -----------| -----------------------| -----------------------------|------------------------------------------------------------------------------------ |
+| select     |                        |                              | Serves as a wrapper to its accessible parts                                         |
+| button     | `combobox`             |                              | Exposes the necessary `combobox` role                                                |
+|            |                        | `aria-controls=IDREF`        | Creates association with the `listbox`                                              |
+|            |                        | `aria-expanded=true / false` | Identifies if the `listbox` popup is displayed. Reflects the `open` attribute       |
+|            |                        | `aria-haspopup=listbox`      | Identifies that the control will invoke a listbox popup. *Note* the [value of this attribute may change](https://github.com/w3c/aria/issues/1024#issuecomment-1011303969).      |
+|            |                        | `aria-controls=IDREF`        | Creates association with the `listbox`                                              |
+| listbox    | `listbox`              |                              | Exposes the necessary `listbox` role                                              |
+| option     | `option`               |                              | Exposes the necessary `option` role                                             |
+| optgroup   | `group`                |                              | xposes the necessary `group` role                                              |
+  
+**Note / open question:** the `<button>` part will need to use `aria-activedescendant` if keyboard focus is to remain on that element when the `<listbox>` popup is invoked. If keyboard focus is actually set to one of the `<option>` elements within the `<listbox>`, then this may not be necessary and focus management can be handled differently.
+
+**Question:** should all accessibile states and properties be covered here?  e.g., `disabled`, `selected`, etc.?  If `<select>` will retain multiselect functionality, it seems reasonable. If not, then it may be beyond what's necessary.
+
+
 ## Events <a class="link" href="#events" id="events"></a>
 
 ### select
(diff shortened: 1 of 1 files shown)


research/src/pages/select/select.proposal.mdx
```diff
@@ -117,6 +117,30 @@ children:
...
+|            |                        | `aria-controls=IDREF`        | Creates association with the `listbox`                                              |
+|            |                        | `aria-expanded=true / false` | Identifies if the `listbox` popup is displayed. Reflects the `open` attribute       |
+|            |                        | `aria-haspopup=listbox`      | Identifies that the control will invoke a listbox popup. *Note* the [value of this attribute may change](https://github.com/w3c/aria/issues/1024#issuecomment-1011303969).      |
+|            |                        | `aria-controls=IDREF`        | Creates association with the `listbox`                                              |
+| listbox    | `listbox`              |                              | Exposes the necessary `listbox` role                                              |
+| option     | `option`               |                              | Exposes the necessary `option` role                                             |
+| optgroup   | `group`                |                              | xposes the necessary `group` role                                              |
```
dandclark, line 137:
```suggestion
| optgroup   | `group`                |                              | Exposes the necessary `group` role                                              |
```

research/src/pages/select/select.proposal.mdx
```diff
@@ -117,6 +117,30 @@ children:
...
+  
+**Note / open question:** the `<button>` part will need to use `aria-activedescendant` if keyboard focus is to remain on that element when the `<listbox>` popup is invoked. If keyboard focus is actually set to one of the `<option>` elements within the `<listbox>`, then this may not be necessary and focus management can be handled differently.
+
+**Question:** should all accessibile states and properties be covered here?  e.g., `disabled`, `selected`, etc.?  If `<select>` will retain multiselect functionality, it seems reasonable. If not, then it may be beyond what's necessary.
```
dandclark, line 141: I think that we'll eventually want this to be comprehensive, it's reasonable to cover `disabled` and `selected` behavior for single-select. Doesn't need to be part of this change though.

research/src/pages/select/select.proposal.mdx
```diff
@@ -117,6 +117,30 @@ children:
   Should this apply to the entire anatomy or solely the &lt;option&gt; and &lt;optgroup&gt; element
 </p>
 
+### Accessibility Implementation <a class="link" id="assistive-tech"></a>
+
+The `<select>` element represents a "combobox" and would be mapped as such across the different accessibility APIs per the [`combobox` role](https://www.w3.org/TR/core-aam-1.1/#role-map-combobox) as defined by CORE AAM and the [HTML AAM spec](https://www.w3.org/TR/html-aam-1.0/#html-element-role-mappings). The `<listbox>` popup would be mapped per the [`listbox` role](https://www.w3.org/TR/core-aam-1.2/) as defined by CORE AAM. **Note:** awaiting clarity on which `listbox` mapping to use, as one may be specific to the deprecated ARIA 1.1 `combobox` pattern.
```
dandclark, line 122: I'm wondering if we should also try to fit a link to https://w3c.github.io/aria/#combobox in here somewhere.
Clicking the `combobox` link in the left column at https://www.w3.org/TR/core-aam-1.1/#role-map-combobox leads to https://www.w3.org/TR/wai-aria-1.1/#combobox, which suggests that the combobox role is strictly used with textboxes, as opposed to the newer guidance in 1.2 that it can also be used with a button.

dandclark (approved):
Thanks for adding this!
LGTM with a couple small questions/suggestions.

research/src/pages/select/select.proposal.mdx
scottaohara: I don't mind adding it.  I just wanted to make sure it was desired before I did more than what was needed.  thanks

research/src/pages/select/select.proposal.mdx
scottaohara: yeh.  can definitely add those.  good call

scottaohara: links have been updated.  we should be adhering to the latest definition of `combobox`.  The ARIA 1.1 definition is not a viable path forward as it is deprecated.

research/src/pages/select/select.proposal.mdx
scottaohara: added in content for `disabled`, `selected` and `required`.   Added in expected behaviors for when selecting an option / dismissing a popup listbox.

Question still remains on the way focus will be treated, but the outlined behaviors would apply in either case (they might just be less relevant if the invoking combobox/button retains focus and `aria-activedescendant` is used to indicate the current selected option).

Expectations for `multiple` are left out for now until resolution of #447

scottaohara:
@dandclark have updated.  please let me know if there's anything that needs further clarification, or if you notice i missed something.

In reviewing the explainer and demos from #448, there are now further considerations and questions about how to spec the accessibility mappings.

For instance, per the "combobox" example in the explainer - this is a use case I've often cited as to where the invoked popup would need to be exposed as a non-modal dialog.

The "Word's color select", and instances where the 'button' can be replaced by a new custom button and then static text, being other examples that needs further consideration.  The color select, for instance, is more like a 'split button'  (related ARIA issue to consider adding a new role for this control type https://github.com/w3c/aria/issues/1670), where the 'button' has its own behavior and the element used to invoke the popup is a separate menu button, invoking a popup with `menu` > `menuitem` roles.

Point is that if the element is going to allow for customization beyond the default experience/mappings, then mappings would need to change as well.  Not saying that shouldn't be allowed by any means. Just that this adds complexity and changes a11y expectations for the control that need to be further defined.

research/src/pages/select/select.proposal.mdx
```diff
@@ -117,6 +117,44 @@ children:
...
+|            |                        | `disabled`                   | The element is exposed as disabled and it cannot be selected by a user, or can a user use arrow keys to navigate to this option. Expose the `option` as `aria-disabled=true`.|
+|            |                        | `selected`                   | The element is exposed as being in the selected state for a single-select `listbox`. Exposed as `aria-selected=true`.  Other `option` elements within a single-select `listbox` must be exposed as `aria-selected=false`.|
+| optgroup   | `group`                |                              | Exposes the necessary `group` role.                                              |
+|            |                        | `disabled`                   | Identifies all `option` elements within the `optgroup` as being in the disabled state. All `option` elements must be treated as if they had been provided the `disabled` attribute themselves.  |
+
+**TODO:** add in accessibility considerations for `multiple` depending on resolution of [issue 447](https://github.com/openui/open-ui/issues/447).
+
+### Dismissal of the popup contnet
```
hidde, line 146: typo: contnet

scottaohara:
@gregwhitworth @dandclark, per some other conversations I've had with Dan, this may end up needing more work.  If `selectmenu` has additional features/functionality beyond a more stylable `select` element replacement, then this accessibility information wouldn't be sufficient for those additional use cases.

gregwhitworth:
Thanks @scottaohara 

@dandclark I've converted this to a draft then to start as a baseline for an updated variant.

gregwhitworth:
@josepharhar are @scottaohara feedback on a11y covered in your updated variant of the explainer?

scottaohara:
@gregwhitworth the a11y of this is still a bit nebulous on a few points - see #822 and #821 for some instances why.

i think this PR should be closed and I can take a crack at some updates / additions to the explainer. i'll probably do a few PRs since some stuff should be relatively straight forward, while others might need discussion.

PR #448, opened 2022-01-26
`<selectmenu>` user documentation

### LINKED ISSUE #438: Adding docs for controls to the site
A new native control is being implemented in Chromium based on the open-ui `select` proposal. Its name is `selectmenu` and it can already be used in Chromium-based browsers (not sure of the version yet, better stick to Canary for now if you want to try), with the `Experimental Web Platform features` flag enabled.

I'm filing this issue to propose writing some technical documentation about `selectmenu` (and others when they become available) in order for early testers to try it out and provide feedback.

While it's way too soon for the control to be production ready and have docs on MDN and so on, it is far enough along that it'd be good for a few early testers to build demos with it and report limitations with the design, and bugs with the implementation.
These early testers can only do so effectively if they have access to documentation about the control (its anatomy, how to use `::part()`, how to extend the markup, etc.).

This documentation needs to be separate from the spec I think, and it could be listed as a new table of content node under "Documentation" on the site.

I drafted some docs on a branch here: https://github.com/captainbrosset/open-ui/blob/selectmenu-docs/research/src/pages/prototypes/selectmenu.mdx

Pinging @dandclark @ststimac and @gregwhitworth who I've discussed this with.

captainbrosset:
This adds a bit of user documentation for the new `<selectmenu>` component, based on the select open-ui proposal.
The aim is for this to be used by early testers to report feedback about the component itself, and the design of the proposal.

This fixes #438.


captainbrosset:
@gregwhitworth since we'll be talking about the `<selectmenu>` control at the next telecon, I'd love it if this doc could be merged so it's available for people to try out.
How about we merge it without it being linked from anywhere yet? That could give us time to figure out where documentation should leave on the site, but we would still be able to give people a URL to the docs at the meeting.

gregwhitworth:
@captainbrosset @dandclark done, as noted in our side chats I think I'd like a more formal approach to this. Given what this is I think it ending up on MDN makes the most sense to me although I wouldn't mind a generic variation of this possibly even using the same examples but being used to explain the model we're leaning on

PR #452, opened 2022-02-02
feat: add toolbar proposal

chpalac:
Add `Toolbar` proposal


chrisdholt:
I see Fabric and Stardust both represented here, but I'm fairly sure that those are both represented now as Fluent UI. If I recall, this came up previously and we had looked at pointing at a single implementation (Fluent UI) rather than Fabric (no longer existing) and Stardust (part of Fluent UI merging as well...).

research/src/pages/toolbar.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+  </div>
+  <span />
+  <button />
+</div>
+```
+
+#### Accessibility <a class="link" href="#accessibility" id="accessibility"></a>
+
+- The element that serves as the toolbar container has role toolbar.
+- If the toolbar has a visible label, it is referenced by aria-labelledby on the toolbar element. Otherwise, the toolbar element has a label provided by aria-label.
```
scottaohara, line 50: This is how the ARIA practices is outlining how to use ARIA to provide a toolbar an accessible name.  Could not this be more open to possibilities?  for instance, rather than using `aria-labelledby`, if a `fieldset` were used as the base container with a `role=toolbar`, then it could leverage the `legend` element to specify a name for the toolbar.

the end goal of this could also be that HTML provides mechanisms to name certain elements.  This topic has come up with HTML for use of naming landmark elements without use of ARIA, or using the title attribute.  a toolbar would be another instance of where'd we need HTML to support this naming feature request.

research/src/pages/toolbar.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+## Use Cases
+
+A `Toolbar` groups visually a set of controls semantically algined and it can also be ane effective way of reducing the number of tab stops
+
+## Prior Art/Examples <a href="#prior-art" id="prior-art"></a>
+
+- [Stardust](https://fluentsite.z22.web.core.windows.net/0.59.0/components/toolbar/definition)
+- [Reakit](https://reakit.io/docs/toolbar/)
+- [Radix UI](https://www.radix-ui.com/docs/primitives/components/toolbar)
```
chrisdholt, line 22: We have other prior art examples given, should we link out to those as well? Would we consider having more generic examples as well such as the toolbar from RoosterJS's rich text editor?

research/src/pages/toolbar.proposal.mdx
```diff
@@ -0,0 +1,72 @@
...
+```html
+<div>
+  <button />
+  <div>
+    <button />
+    <button />
+  </div>
+  <span />
+  <button />
+</div>
```
chrisdholt, line 44: I'd probably prefer to understand the semantic representation of the structure here. Are the buttons here intended to be examples of slotted content? Similar to @scottaohara's question, what's the reason for the structure here? What are the named parts and why? In FAST, we identified a common pattern of having "start" and "end" content relating to layout, etc...that may not make sense here, but I think at the very least if there is additional DOM, understanding its intent is important.

scottaohara: agreed.  there a disconnect between what this snippet is meant to represent, particularly in comparison to what the accessibility requirements are saying are necessary, as well.

gregwhitworth:
Thanks for submitting this @chpalac - per the working mode can you open an issue about your interest in investigating this control/component: https://open-ui.org/working-mode

A key thing to denote within that issue is what your goal is for this component. Given that toolbar appears to generally overlap with other aspects I'd really like discourse on Open UI taking on this variation and you allowing your research to validate/invalidate the need for this component. Let me know if you have any questions

gregwhitworth:
@chpalac hey any update on this?

research/src/pages/toolbar.proposal.mdx
christopherallanperry: For a while now, I've wondered whether there was a way of adding `<label for="something">` to elements other than `<input>`. Scott's comment leads me to think mine wasn't an isolated thought.

scottaohara: yeh, there's an issue on the HTML spec to do this... somewhere.  i'd have to dig.  but it's something that came up again recently per the recent request to add the `<search>` element.

gregwhitworth:
This has been sitting for over a year so closing this out, it can always be re-opened if you want to take it further.

PR #474, opened 2022-02-28
Chore: Automatically mark old issues as 'stale'

andrico1234:
Adds a GitHub action that runs once a day which adds a 'stale' label to issues with no activity in 6 months.

### DIFF
--- a/actions.yml
+++ b/actions.yml
@@ -1,0 +1,18 @@
+name: Add labels to stale issues
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4
+        with:
+          days-before-issue-stale: 180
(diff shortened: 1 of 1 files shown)


andrico1234:
Tested on a dummy repo:
image

gregwhitworth (commented):
Thanks for this. I haven't used Github workflows - will this close the PR/Issue automatically as well or just add the message?

andrico1234:
It can do both,

We can manage when to mark an issue/pr as stale, and when to close it.

I've changed the default configuration so the action doesn't close any PRs or issues. And to only mark issues as stale if there has been no activity for more 6 months.


>days-before-issue-stale: 180
days-before-pr-stale: -1
days-before-issue-close: -1
days-before-pr-close: -1


According to the [docs](https://github.com/actions/stale):
> If set to a negative number like -1, no issues or pull requests will be marked as stale automatically.
In that case, you can still add the stale label manually to mark as stale.

Issue #480, opened 2022-03-11
Bring focusgroup work under the OpenUI umbrella


travisleithead:
Hello Open UI folks! 

I'd like to move the [focusgroup explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/Focusgroup/explainer.md) into OpenUI and continue iterating on it here. I think this community is the right place to continue to build consensus and get implementer feedback on it's "growing up" journey into HTML and beyond.

Note: Microsoft Edge engineers are currently working on early prototypes of this feature.

Thanks for your consideration!

chrisdholt:
I’m for it @travisleithead! (Biased…)

gregwhitworth:
Congrats @travisleithead

PR #516, opened 2022-04-08
Copy focusgroup explainer into OpenUI CG

travisleithead:
Per telecon resolution in https://github.com/openui/open-ui/issues/480, this PR copies the focusgroup explainer from MSEdgeExplainers to OpenUI (with minor tweaks to align with new folder structure) in order to continue collaboration in the right public incubation venue to further develop the proposal.


chrisdholt (approved):
Am I allowed to approve this? 😄 

Looks like a copy to me 👍

andrico1234:
Hey @travisleithead , looks like Gatsby ran into an error while building your changes. Let me know if you need a hand, happy to take a look into this if you're not familiar with the framework 😁

andrico1234:
@travisleithead , i found the problem and it was due to this line:

```
In the following example, the <deep-child> element opts-in to the focusgroup defined on "ancestor".
```

I've found that `mdx` parser struggles with HTML elements that are never closed. 

In the instance above, the renderer must've tried to parse <deep-child> as a regular HTML element. THe build then broke when because there was no corresponding closing tag. 

I wrapped the element in backticks, after which the `mdx` parser treats the value as a string. I hit my head on this issue recently as well, so I hope you don't mind me going ahead and applying the fix 😄 

I replaced the line copied above with the following and the build now passes! 

```
In the following example, the `<deep-child>` element opts-in to the focusgroup defined on "ancestor".
```

andrico1234:
I just viewed the preview server and it all looks great! Approved and merged

andrico1234 (approved):

PR #518, opened 2022-04-14
Turn <popup> into <div popup>

captainbrosset:
The popup implementation in chromium is changing as a result of bikeshedding discussions.

The `<popup>` element is becoming an attribute instead.

Patches have also started to land in Chromium to make `<selectmenu>` use this new implementation.

So this change updates our usage docs on the open-ui site.

### DIFF
--- a/research/src/pages/prototypes/selectmenu.mdx
+++ b/research/src/pages/prototypes/selectmenu.mdx
@@ -167,7 +167,7 @@ You can replace the default listbox part in a similar fashion:
 </selectmenu>
 ```
 
-The `<div popup=popup>` used here is the element proposed at [Popup (Editor's Draft)](http://localhost:8000/components/popup). The element with `behavior="listbox"` is required to be a `<div popup=popup>`. Applying `behavior="listbox"` tells the browser to apply select menu listbox behavior to that element: it will be opened when the `<selectmenu>`'s button is clicked, and the user can select `<option>`s inside it with mouse, arrow keys, and touch.
+The `popup` attribute used in this example's `<div popup=popup>` is from the proposal at [Popup API (Explainer)](https://open-ui.org/components/popup.research.explainer). The element with `behavior="listbox"` is required to be a `<div popup=popup>`. Applying `behavior="listbox"` tells the browser to apply select menu listbox behavior to that element: it will be opened when the `<selectmenu>`'s button is clicked, and the user can select `<option>`s inside it with mouse, arrow keys, and touch.
 
 The above code snippet results in the following style:
 


captainbrosset:
@mfreed7 @dandclark can you take a look at this when you get a chance?

research/src/pages/prototypes/selectmenu.mdx
```diff
@@ -156,18 +156,18 @@ You can replace the default listbox part in a similar fashion:
...
       <option>Option 4</option>
       <option>Option 5</option>
-    </popup>
+    </div>
   </div>
 </selectmenu>
 ```
 
-The `<popup>` used here is the element proposed at [Popup (Editor's Draft)](http://localhost:8000/components/popup). The element with `behavior="listbox"` is required to be a `<popup>`. Applying `behavior="listbox"` tells the browser to apply select menu listbox behavior to that element: it will be opened when the `<selectmenu>`'s button is clicked, and the user can select `<option>`s inside it with mouse, arrow keys, and touch.
+The `<div popup>` used here is the element proposed at [Popup (Editor's Draft)](http://localhost:8000/components/popup). The element with `behavior="listbox"` is required to be a `<div popup>`. Applying `behavior="listbox"` tells the browser to apply select menu listbox behavior to that element: it will be opened when the `<selectmenu>`'s button is clicked, and the user can select `<option>`s inside it with mouse, arrow keys, and touch.
```
andrico1234, line 170: should this be `<div popup=popup>` to match the examples used in the popup explainer?

image

andrico1234:
Looks good to me, but left a single comment! It'd be great to get a quick once over from one of the `popup` owners. Happy to approve and merge when given the thumbs up 😄

research/src/pages/prototypes/selectmenu.mdx
captainbrosset: Thank you for catching this! I had missed that spot.

research/src/pages/prototypes/selectmenu.mdx
```diff
@@ -156,18 +156,18 @@ You can replace the default listbox part in a similar fashion:
...
       <option>Option 4</option>
       <option>Option 5</option>
-    </popup>
+    </div>
   </div>
 </selectmenu>
 ```
 
-The `<popup>` used here is the element proposed at [Popup (Editor's Draft)](http://localhost:8000/components/popup). The element with `behavior="listbox"` is required to be a `<popup>`. Applying `behavior="listbox"` tells the browser to apply select menu listbox behavior to that element: it will be opened when the `<selectmenu>`'s button is clicked, and the user can select `<option>`s inside it with mouse, arrow keys, and touch.
+The `<div popup=popup>` used here is the element proposed at [Popup (Editor's Draft)](http://localhost:8000/components/popup). The element with `behavior="listbox"` is required to be a `<div popup=popup>`. Applying `behavior="listbox"` tells the browser to apply select menu listbox behavior to that element: it will be opened when the `<selectmenu>`'s button is clicked, and the user can select `<option>`s inside it with mouse, arrow keys, and touch.
```
mfreed7, line 170: I think the text is ok as-is, but it raises a good question: are we going to require that the element with `behavior=listbox` is **exactly** a `<div popup=popup...` or maybe any element that supports `popup=popup`?

@dandclark WDYT? Should we open an issue for this question?

mfreed7 (approved):
LGTM! Thanks for making these changes.

research/src/pages/prototypes/selectmenu.mdx
dandclark: Hmm, if we have a dedicated `<listbox>` element it might make sense to limit it to only `<listbox>`.
Otherwise, I think it should probably be `<div>` or `<span>` only. I can't think of any semantically-meaningful elements that would fit.

dandclark, line 170:
```suggestion
The `popup` attribute used in this example's `<div popup=popup>` is from the proposal at [Popup API (Explainer)](https://open-ui.org/components/popup.research.explainer). The element with `behavior="listbox"` is required to be a `<div popup=popup>`. Applying `behavior="listbox"` tells the browser to apply select menu listbox behavior to that element: it will be opened when the `<selectmenu>`'s button is clicked, and the user can select `<option>`s inside it with mouse, arrow keys, and touch.
```

dandclark (approved):
Thanks for updating this!

captainbrosset:
@andrico1234 looks like we have the right reviews in! Thanks all for your help.

andrico1234 (approved):

PR #527, opened 2022-05-09
Update Popup explainer to incorporate all recent changes

mfreed7:
This PR also revamps the explainer significantly, changing the language from "proposal" to "explainer". Many details are also added, such as what "light dismiss" really means, the "nearest open ancestral popup" algorithm, and many more details. I also added links back to many of the recent resolutions where some of these decisions made.

I'm sure there are typos and issues here, but I did a first editing pass. Comments appreciated.

### DIFF
--- a/research/src/pages/popup/popup.research.explainer.mdx
+++ b/research/src/pages/popup/popup.research.explainer.mdx
@@ -48,7 +48,7 @@ pathToResearch: /components/popup.research
 A very common UI pattern on the Web, for which there is no native API, is "popup UI" or "popups". Popups are a general class of UI that have three common behaviors:
  1. Popups always appear **on top of other page content**.
  2. Popups are **ephemeral**. When the user "moves on" to another part of the page (e.g. by clicking elsewhere, or hitting ESC), the popup closes.
- 3. Popups are **"one at a time"** - opening one popup closes others.
+ 3. Popups (of a particular type) are generally **"one at a time"** - opening one popup closes others.
 
 This document proposes a set of APIs to make this type of UI easy to build.
 
@@ -484,7 +484,7 @@ This section contains several HTML examples, showing how various UI elements mig
 
 ## Exceeding the Frame Bounds
 
-Allowing a popup/top-layer element to exceed the bounds of its containing frame poses a serious security risk: such an element could spoof browser UI or containing-page content. While the [original `<popup>` proposal](https://open-ui.org/components/popup.research.explainer#goals) did not discuss this issue, the [`<selectmenu>` proposal](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ControlUICustomization/explainer.md#security) does have a specific section at least mentioning this issue. Some top-layer APIs (e.g. the fullscreen API) make it possible for an element to exceed the frame bounds in some cases, great care must be taken in these cases to ensure user safety. Given the complete flexibility offered by the Popup API (any element, arbitrary content, etc.), there would be no way to ensure the safety of this feature if it were allowed to exceed frame bounds.
+Allowing a popup/top-layer element to exceed the bounds of its containing frame poses a serious security risk: such an element could spoof browser UI or containing-page content. While the [original `<popup>` proposal](https://open-ui.org/components/popup.research.explainer-v1) did not discuss this issue, the [`<selectmenu>` proposal](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ControlUICustomization/explainer.md#security) does have a specific section at least mentioning this issue. Some top-layer APIs (e.g. the fullscreen API) make it possible for an element to exceed the frame bounds in some cases, great care must be taken in these cases to ensure user safety. Given the complete flexibility offered by the Popup API (any element, arbitrary content, etc.), there would be no way to ensure the safety of this feature if it were allowed to exceed frame bounds.
 
 For completeness, several use counters were added to Chromium to measure how often this type of behavior (content exceeding the frame bounds) might be needed. These are approximations, as they merely measure the total number of times one of the built-in “popup” windows, which can exceed frame bounds because of their carefully-controlled content, is shown. The popups included in this count include the `<select>` popup, the `<input type=color>` color picker, and the `<input type=date/etc>` date/time picker. Data can be found here:
 
--- a/research/src/pages/popup/popup.research.explainer.mdx
+++ b/research/src/pages/popup/popup.research.explainer.mdx
@@ -181,7 +181,7 @@ The styling for a popup is provided by roughly the following UA stylesheet rules
 }
 ```
 
-The above rules mean that a popup, when not "shown", has `display:none` applied, and that style is removed when one of the methods above is used to show the popup. Note that the `display:none` UA stylesheet rule is **not** `!important`. In other words, developer style rules can be used to override this UA style to make a not-showing popup visible in the page. In this case, the popup will **not** be displayed in the top layer, but rather at it's ordinary z-index position within the document. This can be used, for example, to animate the show/hide behavior of the popup, or make popup content "return to the page" instead of becoming hidden.
+The above rules mean that a popup, when not "shown", has `display:none` applied, and that style is removed when one of the methods above is used to show the popup. Note that the `display:none` UA stylesheet rule is **not** `!important`. In other words, developer style rules can be used to override this UA style to make a not-showing popup visible in the page. In this case, the popup will **not** be displayed in the top layer, but rather at it's ordinary `z-index` position within the document. This can be used, for example, to animate the show/hide behavior of the popup, or make popup content "return to the page" instead of becoming hidden.
 
 
 
--- a/research/src/pages/popup/popup.research.explainer.mdx
+++ b/research/src/pages/popup/popup.research.explainer.mdx
@@ -231,11 +231,11 @@ Neither of these events are cancellable, and both are fired asynchronously.
 
 ## Focus Management
 
-Elements that move into the top layer sometimes need to move the focus to that element, and sometimes don't. For example, a modal `<dialog>` gets automatically focused because a dialog is something that requires immediate focus and attention. On the other hand, a `<div popup=hint>` doesn't receive focus at all (and is typically not expected to contain focusable elements). Similarly, a `<div popup=async>` should not receive focus (even if focusable) because it is meant for out-of-band communication of state, and should not interrupt a user's current action. Additionally, if the top layer element **should** receive immediate focus, there is a question about **which** part of the element gets that initial focus. For example, the element itself could receive focus, or one of its focusable descendants could receive focus first. To provide control over these behaviors, two attributes can be used on popups:
+Elements that move into the top layer may require focus to be moved to that element, or a descendant element. However, not all elements in the top layer will require focus. For example, a modal `<dialog>` will have focus set to its first interactive element, if not the dialog element itself, because a modal dialog is something that requires immediate attention. On the other hand, a `<div popup=hint>` (which will more often than not represent a "tooltip") does not receive focus at all (nor is it expected to contain focusable elements). Similarly, a `<div popup=async>` should not immediately receive focus (even if it contains focusable elements) because it is meant for out-of-band communication of state, and is not meant to interrupt a user's current action. Additionally, if the top layer element **should** receive immediate focus, there is a question about **which** part of the element gets that initial focus. For example, the element itself could receive focus, or one of its focusable descendants could receive focus. To provide control over these behaviors, two attributes can be used on popups:
 
-- `autofocus`. When present on a popup or the descendant of a popup, this causes focus to be moved to that element when the popup is shown. Note that `autofocus` is [already a global attribute](https://html.spec.whatwg.org/multipage/interaction.html#the-autofocus-attribute), but the existing behavior applies to element focus on **page load**. This proposal extends that definition to be used within popups, and the focus behavior happens **when they are shown**. Note that adding `autofocus` to a popup descendant does **not** cause the popup to be shown on page load, and therefore it does not cause focus to be moved into the popup **on page load**, unless the `defaultopen` attribute is also used.
+- `autofocus`. When present on a popup or one of its descendants, it will result in focus being moved to the specified element when the popup is rendered. Note that `autofocus` is [already a global attribute](https://html.spec.whatwg.org/multipage/interaction.html#the-autofocus-attribute), but the existing behavior applies to element focus on **page load**. This proposal extends that definition to be used within popups, and the focus behavior happens **when they are shown**. Note that adding `autofocus` to a popup descendant does **not** cause the popup to be shown on page load, and therefore it does not cause focus to be moved into the popup **on page load**, unless the `defaultopen` attribute is also used.
 
-- `delegatesfocus`. When present on the popup element itself, this causes the **first focusable descendent** of the popup to be focused when the popup is shown. (Eventually, this attribute might be made applicable to any element, not just popups, to control focus behavior more generally.)
+- `delegatesfocus`. When present on the popup element itself, this causes the **first focusable descendent** of the popup to be focused when the popup is shown. The purpose of this attribute is to handle the cases in which the popup content or ordering is not known prior to rendering, so that `autofocus` cannot be used effectively. (Eventually, the `delegatesfocus` attribute might be made applicable to any element, not just popups, to control focus behavior more generally.)
 
 If both `autofocus` and `delegatesfocus` are both present on the popup element, the popup element is focused when shown.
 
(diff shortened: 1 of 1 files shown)


research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -342,80 +479,49 @@ This section contains several HTML examples, showing how various UI elements mig
...
-</table>
+## Exceeding the Frame Bounds
 
+Allowing a popup/top-layer element to exceed the bounds of its containing frame poses a serious security risk: such an element could spoof browser UI or containing-page content. While the [original `<popup>` proposal](https://open-ui.org/components/popup.research.explainer#goals) did not discuss this issue, the [`<selectmenu>` proposal](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ControlUICustomization/explainer.md#security) does have a specific section at least mentioning this issue. Some top-layer APIs (e.g. the fullscreen API) make it possible for an element to exceed the frame bounds in some cases, great care must be taken in these cases to ensure user safety. Given the complete flexibility offered by the Popup API (any element, arbitrary content, etc.), there would be no way to ensure the safety of this feature if it were allowed to exceed frame bounds.
```
andrico1234, line 487: Unless I've made a silly mistake, won't the following: 

`While the [original `<popup>` proposal](https://open-ui.org/components/popup.research.explainer#goals)...`

just link to the current page's goals heading? 

Is there a different document that this link should point to, or is the wording just a little ambiguous?

I'd figure that the _original <popup> proposal_ would be an older spec (maybe the abandoned one?)

andrico1234:
I did a one-pass of the preview site, and the document looks great and well-formatted!

Happy to approve, but not sure if you want someone more familiar with the spec to review the content :)

andrico1234 (approved):

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -83,59 +93,97 @@ So this markup represents popup content:
...
+[popup i]:not(:popup-open) {
+  display: none;
+}
+
+[popup i] {
+  position: fixed;
+}
+```
+
+The above rules mean that a popup, when not "shown", has `display:none` applied, and that style is removed when one of the methods above is used to show the popup. Note that the `display:none` UA stylesheet rule is **not** `!important`. In other words, developer style rules can be used to override this UA style to make a not-showing popup visibile in the page. In this case, the popup will **not** be displayed in the top layer, but rather at it's ordinary z-index position within the document. This can be used, for example, to animate the show/hide behavior of the popup, or make popup content "return to the page" instead of becoming hidden.
```
scottaohara, line 184:
```suggestion
The above rules mean that a popup, when not "shown", has `display:none` applied, and that style is removed when one of the methods above is used to show the popup. Note that the `display:none` UA stylesheet rule is **not** `!important`. In other words, developer style rules can be used to override this UA style to make a not-showing popup visible in the page. In this case, the popup will **not** be displayed in the top layer, but rather at it's ordinary `z-index` position within the document. This can be used, for example, to animate the show/hide behavior of the popup, or make popup content "return to the page" instead of becoming hidden.
```

scottaohara (commented):
was not able to get through the whole doc, but two things for now.

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -17,35 +17,47 @@ pathToResearch: /components/popup.research
...
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 # Background
 
-There is a need in the Web Platform for an API to create "popup UI". This is a general class of UI that always appear on top of all other content, and have both one-at-a-time and "light-dismiss" behaviors. This document proposes a set of APIs to make this type of UI easy to build.
+A very common UI pattern on the Web, for which there is no native API, is "popup UI" or "popups". Popups are a general class of UI that have three common behaviors:
+ 1. Popups always appear **on top of other page content**.
+ 2. Popups are **ephemeral**. When the user "moves on" to another part of the page (e.g. by clicking elsewhere, or hitting ESC), the popup closes.
+ 3. Popups are **"one at a time"** - opening one popup closes others.
```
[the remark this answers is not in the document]
mfreed7: Yep, done, thanks!

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Great catch! That's from before I moved the explainer to here.

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Thanks! done.

mfreed7:
Thanks for the reviews. @scottaohara let me know if you want me to wait for you to take a look at the rest before I land it.

scottaohara:
going to finish up now.

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -155,27 +203,138 @@ function supportsPopup() {
...
+
+- `autofocus`. When present on a popup or the descendant of a popup, this causes focus to be moved to that element when the popup is shown. Note that `autofocus` is [already a global attribute](https://html.spec.whatwg.org/multipage/interaction.html#the-autofocus-attribute), but the existing behavior applies to element focus on **page load**. This proposal extends that definition to be used within popups, and the focus behavior happens **when they are shown**. Note that adding `autofocus` to a popup descendant does **not** cause the popup to be shown on page load, and therefore it does not cause focus to be moved into the popup **on page load**, unless the `defaultopen` attribute is also used.
```
scottaohara, line 236: note that this use of `autofocus` is also relevant to the initial focus placement doc/wiki that was worked on for the dialog element.  so there is pending precedent for this proposal for the attribute.

```suggestion
- `autofocus`. When present on a popup or one of its descendants, it will result in focus being moved to the specified element when the popup is rendered. Note that `autofocus` is [already a global attribute](https://html.spec.whatwg.org/multipage/interaction.html#the-autofocus-attribute), but the existing behavior applies to element focus on **page load**. This proposal extends that definition to be used within popups, and the focus behavior happens **when they are shown**. Note that adding `autofocus` to a popup descendant does **not** cause the popup to be shown on page load, and therefore it does not cause focus to be moved into the popup **on page load**, unless the `defaultopen` attribute is also used.
```

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -155,27 +203,138 @@ function supportsPopup() {
...
+
+- `delegatesfocus`. When present on the popup element itself, this causes the **first focusable descendent** of the popup to be focused when the popup is shown. (Eventually, this attribute might be made applicable to any element, not just popups, to control focus behavior more generally.)
```
scottaohara, line 238: Would it be good to call out that the purpose of such an attribute would be for instances where the subtree of the popup may not be known prior to its rendering?  Otherwise, it seems like something that specifying `autofocus` on the first focusable element of a popup could have taken care of?

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -155,27 +203,138 @@ function supportsPopup() {
...
+# Behaviors
+
+
+
+## Automatic Dismiss Behavior
+
+For elements that are displayed on the top layer via this API, there are a number of things that can cause the element to be removed from the top-layer, besides the ones [described above](#showing-and-hiding-a-popup). These fall into three main categories:
+
+* **One at a Time.** Another element being added to the top-layer causes existing top layer elements to be removed. This is typically used for “one at a time” type elements: when one popup is shown, other popups should be hidden, so that only one is on-screen at a time. This is also used when “more important” top layer elements get added. For example, fullscreen elements should close all open popups.
+* **Light Dismiss**. User action such as clicking outside the element, hitting Escape, or causing keyboard focus to leave the element should all cause a showing popup to be hidden. This is typically called “light dismiss”, and is discussed in more detail in [this section](#light-dismiss).
```
scottaohara, line 275: just a consistency nit, putting the period inside the bold like the previous

```suggestion
* **Light Dismiss.** User action such as clicking outside the element, hitting Escape, or causing keyboard focus to leave the element should all cause a displayed popup to be hidden. This is typically called “light dismiss”, and is discussed in more detail in [this section](#light-dismiss).
```

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -155,27 +203,138 @@ function supportsPopup() {
...
 * **Other Reasons**. Because the top layer is a UA-managed resource, it may have other reasons (for example a user preference) to forcibly remove elements from the top layer.
```
scottaohara, line 276: same same
```suggestion
* **Other Reasons.** Because the top layer is a UA-managed resource, it may have other reasons (for example a user preference) to forcibly remove elements from the top layer.
```

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -155,27 +203,138 @@ function supportsPopup() {
...
 
-In all such cases, the UA manages the removal of elements from the top layer, by forcibly removing the element from the top layer, and re-applying the `display:none` UA rule.
+In all such cases, the UA is allowed to forcibly remove an element from the top layer and re-apply the `display:none` popup UA rule. The rules the UA uses to manage these interactions depends on the element types, and this is described in more detail in [this section](#classes-of-top-layer-ui).
+
+### Light Dismiss
+
+The term "light dismiss" for a popup is used to describe the user "moving on" to another part of the page, or generally being done interacting with the popup. When this happens, the popup should be hidden. Several actions trigger light dismiss of a popup:
+
+1. Clicking or tapping outside the bounds of the popup. Any `mousedown` event will trigger **all open popups** to be hidden, starting from the top of [the popup stack](#the-popup-stack), and ending with the [nearest open ancestral popup](#nearest-open-ancestral-popup) of the `mousedown` event's `target` `Node`. This means clicking on a popup or its trigger or anchor elements will not hide that popup.
```
scottaohara, line 284: suggestion, maybe `mouseup` would be good here?  

I say this because [WCAG 2.5.2 Pointer Cancellation criterion](https://www.w3.org/WAI/WCAG21/Understanding/pointer-cancellation.html) could then be adhered to, allowing someone who makes an accidental down click outside of a popup (someone with motor disabilities, for instance) would have the chance to move their pointer back into the boundaries of the popup, allowing for the accidental dismissal to be aborted.

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -155,27 +203,138 @@ function supportsPopup() {
...
+2. Hitting the ESC key, or generally any ["close signal"](#close-signal). This will cause the topmost popup on [the popup stack](#the-popup-stack) to be hidden.
```
scottaohara, line 285: matching earlier use of the full word, rather than the abbreviated key.

```suggestion
2. Hitting the Escape key, or generally any ["close signal"](#close-signal). This will cause the topmost popup on [the popup stack](#the-popup-stack) to be hidden.
```

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -185,107 +344,85 @@ The rules the UA uses to manage these interactions depends on the element types,
...
 
+In the table, "hide" means that when the second element is shown (enters the top layer), the first element is removed from the top layer. In contrast, "leave" means both elements will remain in the top layer together.
 
-## Events
-
-There is a [general desire](https://github.com/openui/open-ui/issues/342) to receive events indicating that an element has entered or left the top layer via this API. These events can be used, for example, to populate content for a popup just in time before it is shown, or update server data when it closes.
+### Detailed description of interactions among popup types
 
-It would seem most natural and powerful to define two events:
+This section describes, in detail, the interactions between the three popup types:
```
scottaohara, line 373:
```suggestion
This section details the interactions between the three popup types:
```

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -185,107 +344,85 @@ The rules the UA uses to manage these interactions depends on the element types,
...
 
 
+## Accessibility / Semantics
 
-1. Establish the provided anchor element as an “ancestor” of this popup, for the purposes of light-dismiss behavior. In other words, when a popup is shown, typically all other popups are closed. An exception would be made for all popups in the ancestor chain formed by the anchor element.
-2. The referenced anchor element could be used by the [Anchor Positioning feature](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/CSSAnchoredPositioning/explainer.md).
+Since the `popup` content attribute can be applied to any element, and this only impacts the element's presentation (top layer vs not top layer), this addition does not have any direct semantic or accessibility impact. The element with the `popup` attribute will keep its existing semantics and AOM representation. For example, `<img popup=popup>` will continue to have `role=img`, but will be able to be displayed on top of other content. Similarly, ARIA can be used to modify accessibility mappings in [the normal way](https://w3c.github.io/html-aria/), for example `<img popup=popup role=presentation>`.
```
scottaohara, line 389: Swapping around the examples a bit because while I still think they should be valid, using an image as a popup could definitely have its use cases (i'm immediately thinking of "here's where your credit card code is"), but not the common use case I'd imagine people would go for.  

```suggestion
Since the `popup` content attribute can be applied to any element, and this only impacts the element's presentation (top layer vs not top layer), this addition does not have any direct semantic or accessibility impact. The element with the `popup` attribute will keep its existing semantics and AOM representation. For example, `<article popup=popup>...</article>` will continue to be exposed as an implicit `role=article`, but will be able to be displayed on top of other content. Similarly, ARIA can be used to modify accessibility mappings in [the normal way](https://w3c.github.io/html-aria/), for example `<div popup=popup role=note>...</div>`.
```

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -185,107 +344,85 @@ The rules the UA uses to manage these interactions depends on the element types,
...
 
+As mentioned in the [Declarative Triggers](#declarative-triggers) section, accessibility mappings will be automatically configured to associate the popup with its trigger element, as needed.
 
-## Accessibility / Semantics
+## Disallowed elements
 
-Since the `popup` content attribute can be applied to any element, and only impacts the element's presentation (top layer vs not top layer), this should not have **any semantic or accessibility impact**. I.e., the element with the `popup` attribute will keep its existing semantics and AOM representation. In the event an author needs to extend or modify a particular element's ARIA semantics, this may be done in accordance to existing allowances of [ARIA in HTML](https://w3c.github.io/html-aria/).
+While the popup API can be used on most elements, there are some limitations. For example, calling `showPopup()` on a `<dialog>` element will result in an exception being thrown, as will calling it on an [active fullscreen element](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen). All other element types are valid.
```
scottaohara, line 395: I do not think a dialog should be completely barred from being a popup (unless i'm misunderstanding this line?)  it's also in opposition with the example that follows in the next section, re: `<my-date-picker role=dialog id=datepicker popup=popup>`

can expand on this if need be... but maybe this should just be clarified to mean a 'modal' dialog would throw an exception?  (even though there are tons of instances of custom modal dialogs where people implement a click-outside-to-close-light-dismiss... but, i have too many other hills i've decided to die on...)

scottaohara (commented):
OK, have finished reviewing.  Thank you for all the work on this.  Please let me know if I need to add further clarification to any of my comments / suggestions

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: I've gone back and forth on this. But I do see your point here that `role=dialog` semantics might be entirely appropriate on something that has popup behaviors (like click-outside light dismiss).

I've added explicitly "modal" dialog here.

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Awesome, thanks. I struggled with a good example here - this is better.

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Nice

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Done

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: This is likely worth discussion. I did a rather informal survey of native and web popups, and found that nearly all of them dismiss on mouse-down. Think it's worth an issue?

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Thanks! I hate when those aren't consistent.

On the `displayed` vs `showing`, I fought with that throughout. I wanted to avoid "open"/"closed" because we're using "show"/"hide", but it's awkward in many cases. I like "displayed", I might try to use that more.

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Yep, good point. Done.

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Accepted.

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -155,27 +203,138 @@ function supportsPopup() {
...
-* **Light Dismiss**. User action such as clicking outside the element, hitting Escape, or causing keyboard focus to leave the element. This is typically called “light dismiss”.
+Elements that move into the top layer sometimes need to move the focus to that element, and sometimes don't. For example, a modal `<dialog>` gets automatically focused because a dialog is something that requires immediate focus and attention. On the other hand, a `<div popup=hint>` doesn't receive focus at all (and is typically not expected to contain focusable elements). Similarly, a `<div popup=async>` should not receive focus (even if focusable) because it is meant for out-of-band communication of state, and should not interrupt a user's current action. Additionally, if the top layer element **should** receive immediate focus, there is a question about **which** part of the element gets that initial focus. For example, the element itself could receive focus, or one of its focusable descendants could receive focus first. To provide control over these behaviors, two attributes can be used on popups:
```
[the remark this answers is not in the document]
mfreed7: I like it. Not sure why the diff is terrible, but I got it.

mfreed7:
> OK, have finished reviewing. Thank you for all the work on this. Please let me know if I need to add further clarification to any of my comments / suggestions

Thanks for the great review! I resolved everything except one question. It might warrant a new issue to discuss further.

mfreed7:
I'm going to land this now, but let me know what you think about the mouse-down vs. mouse-up question and I can open an issue for that.

research/src/pages/popup/popup.research.explainer.mdx
scottaohara: i think it's definitely worth at least a discussion.

there are presently a lot of platform behaviors that get a 'pass' with wcag rules because web authors can literally do nothing about them, since they're baked into the platform.  but, if given the chance to change a behavior, it at least deserves being brought up and then further discussion can dictate if there's a reason to adhere to what's been previously done vs going in a different direction.  

thanks!

mfreed7: Filed! https://github.com/openui/open-ui/issues/529

PR #542, opened 2022-05-30
Changing popup="popup" to popup in the selectmenu usage docs

captainbrosset:
Per https://chromium-review.googlesource.com/c/chromium/src/+/3668816 the popup attribute syntax is changing.

It used to require `popup="popup"`. It has been renamed to `popup="auto"` and also now supports the boolean-like behavior so `popup` just works.

This PR changes the `selectmenu` usage docs to match this.

### DIFF
--- a/research/src/pages/prototypes/selectmenu.mdx
+++ b/research/src/pages/prototypes/selectmenu.mdx
@@ -156,7 +156,7 @@ You can replace the default listbox part in a similar fashion:
 </style>
 <selectmenu class="my-custom-select">
   <div slot="listbox">
-    <div popup="popup" behavior="listbox">
+    <div popup behavior="listbox">
       <option>Option 1</option>
       <option>Option 2</option>
       <option>Option 3</option>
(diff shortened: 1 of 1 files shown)


captainbrosset:
cc @mfreed7 and @dandclark for a review on this.

mfreed7 (approved):
LGTM, thanks for updating this.

dandclark (approved):
GitHub won't let me leave an inline comment for this, but can you also change the instances of `<div popup=popup>` on line 170 to `<div popup>` so that it matches the code sample?

Thanks for making these updates!

captainbrosset:
> can you also change the instances of `<div popup=popup>` on line 170 to `<div popup>` so that it matches the code sample?

Thanks for catching this! I had missed it. Will add a commit in a second.

captainbrosset:
@andrico1234 I guess this should be ready to merge, when you have some time. Thank you!

andrico1234 (approved):

PR #550, opened 2022-06-22
Accessibility pass over focusgroup examples

travisleithead:
* use of semantic elements where custom elements aren't really important for understanding
* ensure ARIA roles are present for custom elements for better clarity
* fix nested menuitems (Example 11), see https://github.com/w3ctag/design-reviews/issues/732#issuecomment-1159814503
* Consistency pass for elipses use

### DIFF
--- a/research/src/pages/focusgroup/focusgroup.explainer.mdx
+++ b/research/src/pages/focusgroup/focusgroup.explainer.mdx
@@ -234,16 +234,16 @@ the roving tabindex, but by adding `focusgroup`, no JavaScript is necessary to m
 Example 2:
 
 ```html
-<my-list focusgroup>
-  <my-listitem tabindex="0">...</mylistitem>
-  <my-listitem tabindex="-1">...</mylistitem>
-  <my-listitem tabindex="-1">...</mylistitem>
-  <my-listitem tabindex="-1">...</mylistitem>
-  <my-listitem tabindex="-1">...</mylistitem>
-</my-list>
+<ul focusgroup>
+  <li tabindex="0">...</li>
+  <li tabindex="-1">...</li>
+  <li tabindex="-1">...</li>
+  <li tabindex="-1">...</li>
+  <li tabindex="-1">...</li>
+</ul>
 ```
 
-From outside of the above markup, assume a Tab key press lands focus on the first `<my-listitem>`. From
+From outside of the above markup, assume a Tab key press lands focus on the first `<li>`. From
 this point, the user can use the arrow keys to move from the beginning of the list to the end, or press
 Tab again to move outside of this group. Alternatively, maybe you want to support both Tab _and_ Arrow
 key navigation in your scenario. No problem, just change the setup so that all the list items participate
@@ -252,13 +252,13 @@ in the tab order:
 Example 3:
 
 ```html
-<my-list focusgroup>
-  <my-listitem tabindex="0">...</mylistitem>
-  <my-listitem tabindex="0">...</mylistitem>
-  <my-listitem tabindex="0">...</mylistitem>
-  <my-listitem tabindex="0">...</mylistitem>
-  <my-listitem tabindex="0">...</mylistitem>
-</my-list>
+<ul focusgroup>
+  <li tabindex="0">...</li>
+  <li tabindex="0">...</li>
+  <li tabindex="0">...</li>
+  <li tabindex="0">...</li>
+  <li tabindex="0">...</li>
+</ul>
 ```
 
 The focusgroup doesn't "remember" the last element that was focused when focus leaves the group
--- a/research/src/pages/focusgroup/focusgroup.explainer.mdx
+++ b/research/src/pages/focusgroup/focusgroup.explainer.mdx
@@ -235,11 +235,11 @@ Example 2:
 
 ```html
 <ul focusgroup>
-  <li tabindex="0">...</li>
-  <li tabindex="-1">...</li>
-  <li tabindex="-1">...</li>
-  <li tabindex="-1">...</li>
-  <li tabindex="-1">...</li>
+  <li tabindex="0">…</li>
+  <li tabindex="-1">…</li>
+  <li tabindex="-1">…</li>
+  <li tabindex="-1">…</li>
+  <li tabindex="-1">…</li>
 </ul>
 ```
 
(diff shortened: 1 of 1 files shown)


research/src/pages/focusgroup/focusgroup.explainer.mdx
```diff
@@ -361,48 +361,50 @@ When extending a focusgroup, traversal order is based on document order. Given t
...
 ```html
 <!-- This is an example of what NOT TO DO -->
-<div focusgroup>
-  <div id="A" tabindex="0"></div>
-  <div id="B" tabindex="-1" focusgroup="extend wrap">
-    <!-- 'wrap' will not apply -->
-    <div id="B1" tabindex="-1"></div>
-    <div id="B2" tabindex="-1"></div>
+<div role=toolbar focusgroup>
+  <button id="A" type=button tabindex="0"></button>
```
benbeaudry, line 392: Does specifying the `type=button` attribute bring something valuable to the example? Since we're not inside of a form here, the button won't have any action, right?

Disclaimer: I am not an expert on the button type attribute and might be missing something.

research/src/pages/focusgroup/focusgroup.explainer.mdx
```diff
@@ -480,14 +482,14 @@ linear group as described previously.
...
-<vertical-menu focusgroup="vertical wrap">
-  <menu-item tabindex="-1">Action 1</menu-item>
-  <menu-item tabindex="0">Action 2</menu-item>
-  <menu-group focusgroup="vertical extend">
-    <menu-item tabindex="-1">Sub-Action 3</menu-item>
-    <menu-item tabindex="-1">Sub-Action 4</menu-item>
+<vertical-menu role=menu focusgroup="vertical wrap">
+  <menu-item role=menuitem tabindex="-1">Action 1</menu-item>
+  <menu-item role=menuitem tabindex="0">Action 2</menu-item>
+  <menu-group role=none focusgroup="vertical extend">
```
benbeaudry, line 488: Shouldn't the role be menu here? 

> Submenus, also known as pop-up menus, also have the role menu.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/roles/menu_role#description

research/src/pages/focusgroup/focusgroup.explainer.mdx
```diff
@@ -561,21 +564,24 @@ linear focusgroups (and vice-versa) as a best practice. The following example is
...
+      <menu-item role=menuitem tabindex="-1">Sub-Action 4</menu-item>
+    </vertical-menu>
+  </menu-item>
   <menu-item tabindex="-1">Action 5</menu-item>
 </horizontal-menu>
 ```
 
-When focus is on the `<vertical-menu>` focusgroup item, only a down arrow key will descend into the
+When focus is on the "Action 3" menu item, only a down arrow key will descend into the
 nested focusgroup (not a right arrow key because this is disallowed by the vertical extending
```
benbeaudry, line 582: My mental representation of the focusgroup feature got rusty so please take this comment with a grain of salt: I _think_ that if we are indeed on the menu-item of line 570 and we press the down arrow, nothing would happen.

We would need to be on the root of the focusgroup that extend in the orthogonal direction for a key press in the orthogonal direction to move the focus.

benbeaudry: In other words, the menu-item at line 570 should be the one having `focusgroup="extend vertical"`

benbeaudry (commented):
I have a few questions/comments, but generally LGTM!

research/src/pages/focusgroup/focusgroup.explainer.mdx
```diff
@@ -361,48 +361,50 @@ When extending a focusgroup, traversal order is based on document order. Given t
...
-      <div focusgroup="extend">
-        <div id="B1" tabindex="-1"></div>
-        <div id="B2" tabindex="-1"></div>
+<div role=toolbar focusgroup>
+  <button id="A" type=button tabindex="0"></button>
+  <button id="B" type=button tabindex="-1"></button>
+  <div>
+      <div role=radiogroup focusgroup="extend">
+        <button id="B1" role=radio type=button tabindex="-1"></button>
+        <button id="B2" role=radio type=button tabindex="-1"></button>
```
chrisdholt, line 370: Part of me wants to mention that we probably _shouldn't_ add a role here to button and instead keep it a div. If the intent is OOTB illustration using just HTML, I don't know that radiogroup would yet fit into that...just a note :)

andrico1234:
I did a quick once-over on the preview, and all looks good! Happy to merge after the comments have been resolved :) 

Side note: I did notice that one of the images wasn't loading but it doesn't seem to be as a result of this pull request. If the issue persists after deploy I'll whip up a fix.

research/src/pages/focusgroup/focusgroup.explainer.mdx
travisleithead: I'm OK going back to a div. (I think I borrowed this particular arrangement from an ARIA authoring practices demo, but you have a point.)

research/src/pages/focusgroup/focusgroup.explainer.mdx
travisleithead: I'm pretty sure that's standard practice to make it a simple push-button, otherwise it ends up looking like a submit button without anything to submit... which is weird. (Honestly, I think HTML got it wrong by having submit button be the default, but can't change that now.)

research/src/pages/focusgroup/focusgroup.explainer.mdx
travisleithead: If this was a submenu, that would be correct, but in fact, the "menu-group" is just acting like a generic div for some grouping purpose. The example is highlighting that the vertical extend is just making a continuation of the existing vertical menu (which is assumed to be structurally inline with the other menu options; who knows what the visual presentation is doing.)

Maybe just make it a 'div' instead?

research/src/pages/focusgroup/focusgroup.explainer.mdx
travisleithead: I know that held true for the time when we had strict parent->child relationships for focusgroups... but when we changed extend to search up the tree to make a connection, I wonder if that changed? E.g., is that linkage for the extending focusgroup bidirectional to allow a descent even if the descender is not an immediate child?

travisleithead: (might have to fix the polyfil for this as well...)

andrico1234 (approved):

andrico1234:
@travisleithead ,I've given my approval 😄. Are you happy for me to merge or are there additional changes you'd like to make?

research/src/pages/focusgroup/focusgroup.explainer.mdx
travisleithead: Fixed in https://github.com/openui/open-ui/pull/550/commits/cc22a4a335754d510e3e42a4acf885821db0f45d

research/src/pages/focusgroup/focusgroup.explainer.mdx
travisleithead: Will keep as `menu-group` for now.

travisleithead:
@andrico1234 ready for merging now, I believe. Thanks!

Issue #581, opened 2022-08-16
[popup] Add further clarity that popup is not (presently?) for modal dialogs


scottaohara:
[Per discussion on the open ui discord](https://discord.com/channels/714891843556606052/961979349639245854/1008867345407684769) initiated by @tbondwilkinson, there is some value in further identifying the differences between a "popup" and a "modal dialog" and why [the disallowed elements section](https://open-ui.org/components/popup.research.explainer#disallowed-elements) calls them out as not supporting popup.  

Potentially adding to the goals section (or non-goals), examples, or even expanding the initial background section to help further clarify the differences between UI elements that "float" on screen and how not all floating content should be considered a "popup".

@tbondwilkinson please do add any of your thoughts here, as I'm sure my brief intro to the topic is not doing justice to your concerns.

tbondwilkinson:
I think as a component author coming to see whether the Popup API is a good fit for specific components, having a good sense of where the Popup API is or is not fit would be helpful.

Things like:

1. Content is usually ephemeral, and does not stand alone from the content of the page
2. Popups are displayed one a time, and if others should dismiss when one is shown.
3. Popups leave the rest of the page interactable
4.  Popups may anchor to specific elements or they may float above the page

Which would highlight the expected behavior.

bkardell:
Is #1 inherent? I've been talking a bit to some other folks trying to understand the boundaries of popup use cases and whether or not popups were _inherently_ popups, or they could be sometimes afforded that particular nature.  For example, what about 'hamburger menus' - they are often just content, unless they are acting as a kind of popup because we've hidden the content by default. Or, in a form I might have some helpful text about what belongs in the field ("Your password must contain...") which we'd like to 'hint' one at a time on most devices, but... maybe if I had a reason to print that form I'd like to just display it inline.  If either of those is potentially viable use cases for popup it seems to make #1 a complicated criteria to me.

tbondwilkinson:
I think @mfreed7 mentioned that requirement (or was it more of a suggestion?) to me. Personally, I think it's reasonable for popup to serve use cases that are beyond just supporting content. I would love to hear other's thoughts.

scottaohara:
re:
>they are often just content, unless they are acting as a kind of popup because we've hidden the content by default.

does it overlay other content of the page and it would show/hide based on behaviors that are not commonly associated with the default semantics of the elements in question?  e.g., a list of navigational links?  Then that's a popup.

[...]

PR #593, opened 2022-08-23
Update with recent OpenUI discussion results

mfreed7:
The largest change is the result of [this discussion](https://github.com/openui/open-ui/issues/581) about the differences between `<dialog>` and `popup=manual`. There's a new section attempting to describe those differences. This PR also includes recent changes to the `show` event, and an update on the interaction between top layer element types.

This PR also removes an obsolete section of the `<selectmenu>` page that talks about replaceable shadow DOM.

Closes https://github.com/openui/open-ui/issues/581


research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -75,6 +75,32 @@ This proposal was discussed on [Issue 455](https://github.com/openui/open-ui/iss
...
+
+- A `<div popup=manual>` is **in the top layer**, so it draws on top of other content. The same is not true for a non-modal `<dialog>`. This is likely the most impactful difference, as it tends to be difficult to ensure that a non-modal `<dialog>` is painted on top of other page content.
+- A `<dialog>` element always has **`role=dialog`**, while the `popup` attribute can be applied to the **most-semantically-relevant HTML element**, including the `<dialog>` element itself: `<dialog popup>`.
+- The pop-up API comes with some **"nice to have's"**:
+    - pop-ups are easy to animate both the show and hide transitions. JS is required to animate `dialog.close()`.
+    - pop-ups fire both a "show" and a "hide" event. A `<dialog>` only fires `cancel` and `close`, but no event is fired when it is shown.
+    - pop-ups work with the invoking attributes (e.g. `popuptoggletarget`) to declaratively show/hide them. JS is required to show/close a `<dialog>`.
+
+For the above reasons, it seems clear that `<div popup=manual>` is superior to non-modal `<dialog>` in most cases, and should be preferred.
```
scottaohara, line 101: this is the only line i don't completely agree with - specifically because of the statement `div popup=manual` is superior' and that it should be preferred. 

I'd rather this state something like if someone needs a modeless dialog to have popup behavior, that they use `<dialog popup / popup=manual>`.  if they need a popup that is not meant to be exposed as a dialog, then use `<div popup=...>` and as necessary add the appropriate `role`.

scottaohara (commented):
just one comment.  thanks!

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Yep, great point. Take a look at the new version and LMK if that works for you.

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -357,7 +381,7 @@ The intention of the above set of behaviors is to return focus to the previously
 A new attribute, `anchor`, can be used on a pop-up element to refer to the pop-up's "anchor". The value of the attribute is a string idref corresponding to the `id` of another element (the "anchor element"). This anchoring relationship is used for two things:
 
 1. Establish the provided anchor element as an [“ancestor”](#nearest-open-ancestral-pop-up) of this pop-up, for light-dismiss behaviors. In other words, the `anchor` attribute can be used to form nested pop-ups.
-2. The referenced anchor element could be used by the [Anchor Positioning feature](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/CSSAnchoredPositioning/explainer.md).
+2. The referenced anchor element could be used by the **Anchor Positioning feature**. A (dated) explainer can be found [here](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/CSSAnchoredPositioning/explainer.md), or a more up-to-date draft spec [here](https://tabatkins.github.io/specs/css-anchor-position/).
 
 
 ## Backdrop
```
scottaohara, line 387: the only other thing i was thinking about that should probably be called out is in relation to the backdrop.  

I do think there will be use cases where popups can present with a backdrop that obscures the background, and with light dismiss behaviors, this will help ensure that if someone wants to get away from the popup and engage with the content beneath, they'll be able to do so with relative ease.

However, it seems like this could be problematic with a `popup=manual` where there is no light dismiss behavior.  someone could keyboard escape and then be prevented from seeing where keyboard focus has moved to on the page. 

Outside of author guidance to say "dont' use backdrop on `popup=manual` or "if you need to fully obscure the content beneath a popup, maybe consider whether that popup should be a modal dialog or not?"

Alternatively, outside of barring `popup=manual` from having `::backdrop` - would it be possible to ensure that if focus were to have left a popup in this state, that the backdrop would be reset to transparent?  So it would only obscure the background so long as focus was within the popup?

research/src/pages/popup/popup.research.explainer.mdx
scottaohara: lgtm.  thank you!

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: This is an interesting point, and I've added a few more sentences to guide away from this behavior for `popup=manual`.

The only thing I'd add is that not only will keyboard navigation be broken in this case, but so will mouse (and all other) navigation. If the page is obscured, it won't be visible, but `pointer-events:none !important` will still ensure the `::backdrop` can't trap clicks either. Perhaps that makes this less of a risky thing, since the site will be broken for all users, right?

I'm afraid of trying to bar `popup=manual` from having `::backdrop`, since it'll be the only top layer element *not* to have a `::backdrop`, and perhaps there are valid use cases?

scottaohara: thumbs up to the added guidance, thank you.

i also think there _could_ be a use case for a backdrop with a `popup=manual`, so yeh i don't think it should be barred.  But here's an example of what I was thinking would be a potentially safer default - where if focus is within or hovered over the popup, the backdrop gets the styling.  But if focus is outside of the popup - and importantly - if focus is outside of the popup BUT mouse hover is still over the popup (as someone with low/no mobility may not be able to easily move their mouse and it unfortunately would just be atop a popup) the backdrop goes away since focus gets detected outside the popup, meaning it needs to be visible.

https://codepen.io/scottohara/pen/xxWoxGL

whatcha think?

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -75,6 +75,31 @@ This proposal was discussed on [Issue 455](https://github.com/openui/open-ui/iss
 
 There have been **many** discussions and resolutions at OpenUI of various aspects of this proposal, and those are listed in the [Design Decisions](#design-decisions) section.
 
+## Pop-Up vs. Dialog
+
+A natural question that [commonly arises](https://github.com/openui/open-ui/issues/581) is: what's the difference between a "pop-up" and a "dialog"? There are two ways to interpret this question:
+ 1. What are the difference between these general UX patterns and words?
+ 2. What are the technical implementation differences between `<div popup>` and `<dialog>`?
+
+In both cases, an important distinction needs to be made between **modal** and **non-modal** (or **modeless**) dialogs. With a **modal** dialog, the rest of the page (outside the dialog) is rendered **`inert`**, so that only the contents of the dialog are interactable. Importantly, a pop-up is **non-modal**. Almost by definition, a pop-up is "ephemeral" and goes away (via light dismiss) if the user changes their attention to something else, by clicking or tabbing to something else. For these reasons, if the dialog in question needs to be **modal**, then the `<dialog>` element is the way to go.
```
hidde, line 84: Might “not permanent” work here instead of ‘"ephemeral”’?

```suggestion
In both cases, an important distinction needs to be made between **modal** and **non-modal** (or **modeless**) dialogs. With a **modal** dialog, the rest of the page (outside the dialog) is rendered **`inert`**, so that only the contents of the dialog are interactable. Importantly, a pop-up is **non-modal**. Almost by definition, a pop-up is not permanent: it goes away (via light dismiss) when the user changes their attention to something else, by clicking or tabbing to something else. For these reasons, if the dialog in question needs to be **modal**, then the `<dialog>` element is the way to go.
```

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -75,6 +75,31 @@ This proposal was discussed on [Issue 455](https://github.com/openui/open-ui/iss
...
+3. Custom components that need to "manually" control pop-up behavior.
+
+Given these use cases, it's important to call out the technical differences between a non-modal `<dialog>` and a `<div popup=manual>`:
+
+- A `<div popup=manual>` is **in the top layer**, so it draws on top of other content. The same is not true for a non-modal `<dialog>`. This is likely the most impactful difference, as it tends to be difficult to ensure that a non-modal `<dialog>` is painted on top of other page content.
+- A `<dialog>` element always has **`role=dialog`**, while the `popup` attribute can be applied to the **most-semantically-relevant HTML element**, including the `<dialog>` element itself: `<dialog popup>`.
+- The pop-up API comes with some **"nice to have's"**:
+    - pop-ups are easy to animate both the show and hide transitions. JS is required to animate `dialog.close()`.
+    - pop-ups fire both a "show" and a "hide" event. A `<dialog>` only fires `cancel` and `close`, but no event is fired when it is shown.
+    - pop-ups work with the invoking attributes (e.g. `popuptoggletarget`) to declaratively show/hide them. JS is required to show/close a `<dialog>`.
```
hidde, line 99: maybe here we could say “attributes in HTML” to contrast with JS in the `<dialog>` example ?

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -357,7 +381,7 @@ The intention of the above set of behaviors is to return focus to the previously
 A new attribute, `anchor`, can be used on a pop-up element to refer to the pop-up's "anchor". The value of the attribute is a string idref corresponding to the `id` of another element (the "anchor element"). This anchoring relationship is used for two things:
 
 1. Establish the provided anchor element as an [“ancestor”](#nearest-open-ancestral-pop-up) of this pop-up, for light-dismiss behaviors. In other words, the `anchor` attribute can be used to form nested pop-ups.
-2. The referenced anchor element could be used by the [Anchor Positioning feature](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/CSSAnchoredPositioning/explainer.md).
+2. The referenced anchor element could be used by the **Anchor Positioning feature**. A (dated) explainer can be found [here](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/CSSAnchoredPositioning/explainer.md), or a more up-to-date draft spec [here](https://tabatkins.github.io/specs/css-anchor-position/).
```
hidde, line 384: Suggestion to avoid 'here' as link text: 

```suggestion
 2. The referenced anchor element could be used by the **Anchor Positioning feature** ([(dated) explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/CSSAnchoredPositioning/explainer.md); [more up-to-date draft spec](https://tabatkins.github.io/specs/css-anchor-position/)).
 ```

**Review by hidde — COMMENTED** [review]
added a few wording comments

File: research/src/pages/popup/popup.research.explainer.mdx (hunk 2 above)
Comment (mfreed7) [reply research/src/pages/popup/popup.research.explainer.mdx:387]: Cool demo! Helps me see your point, for sure.

This is an interesting suggestion. I'm not sure there's precedent for anything like this, where UA styling (and lack of developer stylability) of an element depends on where the mouse or focus is. Do you know of anything we could refer to like that?

File: research/src/pages/popup/popup.research.explainer.mdx (hunk 5 above)
Comment (mfreed7) [reply research/src/pages/popup/popup.research.explainer.mdx]: Thanks - better.

File: research/src/pages/popup/popup.research.explainer.mdx (hunk 4 above)
Comment (mfreed7) [reply research/src/pages/popup/popup.research.explainer.mdx]: Done

File: research/src/pages/popup/popup.research.explainer.mdx [hunk 6]
```diff
@@ -75,6 +75,31 @@ This proposal was discussed on [Issue 455](https://github.com/openui/open-ui/iss
...
+1. "Toasts" or asynchronous notifications, which stay onscreen until dismissed manually or via a timer.
+2. Persistent UI, such as teaching-UI, that needs to stay on screen while the user interacts with the page.
+3. Custom components that need to "manually" control pop-up behavior.
+
+Given these use cases, it's important to call out the technical differences between a non-modal `<dialog>` and a `<div popup=manual>`:
+
+- A `<div popup=manual>` is **in the top layer**, so it draws on top of other content. The same is not true for a non-modal `<dialog>`. This is likely the most impactful difference, as it tends to be difficult to ensure that a non-modal `<dialog>` is painted on top of other page content.
+- A `<dialog>` element always has **`role=dialog`**, while the `popup` attribute can be applied to the **most-semantically-relevant HTML element**, including the `<dialog>` element itself: `<dialog popup>`.
+- The pop-up API comes with some **"nice to have's"**:
+    - pop-ups are easy to animate both the show and hide transitions. JS is required to animate `dialog.close()`.
```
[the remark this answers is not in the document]
mfreed7: Ok, I think I addressed your comment - LMK if not.

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Thanks - done.

research/src/pages/popup/popup.research.explainer.mdx
scottaohara: hmm, i'm going to have to think on that if anything like this has been done at the platform level.  This is more like something i have had to inform developers on how to do, because of custom components they've built, or customizations they've made to the platform features.  I'll keep thinking though.

mfreed7: Ok, thanks. Let's keep discussing, and I'm happy to incorporate anything we come up with. But ok to land this PR in the meantime?

mfreed7:
Any other comments here?

mfreed7:
Are we ok to land this PR? @andrico1234

mfreed7:
> Are we ok to land this PR? @andrico1234

@gregwhitworth can you please help get this pushed? I keep having to refer to this PR rather than just directly to the explainer. I believe all of the comments have been addressed.

gregwhitworth:
@mfreed7 done

PR #613, opened 2022-09-30
Add motivation for attribute vs. element

mfreed7:
There have been more questions lately about why we're implementing this via a content attribute, and not, e.g. a `<popup>` element. For example, [here](https://github.com/w3ctag/design-reviews/issues/743#issuecomment-1233130208). This adds a section that goes into more detail. I also updated some links and the TOC.

@andrico1234 I could use help getting this landed soon, if possible. Thanks!

@scottaohara if you have time for a review, that'd be helpful!

### DIFF
--- a/research/src/pages/popup/popup.research.explainer.mdx
+++ b/research/src/pages/popup/popup.research.explainer.mdx
@@ -6,7 +6,7 @@ pathToResearch: /components/popup.research
 ---
 
 - [@mfreed7](https://github.com/mfreed7), [@scottaohara](https://github.com/scottaohara), [@BoCupp-Microsoft](https://github.com/BoCupp-Microsoft), [@domenic](https://github.com/domenic), [@gregwhitworth](https://github.com/gregwhitworth), [@chrishtr](https://github.com/chrishtr), [@dandclark](https://github.com/dandclark), [@una](https://github.com/una), [@smhigley](https://github.com/smhigley), [@aleventhal](https://github.com/aleventhal)
-- August 24, 2022
+- September 30, 2022
 
 Please also see the [WHATWG html spec PR for this proposal](https://github.com/whatwg/html/pull/8221).
 
@@ -17,6 +17,7 @@ Please also see the [WHATWG html spec PR for this proposal](https://github.com/w
 - [Background](#background)
   - [Goals](#goals)
   - [See Also](#see-also)
+  - [Pop-Up vs. Dialog](#pop-up-vs-dialog)
 - [API Shape](#api-shape)
   - [HTML Content Attribute](#html-content-attribute)
   - [Showing and Hiding a Pop-up](#showing-and-hiding-a-pop-up)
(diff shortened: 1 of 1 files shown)


research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -691,7 +695,46 @@ Each of these options is significantly different from the others. To properly ev
...
+
+```html
+<menu popup=auto>
+  <li><button>Item 1</button></li>
+  <li><button>Item 2</button></li>
+  <li><button>Item 3</button></li>
+</menu>
+```
+
+Similarly here, the developer is building **a menu**, so they should use the `<menu>` element. This is natural and expected.
```
scottaohara, line 722:
```suggestion
Similarly here, the developer is building **a menu**, so they should use the `<menu>` element. This is adheres to the semantic definition of a `<menu>` element, and allows the content to be programmatically exposed as a listing of buttons.
```

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -691,7 +695,46 @@ Each of these options is significantly different from the others. To properly ev
...
+
+In both cases, not only does the use of the semantically-correct element make it easier to understand the content, it also enforces the appropriate content rules and behaviors that are specific to each element. Further, using the correct element allows the UA to correctly represent the content in the accessibility tree, making assistive technologies better able to assist users in navigating the content.
+
+In an alternative proposal where the pop-up behavior is enabled via a special element, e.g. `<popup>`, all of the above would need to be carefully managed by the developer via ARIA roles and attributes:
+```html
+<popup role=dialog>
+<popup role=menu>
```
scottaohara, line 729: correction here, because although one would think that `<menu>` == `role=menu`, the menu element is really just a semantically different way of specifying a `<ul>` elment.

```suggestion
<popup role=list>
```

research/src/pages/popup/popup.research.explainer.mdx
```diff
@@ -691,7 +695,46 @@ Each of these options is significantly different from the others. To properly ev
...
+  <li><button>Item 2</button></li>
+  <li><button>Item 3</button></li>
+</menu>
+```
+
+Similarly here, the developer is building **a menu**, so they should use the `<menu>` element. This is natural and expected.
+
+In both cases, not only does the use of the semantically-correct element make it easier to understand the content, it also enforces the appropriate content rules and behaviors that are specific to each element. Further, using the correct element allows the UA to correctly represent the content in the accessibility tree, making assistive technologies better able to assist users in navigating the content.
+
+In an alternative proposal where the pop-up behavior is enabled via a special element, e.g. `<popup>`, all of the above would need to be carefully managed by the developer via ARIA roles and attributes:
```
scottaohara, line 726:
```suggestion
In an alternative proposal where the pop-up behavior is enabled via a special element, e.g. `<popup>`, all of the above would need to be carefully managed by the developer via ARIA roles and attributes as there is no single role that can consistently be used to identify a `<popup>` element:
```

scottaohara (commented):
here you go @mfreed7 

lemme know if you think i could expand on anything further.

andrico1234 (approved):

andrico1234:
Approved! :)

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Ugh. Really? Ugh. Ok, done. 😢

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Done.

research/src/pages/popup/popup.research.explainer.mdx
mfreed7: Done.

mfreed7:
Thank you both!

research/src/pages/popup/popup.research.explainer.mdx
scottaohara: that is the correct response.

PR #628, opened 2022-10-21
Fix feature detection code

mfreed7:
The `popup` attribute used to live on `Element`, but as part of the spec review, it was moved to `HTMLElement`. As a result, the feature detection code needs to change.


jonathantneal (approved):

andrico1234 (approved):

mfreed7:
Thanks!

Issue #415, opened 2021-11-02
[Popup] Should blur() be a light dismiss trigger?


mfreed7:
In some cases, it might make sense for blur (focus leaving the `<popup>`) to trigger light dismiss. For example, if the `<popup>` is the menulist for a `<selectmenu>`, when the list is opened, it is focused, and tabbing out feels like it should close. However, in most other cases, it seems like this behavior would likely be confusing. In particular, because the `<popup>` is not focused by default (without `autofocus` or `delegatesfocus`), it feels especially weird that tabbing over to `<popup>` and then leaving it should close it. But it came up in a discussion, so I thought I should raise it here. Perhaps it should be a light-dismiss trigger only if the `<popup>` is auto-focused?

scottaohara:
This would fundamentally depend on what the popup represents though. Eg the role that is exposed.

melanierichards:
Not wishing to detract from the semantic mapping discussion, I wanted to point to this piece of the initial explainer:

> An opened popup will have “light dismiss” behavior, meaning that it will be automatically hidden when:
> 
> The user hits the escape key,
> The layout of the popup or its anchor element is changed.
> **Focus moves outside of the popup (and its invoking and anchor elements, if applicable).**

The carveout for invoking/anchor elements was meant to solve for things like comboboxes, where you wouldn't want light dismissal of the popup when focus is in the text field. There's something attractive to your line of thinking on `autofocus`/`delegatesfocus`, but just wanted to add that there are cases where you wouldn't autofocus the popup but would expect it to dismiss on blur.

jonathantneal:
Hello 👋 

This item is labelled for discussion in the next teleconference. @melanierichards, @mfreed7, as editors of the proposal (and @mfreed7 as the creator of this issue), would you be available and interested in discussing this in the teleconference for additional feedback or consensus?

If so, @melanierichards, I believe you mentioned you would not be available tomorrow. Would next week work best for you both?

dandclark:
There's another special case for this in https://github.com/openui/open-ui/issues/486, regarding what should happen when the popup is open and the user clicks the button that has `triggerpopup`.

In the `<selectmenu>` prototype, this closes the popup and then immediately reopens it, but we probably want to do something better. Either closing the popup (and keeping it closed), or leaving the popup open both seem like reasonable choices.

scottaohara:
Closing would be my vote, and would be useful for touch screens / people using screen readers on touch screens.

mfreed7:
[...]

PR #630, opened 2022-10-31
Remove `blur` as a light dismiss trigger

mfreed7:
Per the [resolution](https://github.com/openui/open-ui/issues/415#issuecomment-1285995113), changing focus is no longer a light dismiss trigger.

### DIFF
--- a/research/src/pages/popup/popup.research.explainer.mdx
+++ b/research/src/pages/popup/popup.research.explainer.mdx
@@ -523,6 +523,7 @@ As described in this section, the popover types (`auto` and `manual`) each have 
 * Popover (**`popover=auto`**)
     * When opened, force-closes other `popover=auto`s, except for [ancestor popovers](#nearest-open-ancestral-popover).
     * It would generally be expected that a popover of this type would either receive focus, or a descendant element would receive focus when invoked.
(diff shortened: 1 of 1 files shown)


jonathantneal (approved):

andrico1234 (approved):

mfreed7:
Thanks!

research/src/pages/popup/popup.research.explainer.mdx
jandrade, line 1: @mfreed7 I'm not sure if I should tag you at this point(please let me know if I should tag somebody else, and I will):

By looking at the current spec, in the `Automatic Dismiss Behavior` section, the blur behavior is still included in the "Light dismiss" bullet point. Is it correct to say that this should be removed from there as well?

Thanks!

Screenshot 2024-01-08 at 12 31 25 PM

> **Light Dismiss.** User action such as clicking outside the element, hitting Escape, **or causing keyboard focus to leave the element** should all cause a displayed popover to be hidden.

mfreed7: Thanks for pointing that out. I've put up https://github.com/openui/open-ui/pull/975 to fix it. Though note that this is just the explainer, and the "spec" is actually correct and here: https://html.spec.whatwg.org/multipage/popover.html#popover-light-dismiss

Issue #548, opened 2022-06-11
[selectmenu] Method to change arrow icon?


hidde:
In our design system / authoring tool, we have a selectmenu-like component that uses a custom arrow:

<img width="168" alt="dropdown with as its icon a double arrow, one pointing up one pointing down" src="https://user-images.githubusercontent.com/178782/173187625-88765de1-b97f-4813-a904-4e25e286ebd9.png">

When I tried to recreate it with `<selectmenu>`, I got the default `<selectmenu>` arrow.

<img width="97" alt="select with arrow pointing down" src="https://user-images.githubusercontent.com/178782/173187694-0564be70-58fc-4b8e-b8e7-603c90ed9010.png">

I was able to insert my own arrow by creating a custom button (see linked Codepen below), but it would be nice to do it with the default button.

Could we add a way to change the arrow icon without replacing the entire button (maybe with `::marker`?)?

Related Codepens:
- [Example with my own custom button](https://codepen.io/hidde/pen/XWZyNqg?editors=1100) 
- [Example with just a standard selectmenu and options](https://codepen.io/hidde/pen/JjpeKQP?editors=1100)

captainbrosset:
I agree that this would be useful. I wrote several demos with the selectmenu component, and for some of them I only wanted a custom icon. Having to replace the entire button part for this seems a bit much.

chrisdholt:
We added a slot for the icon as well with FAST's select as well for this common use case. I'd love to see either `::marker` or a slot unique to this to support the simple scenario.

mfreed7:
I’m going to miss the meeting today, but I wanted to register my support for this proposal. It would seem to make the most sense to just add a slot for the icon, akin to the other two slots we already have.

gregwhitworth:
For ease of finding it in the minutes, the resolution was:

> Resolved: <selectmenu>'s shadow DOM has a <slot> around the button's default arrow icon so that the icon can be replaced by authors without replacing the entire button. The icon will also be exposed as a CSS `::part()`.

dandclark:
I'm going to close this since we had a clear resolution. I filed https://bugs.chromium.org/p/chromium/issues/detail?id=1357866 to update the Chromium prototype.

dandclark:
Reopening because I just realized we never decided what the name of the slot should be. Maybe `button-icon`? Or `dropdown-icon`, `arrow-icon`, just `icon`?

chrisdholt:
[...]

PR #640, opened 2022-11-17
Add `<marker>` part to anatomy and lists of parts

hidde:
This adds a `marker` part to the explainer and editor's draft as per our resolution to add a marker part https://github.com/openui/open-ui/issues/548#issuecomment-1262650219.

Also fixed the color contrast of the parts

screenshot of changes described above, with marker part added just after currently selected value part and the marker described as ‘something that indicates this button opens a listbox, e.g. an icon.’

### DIFF
--- a/research/src/components/anatomy.css
+++ b/research/src/components/anatomy.css
@@ -23,6 +23,10 @@
 
 /* Anatomy boxes */
 
+.component-anatomy {
+  font-weight: bold;
+}
+
 .component-anatomy .host,
 .component-anatomy .slot,
 .component-anatomy .part,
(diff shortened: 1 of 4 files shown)


andrico1234:
Thanks @hidde ,

Happy to approve, but it looks like we're having some issues with Netlify at the moment, so not sure if we want to hold off merging until we get it resolved

hidde:
Would be good to get input from at least 1-2 others maybe to see if the way I added it makes senee

(For selfish reasons, I would love to see it merged asap so I can refer to it in slides / posts, but no rush :-))

Issue #650, opened 2022-12-19
Normalize css vs popover


scottaohara:
Not really sure anything should be done about this here, but wanted to point out that while making codepen examples for popover, i have had to turn off the inclusion of normalize.css due that style sheet's use of `display: block` on HTML elements like `menu`, `article`, and such. 

May be worth a mention in future communication about the attribute that some of these sheets that were made for the purpose of resetting / normalizing HTML elements may actually cause issues with using new web features that these style sheets obviously could never have accounted for.

gregwhitworth:
@scottaohara thanks - yeah I'm not entirely sure what to do with this issue outside of possibly raising it as a concern in the CG so that folks giving talks, etc can make sure to call this out. What specific issues were you hitting?

scottaohara:
@gregwhitworth that the popover was shown by default because normalize was adding display block, per the old 'add display block to all the new html5 elements', to elements with the popover attribute.

mfreed7:
This is a great point! I wonder if someone (OP, hint hint) would be willing to file an issue with the normalize.css folks about this? I guess this applies to any "reset" stylesheet framework, of which there are several.

@jh3y I wonder if you have some visibility into the folks behind those and could reach out? Or at least mention this in a post somewhere?

jh3y:
> This is a great point! I wonder if someone (OP, hint hint) would be willing to file an issue with the normalize.css folks about this? I guess this applies to any "reset" stylesheet framework, of which there are several.
> 
> @jh3y I wonder if you have some visibility into the folks behind those and could reach out? Or at least mention this in a post somewhere?

This is only in places where `popover` isn't supported though, right? Because we have:

```css
[popover]:closed:not(dialog[open]) { display: none; }
```

In the user agent stylesheet which takes precedence over the one before it that would style `display: block`. It's definitely something to be mindful of if there are 3rd party stylesheets that might override this. I have this with `open-props` in some of the demos where I need to do something like:

```css
:where([popover]) {
  margin: auto;
  border-width: initial;
  border-style: solid;
}
```

To reset myself. Can see that in this basic demo: https://codepen.io/web-dot-dev/pen/qBowZXP

jh3y:
Hmm. I have some stylesheets where I've had to put a reset or something in place. 

Perhaps this is something to make more prominent in the article/explainer? I didn't see this behavior with `article`. But, I tried `menu` and sure enough it applied `display: flex`. And ideally, you'd want something like:

```css
[...]

PR #654, opened 2023-01-20
Update popover.research.explainer.mdx

jh3y:
- Adds a note about being mindful of stylesheets that you haven't authored directly in case they interfere with `display` values (Mentioned in #650)
- Removes mention of `defaultopen`
- Changes events section to detail `beforetoggle` and how it works

### DIFF
--- a/research/src/pages/popup/popover.research.explainer.mdx
+++ b/research/src/pages/popup/popover.research.explainer.mdx
@@ -284,7 +284,7 @@ The above rules mean that a popover, when not "shown", has `display:none` applie
 }
 ```
 
-Be mindful when using other stylesheets that you haven't authored. These may set the `display` value on your popover elements. For example, if you use a `menu` element as a `popover` and you use [normalize.css](https://necolas.github.io/normalize.css/). In this case, normalize.css has a rule that sets `display: flex` on `menu` elements.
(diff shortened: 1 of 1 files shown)


research/src/pages/popup/popover.research.explainer.mdx
```diff
@@ -284,6 +284,7 @@ The above rules mean that a popover, when not "shown", has `display:none` applie
 }
 ```
 
+Be mindful when using other stylesheets that you haven't authored. These may set the `display` value on your popover elements. For example, if you use a `menu` element as a `popover` and you use [normalize.css](https://necolas.github.io/normalize.css/). In this case, normalize.css has a rule that sets `display: flex` on `menu` elements.
```
scottaohara, line 287: did some more research on this and the latest version of normalize css doesn't have this styling issue.  However, CodePen is using a ridiculously old version (5 vs latest 8.x), so just modifying this text to indicate that without specifically calling them out (well... outside of this clarifying comment). 

```suggestion
Be mindful when using other stylesheets that you haven't authored. These may set the `display` value on your popover elements. For example, if you use a `menu` element as a `popover` and you use earlier versions of [normalize.css](https://necolas.github.io/normalize.css/). In this case, normalize.css has a rule that sets `display: flex` on `menu` elements.
```

jh3y: Ahh good research Scott 👏 
I'd assumed CodePen was using the latest. Thanks for this!

scottaohara: as would i, but lol assumptions :)

andrico1234 (approved):

PR #656, opened 2023-01-30
Add selectmenu explainer

josepharhar:
This is a basic explainer for the selectmenu element. There may be some features I left out but I figure that we should at least get it started.


josepharhar:
Nevermind, I was just made aware of https://open-ui.org/prototypes/selectmenu which should be reconciled with this new doc first

PR #686, opened 2023-03-21
[Popover explainer] Add article "Open UI and the Popover API" in the link section

brechtDR:
Adding an extra article I created about Open UI and the Popover API

link: https://utilitybend.com/blog/open-ui-and-the-popover-api

### DIFF
--- a/site/src/pages/components/popover.research.explainer.mdx
+++ b/site/src/pages/components/popover.research.explainer.mdx
@@ -208,7 +208,7 @@ There are several conditions that will cause `showPopover()`, `hidePopover()`, a
 4. Calling any of the three methods on a valid popover that is not connected to a document. This will throw an `InvalidStateError` `DOMException`.
 
 {/* Removing until it's back in action */}
-{/* ### Page Load Trigger
+{/\* ### Page Load Trigger
 
 As mentioned above, a `<div popover>` will be hidden by default. If it is desired that the popover should be shown automatically upon page load, the `defaultopen` attribute can be applied:
 
@@ -228,9 +228,10 @@ Note that more than one `manual` popover can use `defaultopen` and all such popo
 The `defaultopen` content attribute can also be accessed via IDL:
 
 ```javascript
-myDiv.defaultOpen = true;
+myDiv.defaultOpen = true
 ```
-*/}
+
+\*/}
 
 ### CSS Pseudo Class
 
@@ -742,3 +743,4 @@ https://github.com/openui/open-ui/issues?q=is%3Aissue+is%3Aopen+label%3Apopover+
 
 - [Pop-ups: They're making a resurgence! – developer.chrome.com](https://developer.chrome.com/blog/pop-ups-theyre-making-a-resurgence/)
 - [Dialogs, modality and popovers seem similar. How are they different? – hidde.blog](https://hidde.blog/dialog-modal-popover-differences/)
+- [Open UI and the Popover API – utilitybend.com](https://utilitybend.com/blog/open-ui-and-the-popover-api)
--- a/site/src/pages/components/popover.research.explainer.mdx
+++ b/site/src/pages/components/popover.research.explainer.mdx
@@ -208,7 +208,7 @@ There are several conditions that will cause `showPopover()`, `hidePopover()`, a
 4. Calling any of the three methods on a valid popover that is not connected to a document. This will throw an `InvalidStateError` `DOMException`.
 
 {/* Removing until it's back in action */}
-{/\* ### Page Load Trigger
+{/* ### Page Load Trigger
 
 As mentioned above, a `<div popover>` will be hidden by default. If it is desired that the popover should be shown automatically upon page load, the `defaultopen` attribute can be applied:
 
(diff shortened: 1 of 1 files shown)


brechtDR:
@scottaohara Can  you be a bit more precise. 

The toggle tips were updated in the demo, they seem to work fine with voice over and navigation for now (you were right about the navigation, could do better). I do want to enhance them further, but as this is more of a entry-level tech demo, i want to keep things simple and not clutter it too much with JS. However, I don't mind adding some aria-labels to them or such.

Any more suggestions would be welcome.. I know the tooltip role isn't ideal in this case

For the modal... I think it's pretty clear. I obviously say that this isn't the best practice as you quoted in the comment. And it's a great demonstration of how the basics of the popover api work.

scottaohara:
Thanks for updating the toggle tip demos. the popover positioning is better now. re: a role to use. i prefer "note" for the container element for things like that.  it semantically conveys it's additional information about the associated subject, and has better support for being announced as a container element than tooltip does.  but that's just what has worked for me, and situationally other container element roles may be more appropriate.  but referring to these as something different than tooltip might be good for preventing author confusion... especially since the "tooltip" like "hint" popover was specifically punted from v1. 

Re: more precise with the popover dialog content -  for instance both of your "i'm a popup that wants to be a modal" examples don't behave as modals.  Each visually looks like a modal dialog, since they obscure the screen visually, but neither act that way as one can easily escape them using keyboard alone or a screen reader.

You have text in your post that states:
>If we would like to create a modal where we don’t want a light dismiss, we can easily create this by doing the the following:

followed by a code example which does not have `aria-modal=true` to help trap a screen reader's navigation to within the dialog popover.  And if this was an example used in context of a website with other interactive elements, those elements would need a tabindex=-1 or use the inert attribute to ensure the interactive elements and any other static content was actually inaccessible to screen readers or other AT while the 'modal' was open.

Stuff like this is why the post is coming across as sending mixed messages.  The code snippets/examples seem to tell the story "just do this and you can make a popover modal" but then the post also states that popover won't contain focus - and the codepen examples also don't do this, so these are more effective as how these _don't_ create 'modal' dialogs, and should be more overtly called out as not making the cut.

Saying that one can use javascript or the dialog element gives a false equivalence between the two without further nuance.  While one could use javascript to help focus trap within the popover, JavaScript alone would do nothing to mitigate screen readers being able to leave the popover - as Tab key is merely one way for these users to navigate a web page.  Hence my mention of aria-modal or the inert attribute being necessary to fully create a modal dialog with a popover... and why more strongly advocating for the dialog and its showModal method would be preferable.

i don't typically log into discord, but i have in case you want to discuss this further outside of back and fourths in this issue.

brechtDR:
Things changed: 
- changed the tooltips to have the note role in the article and in the code
- removed the confusion of the modal by just not mentioning it as a modal anymore and just as a popover
- edited the text to be more precise that JS is only for the focus trapping, changed"or" to "and"
- removed the "wants to be a modal" in the demo, although it was already using a dialog element. also added the aria-role for people just stumbling on the demo instead of the article

The reason I kept it simple was to prevent turning this article to "how to create a modal with open-ui with good a11y". And keep the original focus of the article to "this is Open UI and the popover API and this is what it can do". 

Reason i kept the note for the modal is mostly because of the focus trapping and especially to emphasise that this popover is not a modal perse. Nothing is holding me back to write an article later on explaining how to create a modal with focustrapping and the dialog element, using Open UI to trigger it. But this "intro" article is not the place. Which doesn't mean i have no understanding for the concerns you have about this. I do believe that these things will need to be educated, in good time and more in depth and to the point.

I don't really agree about "just the way i show the original demo is what makes people think that this is a modal". If you strip all the CSS it basically looks the same but without colours and rounded corners. 

I hope by creating this article that I can get more people interested in Open UI and they can start creating demos themselves or search more information on the subject, at least, that's the main goal of this article.

brechtDR:
@scottaohara Thank you for all the insights... 

> you will need to, for example, implement your own focus trapping with JavaScript, and use additional ARIA attributes. Alternatively, one could simply use the <dialog> element with the .showModal() method.

I really was thinking hard on how I could change that sentence without deriving too much from the essence. You just nailed it by just rephrasing and being more clear + using a bit of stronger language.
I'm not a native English speaker and that seemed to be a hard one for me. I still struggle from time to time and it's kind of the reason why I started writing in the first place.

So really, thank you for that.

> we be absolutely clear that the demos meet the necessary accessibility requirements, or they are clearly stated as being demonstrations of the popover features, but not necessarily meeting accessibility requirements.

I totally agree with this was all very insightful and helpful feedback 💯

scottaohara:
really appreciate you sticking with me on this @brechtDR updates look good to me (there's an instance of "you you" in the updated content, just fyi).

when i write my post on popover, i'll be sure to link to this one as well.

scottaohara (approved):

brechtDR:
@scottaohara Ah 😅 small oversight. Fixed and you're welcome, it was my pleasure to improve the article.

gregwhitworth:
First off, thank you @scottaohara for the thorough review and ensuring that any examples we're landing (and people may copy and use) are accessible; and thank you @brechtDR for taking the feedback :)

Issue #683, opened 2023-03-17
Remove the Select element from proposals


gregwhitworth:
Now that `selectmenu` explainer has landed via #659 we should have the old one deleted. While we could just simply hide it from the menu I don't see the value in keeping it around since we're not actively trying to move this version forward.

josepharhar:
There are two old select pages I can see:
1. https://open-ui.org/components/select/
2. https://open-ui.org/components/select.research/

Which one(s) of them do you want to delete? Should the path(s) be kept in use to redirect to the new explainer?

josepharhar:
Ok I think I'm going to delete the first one and keep the second "research" one

PR #690, opened 2023-03-23
Delete old select explainer

josepharhar:
Fixes https://github.com/openui/open-ui/issues/683


gregwhitworth:
@josepharhar I'll land this one but can you please join the Open UI CG and ensure your github account is attached to ensure that the IP passes. This one is non-substantiative so approving.

PR #691, opened 2023-03-24
feat: rudimentary contributor bubble [Resolve #679]

### LINKED ISSUE #679: Add contributor bubbles to the Open UI site
The Astro site has an example for adding this to a site

Here's the link to the component they've made: https://github.com/withastro/docs/blob/main/src/components/ContributorList.astro
here's the page in the site: https://docs.astro.build/en/getting-started/
here's the usage of the component: https://github.com/withastro/docs/blob/main/src/content/docs/en/getting-started.mdx?plain=1#L114

jh3y:
Hey @andrico1234 / @gregwhitworth 

This is a "rough" cut of getting the contributor bubbles into the front page. The issue is that the request will get rate limited for contributors if we hit it too often. The options are:

- Stick with it as-is as we don't build the site often and shouldn't get rate limited
- Store the contributors into a synced file that we can run manually `node src/data/sync.js`. Not ideal, but works.
- Set up Github API tokens and use them in the component. This requires saving them as secrets in Netlify, etc. I don't have the keys to those gates so would need someone to set them up or send me them if we were to go down that route.

┬┴┬┴┤•ᴥ•ʔ├┬┴┬┴


gregwhitworth (approved):

gregwhitworth:
I agree with keeping it as-is for now and thank you for doing this! This has long been a desire of mine to land so thank you!

PR #707, opened 2023-03-31
Add table of contents to selectmenu explainer

josepharhar:
This patch adds a table of contents and also:
- shuffles around some of the headings
- replaces "popup" with "popover"

### DIFF
--- a/site/src/pages/components/selectmenu.mdx
+++ b/site/src/pages/components/selectmenu.mdx
@@ -7,17 +7,32 @@ layout: ../../layouts/ComponentLayout.astro
 import SelectAnatomy from '../../components/select-anatomy'
 import Image from '../../components/image.astro'
 
-## Background
+## Table of Contents
+
+- [Background](#background)
+- [Testing out the selectmenu element](#testing-out-the-selectmenu-element)
+- [Anatomy of the selectmenu element](#anatomy-of-the-selectmenu-element)
+- [Styling selectmenu](#styling-selectmenu)
+  - [Styling parts of the control](#styling-parts-of-the-control)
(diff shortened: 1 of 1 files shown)


site/src/pages/components/selectmenu.mdx
```diff
@@ -188,7 +207,7 @@ You can replace the default listbox part in a similar fashion:
 </selectmenu>
 ```
 
-The `popup` attribute used in this example's `<div popup>` is from the proposal at [Popup API (Explainer)](https://open-ui.org/components/popup.research.explainer). The element with `behavior="listbox"` is required to be a `<div popup=popup>`. Applying `behavior="listbox"` tells the browser to apply select menu listbox behavior to that element: it will be opened when the `<selectmenu>`'s button is clicked, and the user can select `<option>`s inside it with mouse, arrow keys, and touch.
+The `popover` attribute used in this example's `<div popover>` is from the proposal at [Popover API (Explainer)](https://open-ui.org/components/popover.research.explainer). The element with `behavior="listbox"` is required to be a `<div popover="auto">`. Applying `behavior="listbox"` tells the browser to apply select menu listbox behavior to that element: it will be opened when the `<selectmenu>`'s button is clicked, and the user can select `<option>`s inside it with mouse, arrow keys, and touch.
```
hidde, line 210: would it make sense to, instead of the explainer, refer to the Popover API and link to to https://html.spec.whatwg.org/dev/popover.html#the-popover-attribute instead here?

andrico1234 (approved):
Thanks @josepharhar ,

I've approved, so feel free to merge when you're ready.

PR #714, opened 2023-04-07
[select] Use the new ::behavior selector and clarify that ::part doesn't work

### LINKED ISSUE #645: [selectmenu] Add CSS selector to consistently select selectmenu parts
There's currently no way to select a `<selectmenu>` part, especially if it's provided externally.

For example, if we want to select the listbox part:
* `selectmenu::part(listbox)` only works if we are using the listbox in the UA shadow DOM
* `selectmenu [behavior=listbox]` should work for most cases when the listbox is author-provided
* However, there doesn't seem to be a way if, e.g., the `selectmenu` is in a shadow DOM, and the listbox is passed through from an outer tree scope in a way similar to #565 

While this doesn't seem to be a big issue from the author's perspective (since the author should always know how to select the part they provided), it blocks the browser from adding a proper UA style sheet to position the listbox (as in this [Chromium patch](https://chromium-review.googlesource.com/c/chromium/src/+/4082046)). 

Should we introduce new CSS pseudo element selectors that can select the parts, regardless of where they are (whether in UA shadow DOM or not)? For example, `selectmenu::behavior(listbox)`.

This may be of interest for developers as well.

xiaochengh:
Fixes #645

### DIFF
--- a/site/src/pages/components/selectmenu.mdx
+++ b/site/src/pages/components/selectmenu.mdx
@@ -52,18 +52,18 @@ Selectmenu provides a variety of tools to help with styling, including pseudo-se
 
 ### Styling parts of the control
 
-One way to style the control to match your requirements is to use the CSS `::part()` pseudo-element to select the different parts within the control's anatomy that you wish to style.
+One way to style the control to match your requirements is to use the CSS `::behavior()` pseudo-element, newly introduced for the selectmenu element, to select the different parts within the control's anatomy that you wish to style.
 
 ```html
 <style>
-  .my-select-menu::part(button) {
+  .my-select-menu::behavior(button) {
     color: white;
     background-color: #f00;
     padding: 5px;
     border-radius: 5px;
   }
 
-  .my-select-menu::part(listbox) {
+  .my-select-menu::behavior(listbox) {
     padding: 10px;
     margin-top: 5px;
     border: 1px solid red;
@@ -84,7 +84,9 @@ The above example results in the following style:
   alt="The rendering of a selectmenu control with the above HTML and CSS code"
 />
 
-`::part()` can be used to style the `button`, `selected-value`, and `listbox` parts of the control.
+`::behavior()` can be used to style the `button`, `selected-value`, and `listbox` parts of the control.
+
+Note that the `::part()` pseudo-element selector only matches elements in the user-agent shadow tree, and hence, fails to match any author-provided part (as in [slotting your own content](#slotting-your-own-content)).
 
 ### `:open` and `:closed` pseudo-selectors
 


andrico1234:
@gregwhitworth @josepharhar , are you able to review the content for this change? :)

site/src/pages/components/selectmenu.mdx
```diff
@@ -84,7 +84,9 @@ The above example results in the following style:
   alt="The rendering of a selectmenu control with the above HTML and CSS code"
 />
 
-`::part()` can be used to style the `button`, `selected-value`, and `listbox` parts of the control.
+`::behavior()` can be used to style the `button`, `selected-value`, and `listbox` parts of the control.
+
+Note that the `::part()` pseudo-element selector only matches elements in the user-agent shadow tree, and hence, fails to match any author-provided part (as in [slotting your own content](#slotting-your-own-content)).
```
gregwhitworth, line 89: can you please wrap this with a `div` with a class of `.note`

site/src/pages/components/selectmenu.mdx
```diff
@@ -52,18 +52,18 @@ Selectmenu provides a variety of tools to help with styling, including pseudo-se
 
 ### Styling parts of the control
 
-One way to style the control to match your requirements is to use the CSS `::part()` pseudo-element to select the different parts within the control's anatomy that you wish to style.
+One way to style the control to match your requirements is to use the CSS `::behavior()` pseudo-element, newly introduced for the selectmenu element, to select the different parts within the control's anatomy that you wish to style.
```
gregwhitworth, line 55: We should link out to the definition of behavior. Do we have one in CSSWG yet? If not we should probably expound on this.

site/src/pages/components/selectmenu.mdx
gregwhitworth: Also see my above comment on whether `::behavior` is defined somewhere as this should be covered in that definition; not in `selectmenu` itself.

gregwhitworth (changes requested):

site/src/pages/components/selectmenu.mdx
josepharhar: I don't know of a CSSWG issue proposing `::behavior` yet.
If we remove the "newly introduced for the selectmenu element" part is that good enough to merge this now?
I'm hesitant to open a CSSWG issue proposing this just yet since domenic's general feedback could change how this works. I haven't even opened an HTML issue for selectmenu yet.

xiaochengh: This is not in CSSWG yet.

> I'm hesitant to open a CSSWG issue proposing this just yet since domenic's general feedback could change how this works. I haven't even opened an HTML issue for selectmenu yet.

I feel the same. Should we create a draft spec of `::behavior` somewhere at open-ui?

gregwhitworth:
@josepharhar @xiaochengh I'm presuming we should abandon this PR?

xiaochengh:
Yeah, it doesn't seem relevant with the element model.

PR #724, opened 2023-04-21
Start an explainer for exclusive accordions.

dbaron:


dbaron:
This is still incomplete, but I think there's probably enough here to be ready for review.  See [preview](https://deploy-preview-724--open-ui.netlify.app/components/accordion.explainer/).

site/src/pages/components/accordion.explainer.mdx
```diff
@@ -0,0 +1,157 @@
...
+The existing HTML
+[`details` element](https://html.spec.whatwg.org/multipage/C#the-details-element)
+represents a disclosure widget.
+A sequence of `details` elements can be used for a non-exclusive accordion.
+However, there is not a current web platform feature
+for exclusive accordions.
+
+It is worth noting that the HTML `details` element
+is not currently as styleable as it needs to be to meet many use cases.
+Improving styleability of `details` will be addressed in a separate proposal,
```
mfreed7, line 52: It'd be helpful to have a "For example, you can't style ..." sentence right here.

site/src/pages/components/accordion.explainer.mdx
```diff
@@ -0,0 +1,157 @@
...
+intended to address both better ability to style/replace the disclosure triangle,
+and (maybe separately) to fit the parts of a `details`
+into different types of layout (such as grid).
+
+# Developer Demand for Exclusive Accordion
+
+Design systems that offer accordion widgets vary
+as to whether those widgets are exclusive or not.
+Based on the systems documented in the
+[accordion research](/components/accordion.research),
```
mfreed7, line 62: The current accordion research doesn't show much...

site/src/pages/components/accordion.explainer.mdx
```diff
@@ -0,0 +1,157 @@
...
+that have the same value for this attribute
+would form an exclusive accordion.
+The syntax of this attribute and the rules for matching its values would match
+[the `name` attribute](https://html.spec.whatwg.org/multipage/C#naming-form-controls:-the-name-attribute)
+and its use in defining
+[radio button groups](https://html.spec.whatwg.org/multipage/C#radio-button-group).
+
+When these attributes are used to form a group, it means that
+any user action or API that causes one of the `details` elements
+in the group to open causes all other elements in the group to close.
```
mfreed7, line 86: Doesn't it also mean the keybindings and perhaps some a11y attributes will change accordingly?

site/src/pages/components/accordion.explainer.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+## Open design questions
+
+### Name
+
+I suggest that this attribute might be called `group`.
```
mfreed7, line 92: Why not call it `name` to match the existing `name` attribute exactly? The explainer is ok as-is, I'm just curious why this wouldn't be the first choice.

site/src/pages/components/accordion.explainer.mdx
```diff
@@ -0,0 +1,157 @@
...
+* indicate that something happens when multiple elements use the same value, and
+* indicate that the attribute causes the grouped elements
+  to become exclusive (so that only one can be open).
+
+A few other possibilities for the name are `name` or `exclusive`.
+
+### Interaction with `open` attribute
+
+Should this grouping cause the insertion of a `details` element with
+the `open` attribute already set to have the `open` attribute removed
```
mfreed7, line 106: Ohhhh good question. Not sure how this one will end up, but I'm glad you thought of it early.

mfreed7 (approved):
LGTM with a few nits. Nice!

site/src/pages/components/accordion.explainer.mdx
dbaron: Added (with somewhat different wording).

site/src/pages/components/accordion.explainer.mdx
dbaron: Yes, it doesn't show much... though it does have one sentence saying which of the systems listed are exclusive or not, and it has links to demo pages for each of them.

(I didn't take the time to figure out how to create images... and it's also not clear how useful images would be for this question since they don't show the interaction behavior.)

site/src/pages/components/accordion.explainer.mdx
dbaron: Possibly.  I added this as a separate open question, and pointed out its relationship to the "is grouping useful for non-exclusive accordions" question.

site/src/pages/components/accordion.explainer.mdx
dbaron: I changed it to be a little more neutral, but list `name` first.  I also mentioned that matching existing features could be an advantage or a disadvantage.

PR #727, opened 2023-04-24
Add three more accordion components.

dbaron:
This creates new files for Material UI and for Tailwind Elements, and adds only the Accordion component to them.

It also drops the "Concepts" part of the accordion research page since it only has data from one system and thus isn't very useful.

### DIFF
--- a/site/src/sources/materialUI.json
+++ b/site/src/sources/materialUI.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "../schemas/design-system.schema.json",
+  "lastUpdated": "2023-04-24",
+  "name": "Material UI",
+  "description": "Material UI is a library of React UI components that implements Google's Material Design.",
+  "url": "https://mui.com/material-ui/getting-started/overview/",
+  "by": "MUI",
+  "components": [
+    {
+      "name": "Accordion",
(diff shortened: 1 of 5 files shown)


dbaron:
Previews of the affected pages:
https://deploy-preview-727--open-ui.netlify.app/components/accordion.research
https://deploy-preview-727--open-ui.netlify.app/design-systems
https://deploy-preview-727--open-ui.netlify.app/research/component-matrix

site/src/sources/materialUI.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "../schemas/design-system.schema.json",
+  "lastUpdated": "2023-04-24",
+  "name": "Material UI",
+  "description": "Material UI is a library of React UI components that implements Google's Material Design.",
+  "url": "https://mui.com/material-ui/getting-started/overview/",
+  "by": "MUI",
+  "components": [
+    {
+      "name": "Accordion",
```
mfreed7, line 10: Do we typically just import the components we're interested in? Or is there a way to import links to all of the Material and Tailwind components? I'm only saying this if it's easy. If each component takes work to figure out, then we can do it piecemeal like this.

PR #731, opened 2023-04-27
Update Exclusive Accordion explainer based on discussion in April 27 meeting.

dbaron:
Also, link to proposal for details styling.

### DIFF
--- a/site/src/pages/components/accordion.explainer.mdx
+++ b/site/src/pages/components/accordion.explainer.mdx
@@ -151,14 +151,23 @@ and insertion through other methods like setting `innerHTML`.)
 For many accordions, it is desirable for the print layout of the page
 to have all items opened.
 This is currently not possible to do from a style sheet,
-but it probably should be possible.
+but it probably should be possible,
+and it's possible that it should be the default.
 The presence of exclusive accordion semantics make this more difficult,
 because of the rule that only one item can be open.
+(However, some of the time it will make more sense to print what is
+on the screen rather than printing all of the content.
+It's also not clear to me how often distinguishing between these cases
+should be a developer decision or should be based on a user decision.)
 
 In theory this could be addressed with a model
 where details elements were implemented on top of developer-exposed CSS Toggles,
 and developers could disable the effects of those CSS Toggles when printing.
 However, this would be a relatively major change, worthy of its own explainer.
+
+It's also possible that developers could do this in Javascript,
+specifically for printing,
+but then it certainly wouldn't be a default and probably wouldn't be done often.
 
 We should find and consider alternative approaches for addressing this issue.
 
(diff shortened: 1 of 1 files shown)


site/src/pages/components/accordion.explainer.mdx
```diff
@@ -144,6 +146,24 @@ violate the exclusivity semantics?
...
 
+### Printing
+
+For many accordions, it is desirable for the print layout of the page
+to have all items opened.
+This is currently not possible to do from a style sheet,
+but it probably should be possible.
+The presence of exclusive accordion semantics make this more difficult,
+because of the rule that only one item can be open.
+
```
mfreed7, line 157: Worth a sentence or two also here that sometime's that's *not* what people want? I.e. sometimes (often?) when people hit print, they want to print what they're seeing, which would be just one open option. And perhaps that leads to this being a JS-driven thing during printing, where if the desire is to print "all", then the JS removes the `name` attribute and adds the `open` attribute to all `<details>`?

site/src/pages/components/accordion.explainer.mdx
```diff
@@ -154,6 +174,12 @@ If this is the case,
 it seems likely that such changes
 may also be desirable for non-exclusive accordions (see next question).
 
+Based on the [discussion on 2023 April 27](https://github.com/openui/open-ui/issues/725#issuecomment-1525924478),
+it seems like at least default keyboard behavior for accordions
+is undesirable because it leads to bad user experience
+particularly for users on small or highly-zoomed displays
+(where the content of a single item is less likely to fit within the display).
```
mfreed7, line 181: ...or perhaps it leads to the need for some other feature (within focusgroup?) that allows the sub-items to be accessibly scrolled also? I don't know.

site/src/pages/components/accordion.explainer.mdx
```diff
@@ -172,7 +198,9 @@ then that would favor a design that separates the ideas
 of connecting the `details` elements into a group
 and making the group exclusive.
 
-## Possibly-resolved design questions
+In the [discussion on 2023 April 27](https://github.com/openui/open-ui/issues/725#issuecomment-1525924478),
+no use cases for a platform feature for connecting a non-exclusive accordion
+were raised.
```
mfreed7, line 203: ...other than the potential keyboard interaction pattern issue raised above?

mfreed7 (approved):
LGTM with some comments. Take them or leave them.

site/src/pages/components/accordion.explainer.mdx
dbaron: I added a caveat pointing to the previous issue, although I think my impression of the discussion of that issue was perhaps of a firmer conclusion to it than your impression.

Issue #559, opened 2022-06-30
Panelset & tabs: defining separate goals for each pattern


smhigley:
@bkardell and I went through a number of examples of more traditional application tabs and examples of the proposed panelset pattern, and came up with a proposal to separate them into two different platform-level solutions with discrete purposes and use cases, outlined below.

---
Through a lot of discussion and looking at examples, as well as based on evidence in existing UI toolkits, we believe that there is a need for two separate solutions for something that many users would just call 'tabs'.  In this document we're attempting to explain the differences between them, what their uses are and offer some working terminology.

There may or may not be implementation detail overlap between these ideas, but their basic purpose, limits and likely semantics would differ. They probably benefit from work happening in parallel to coordinate similar APIs in areas like CSS pseudo-selector naming or open/close DOM attributes.

## Statement of purpose for each element
A brief description of the basic purpose and functionality of each proposed element.

### Application tabs
The primary purpose of application tabs is to allow users to easily switch between multiple possible panes of content that are specifically not intended to be viewed at the same time. As such there are functional differences, covered in more detail in the next section.  Possible examples include browser tabs, file tabs in a code editor, multiple views within an app such as map/list, or email/calendar/contacts. 

Some implementation challenges that a tabs element could solve for authors include:
- Semantics
- Keyboard interaction
- Close buttons
- Tablist overflow

### Panelsets
The primary purpose of panelsets is to provide a mechanism to reflow multiple sections of content between a tab-like one-at-a-time presentation and a linear, sequential presentation. So in essence, the title/content ordering for multiple sections would switch between these two options:

1. Title 1, Title 2, TItle 3
   Content #

2. Title 1, Content 1
   Title 2, Content 2
   Title 3, Content 3

Visually, the first option could be presented as tabs, a carousel, or anything with a similar content-switching interface.

The second option could be presented as accordions or simply linear content.

Some implementation challenges this element or elements could solve for authors include:
- Reordering of content on the page when reflowing between two presentations
- Automatically triggering a change in presentation based on container size
- Changes in interaction between switching out a content pane vs. expanding/collapsing multiple panes
- Potentially changes in semantics between the two presentations (depending on whether we can find a set of semantics that would work for each case)

## Differences between Tabs and panelsets
- When printing, tabs would only print the currently shown tab content, while panelsets may have the ability to print all content, regardless of the current onscreen presentation
- Search in page and similar functionality: panelsets would search all content, tabs would only search the currently open content
[...]

PR #746, opened 2023-05-08
[exclusive-accordion] Add sections on existing details issues and spicy sections

dbaron:


dbaron:
See [preview](https://deploy-preview-746--open-ui.netlify.app/components/accordion.explainer/).

site/src/pages/components/accordion.explainer.mdx
```diff
@@ -289,3 +317,21 @@ while intersecting,
 are different enough
```
mfreed7, line 317: Is it worth adding a link to https://github.com/openui/open-ui/issues/559 above?

I also don't quite understand the difference between "panelset" and "spicy sections". They both seem like a thing that goes between accordions and tabs. I'm not saying you need to change anything about the explainer, per se, just asking if you agree. If so, I suppose you could add a sentence along those lines. Or combine the two sections together. Or leave them as-is.

mfreed7 (approved):
LGTM, just one rambling comment. Take it or leave it.

bkardell:
Yeah, sorry I didn't have enough time to comment on this but spicy-sections _is_ panel-set (or you might see panselset).  I made a comment in the commit.

bkardell:
probably close enough, yeah. @mirisuzanne wdyt?

mirisuzanne:
Yeah, that makes sense to me.

PR #721, opened 2023-04-12
Add (stub-ish) research pages for accordion and disclosure.

dbaron:
This adds stub-ish research pages for the accordion and disclosure components.

PR #750, opened 2023-05-12
Update KoliBri / Public UI Spec to 1.5

deleonio:
We release a new version with more components and features. Thank you for merging.


site/src/sources/kolibri.json
```diff
@@ -23,26 +23,13 @@
       ]
     },
     {
-      "name": "Accordion",
-      "url": "https://public-ui.github.io/docs/components/accordion",
-      "definition": "Die **Accordion**-Komponente ist ein Aufklapp-Menü. Klickt man auf den Kopfbereich, bestehend aus Icon und Überschrift, öffnet sich ein Textfeld mit zusätzlichen Informationen. Somit ist es ein interaktives Navigationselement, welches dazu dient, umfangreiche Inhalte platzsparend darzustellen.\n\nAccordions kommen immer dann zum Einsatz, wenn einem thematischen Oberbegriff zugeordnete Inhalte angezeigt oder verborgen werden sollen. Sie erlauben umfangreichere Detailinformationen zu einem Oberbegriff, als es aus Gründen der Übersichtlichkeit eigentlich sinnvoll wäre. Sie überlassen es den Besucher:innen selbst, ob sie sich diese Informationen anzeigen lassen möchten.",
+      "name": "accordion",
```
dbaron, line 26: So changing this from "Accordion" to "accordion" means that it no longer appears in the list of widget names by design system; compare these pages:
https://open-ui.org/components/accordion.research/
https://deploy-preview-750--open-ui.netlify.app/components/accordion.research/

This could be solved by adding an 'openUIName' field saying 'Accordion', but then you'd have something that looks like this in the Names section:
https://open-ui.org/components/disclosure.research/
except with one line for "Accordion" and one line for "accordion".

(I think that's why I changed "Accordion" to uppercase at some point in the past, because I was working on the page.)

I think it would be better to uppercase all of the "name" fields here.  (And a later PR might also add appropriate `openUIName` fields to make the widgets show up on the relevant research pages when that isn't sufficient.)

site/src/sources/kolibri.json
```diff
@@ -23,26 +23,13 @@
       ]
     },
     {
-      "name": "Accordion",
-      "url": "https://public-ui.github.io/docs/components/accordion",
```
dbaron, line 27: I probably manually added this `url` field as well (and also for Details) -- but it would be good to keep it, and probably also to generate similar `url` fields for the other widgets so that they show up properly in the generated pages.

dbaron (changes requested):
Thanks for updating this.

A few comments, mainly related to the fact that this is overwriting manual changes that I made in #721 in order to make this data actually show up on two of the research pages.  I think at least it would be good to not undo those changes (see more details below), but potentially you could make similar changes elsewhere in the file (which seemed out of scope for what I was doing in #721, although I did start to do it and then realize it would be a decent amount of work).

I suspect you have a tool that manually generates this... I'd be interested to know what.

site/src/sources/kolibri.json
deleonio: Thank you @dbaron for your hint. I have added all urls.

site/src/sources/kolibri.json
deleonio: I have add the `openUiName` to all our meta informations in the `kolibri.json`. I hope it was the right way. Please let me know, if I should do something else.

deleonio:
Hi @dbaron,

yes, we have a tiny script for build the json file for OpenUi: https://github.com/public-ui/kolibri/blob/main/packages/components/scripts/open-ui.js

dbaron:
I'm hoping you'll be able to do the update I mentioned above, and then we can merge this.  (I think it's certainly easier for you to do, given the script that you have in the other repository!)

deleonio:
> I'm hoping you'll be able to do the update I mentioned above, and then we can merge this. (I think it's certainly easier for you to do, given the script that you have in the other repository!)

👍 Done.

PR #759, opened 2023-05-26
Panelset isn't panelset - update accordion.explainer.mdx

bkardell:
Sorry I'm not sure how I missed what was actually _in_ 748 when merged, but there is some real confusion there that is probably easier to correct in a PR and a comment.

This PR _removes_ references to the old 2015 proposal which was _also_ called panelset. I'm sorry 😆 
While there are some surface similarities in that what was defined was a general structure on which several views could be laid, they are in almost every practical way very different things and thus most of the content/critique in there is not talking about the actual work this group did. That work was preliminary referred to as "spicy-sections" because: a) it followed hixie's original idea [in the HTML spec of just putting a marker element around some sections](https://whatwg.org/specs/web-apps/2006-01-01/#switch) b) we didn't want people to take it too seriously yet or get into bikeshedding names.

Eventually we did bikeshed names, and (not because I advocated it, I didn't) [people chose 'panelset' again](https://github.com/tabvengers/spicy-sections/issues/49)

Sorry.  I hope this PR helps, I'm happy to help answer questions or fill it in or review a different request instead.

### DIFF
--- a/site/src/pages/components/accordion.explainer.mdx
+++ b/site/src/pages/components/accordion.explainer.mdx
@@ -341,46 +341,22 @@ particularly to understand
 what integration with CSS and Javascript it would have,
 and how that integration would affect its use.
 
-## Panelset
+## Panelset (aka "Spicy-Sections")
 
-The [Panels and Panel Sets proposal](https://bkardell.com/common-panel/index.src.html)
-(also see [introductory article](https://www.oreilly.com/content/panels-and-panel-sets/))
-is significantly more general,
-and proposes to cover accordions, carousels, tabs, and decks.
-(It is less general than CSS Toggles,
-but more general than all the others mentioned here.)
-It was also previously known as spicy sections;
-see the [spicy-sections repository](https://github.com/tabvengers/spicy-sections)
-(also see [proposal to transfer](https://github.com/openui/open-ui/issues/517)).
-
-It [appears to propose](https://bkardell.com/common-panel/index.src.html#requirements-for-panels-and-panel-sets)
-a consistent interaction model across all of these types,
-which is a different approach from that taken in the CSS Toggles work
-and likely not compatible with user expectations.
-However, this is probably not an integral aspect of the proposal.
+Within Open-UI there was another proposal 
+which was to be called [panelset](https://tabvengers.github.io/spicy-sections/).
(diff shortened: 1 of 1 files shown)


site/src/pages/components/accordion.explainer.mdx
```diff
@@ -341,46 +341,22 @@ particularly to understand
...
-for multiple UI components
-that present things that are semantically headings and sections.
-This differs from the approach proposed here which builds technology
-for just a single presentation.
-The approach of building one thing has clear advantages
-because it is a single control for a single underlying semantic,
-but also presents issues because the different interaction models
-have different accessibility characteristics,
-which creates some difficult problems
-as demonstrated in the work on CSS toggles.
```
dbaron, line 383: Maybe this part should be kept?  I *think* it's still true.

bkardell: I'm not sure what specifically you mean there.  Line 364 'preferreddisplay' thing is from the old proposal and I think if you want to draw a comparison with toggles. The preceding 6 lines, I believe, cover the relevant bits. The different a11y characteristics of each are handled in the panelsets demo as well.  If there is a problem there it is well expressed in my opinion in the linked #559 which is, as you can see, complicated.  

re: 366, if you look on https://tabvengers.github.io/spicy-sections/demonstration/#tabset-state-list on a wide enough screen you'll see tabs using this. It's not just CSS, it's minimal shadow dom. I don't think styling is an issue - @mirisuzanne or @jonathantneal can perhaps agree or deny, but we had several examples. Yay shadow dom?

I thought leaving in the 

> The panelset proposal has enough open questions that it's a bit hard to assess 

and linking in that issue kind of covered it well enough.

But if we want to compare them, I'm not sure if we should compare the whole thing vs the affordance part?  That is, there are two levels to our proposal, an attribute and the css property.  The reasons for the css property seem to apply equally well to this proposal which just sidesteps it and says "use some JavaScript".  That seems an equally easy change to make for now to ours, so let's leave that aside and compare just ... the markup? Because I mean, if we are changing attributes then some characteristics might change on summary details with name too, right?

dbaron (approved):
OK, I think it makes sense to merge this so we get rid of things that are wrong, and perhaps I'll try to write some additional analysis later.

PR #722, opened 2023-04-17
Update Bootstrap links from 4.* to 5.3

julien-deramond:
This PR mainly fixes Bootstrap's URL pointing to v4.x instead of v5.3 (latest version of Bootstrap).

Some extra modifications:
- Typo in `site/src/pages/components/breadcrumb.mdx`
- Changed the `lastUpdated` date to today's date in `site/src/sources/bootstrap.json`
- Renamed "Alerts" to "Alert" in `site/src/sources/bootstrap.json` to match the name of the component displayed in the documentation so that the Bootstrap component is displayed in the list of references.
- Replaced "description" which doesn't exist in the schema by "definition" in `site/src/sources/bootstrap.json`
- Removed Jumbotron and Media components that don't exist in Bootstrap anymore

This is a first easy batch of modifications. If needed I could also update the screenshots of the components in the documentation based on the new rendering of Bootstrap's components.

PR #760, opened 2023-05-28
Clean up source data

gfellerph:
## What is happening here?

This pull request aims to clean up some of the json data in the sources by:
- renaming `description` to `definition` on the component level
- removing `category` as it's unused
- fixing typos, e.g. `decription` to `description` and switching invalid property names to their possibly intended valid counterpart
- pointing datasets to correct schema
- marking the `image` property on the concepts level as optional (about 576 images are missing currently)
- removing duplicate properties
- fixing indentation

## What effects does this have?

Some descriptions that were previously not displayed should now be visible. Also, some data that was not visible on the research pages due to typos should now be displayed.

This PR also reduces the number of JSON schema warnings to 1 (a missing URL for the chromium file upload where no useful URL can be found as it's the native component viewed in a chromium browser).


dbaron:
It's possible that the existing data should instead be cleaned up, e.g., as some was in #722.  Do you know if there's something that uses these values?

gfellerph:
@dbaron thats possible. Antd (https://github.com/openui/open-ui/blob/main/site/src/sources/antd.json#L323) uses `description` which could be migrated to `definition`.  `category` (https://github.com/openui/open-ui/blob/main/site/src/sources/bootstrap.json#L286) is only used by bootstrap and the data does not seem to be displayed on the research pages. Do you think it could be dropped?

gregwhitworth:
@gfellerph that seems reasonable to me; we don't refresh these too often to where any of the future collisions will be too much of a burden.

dbaron:
> @dbaron thats possible. Antd (https://github.com/openui/open-ui/blob/main/site/src/sources/antd.json#L323) uses `description` which could be migrated to `definition`. `category` (https://github.com/openui/open-ui/blob/main/site/src/sources/bootstrap.json#L286) is only used by bootstrap and the data does not seem to be displayed on the research pages. Do you think it could be dropped?

Both of those edits sound good to me.  (Sorry I missed this!)

dbaron (commented):
As described above in the discussion in the PR, I think a better path would be to clean up the data -- which would cause some data not currently displayed on the website to be displayed.

dbaron:
(And when you update this PR, please use the "Re-request review" icon next to my name in the reviewers list -- that will make it much more likely that I see it.)

site/src/schemas/component.schema.json
```diff
@@ -52,7 +52,7 @@
           "openUIName": { "type": "string" },
           "image": { "type": "string" }
         },
-        "required": ["name", "image"]
```
gfellerph, line 55: Having "image" as a required property on the concept object generates 576 warnings on the current dataset.

gfellerph:
@dbaron, @gregwhitworth, I updated this PR to clean up the data instead of extending the schema. In the process I noticed some other warnings and went ahead and fixed them. If any of these fixes don't fit into this change or need further discussion, I'll be happy to submit them in a separate pull request.

I also updated the [description of this PR](https://github.com/openui/open-ui/pull/760#issue-1729673465) to reflect the new changes.

site/src/sources/evergreen.json
```diff
@@ -194,8 +194,7 @@
           "name": "selection",
           "image": "evergreen-menu-selection.png"
         }
-      ],
-      "url": "https://evergreen.segment.com/components/menu"
```
dbaron, line 198: Maybe drop the trailing `/` from the one ~20 lines above?

site/src/schemas/component.schema.json
dbaron: I have mixed feelings about this change.  Many of the pages are less useful than they should be because we don't have images.  But I also recognize that it's hard to produce all the necessary images and we're not always going to get them...

dbaron (approved):
Two small comments, but given the amount of work you've done and the amount that this improves things, I'm going to go ahead and merge this anyway.  It might be worth doing followups for the comments -- although I don't think it's critical.

Thanks very much for cleaning this up.

site/src/schemas/component.schema.json
gfellerph: In hindsight, it would maybe be better to keep it for encouraging adding images to future research instead of "fixing all the warnings". My initial thought was that it reflects the current state better.

Issue #526, opened 2022-04-29
[popup] "Interest"-based triggering for popovers


mfreed7:
The explainer currently [says this](https://github.com/openui/open-ui/blob/main/research/src/pages/popup/popup.research.explainer.mdx#classes-of-ui---dismiss-behavior-and-interactions) about the behavior for `popup=hint` (which I'll call "tooltips" for clarity):

> Hint/Tooltip (popup=hint)
> - When opened, force-closes only other hints, but leaves all other popup types open.
> - Dismisses on [close signal](https://wicg.github.io/close-watcher/#close-signal), click outside, when no longer hovered (after a timeout), or when the anchor element loses focus.

For the light dismiss triggers, everything except `when no longer hovered` is well defined. However, since we haven't discussed a declarative way to **trigger** tooltips, it isn't clear how "no longer hovered" should be defined. And even if there was a declarative trigger, what happens for tooltips that are programmatically triggered via `tooltip.showPopup()`? In order to answer questions about light dismiss for tooltips, I think we need to figure out how declarative invocation works.

Here's a quick strawman off-the-cuff idea:
```html
<img src=questionmark.png hoverhint=foo>
<div popup=hint id=foo>Tooltip</div>
<style>
img {
  hint-trigger-delay=500ms;
}
[popup=hint] {
  hint-hide-delay=1000ms;
}
</style>
```

In other words, there is a new content attribute, `hoverhint`, which points to an element with `popup=hint`. When the `<img>` element (in this example, but could be any element type) is hovered, after a delay set by the `hint-trigger-delay` CSS property, the tooltip is shown. When that tooltip (or the triggering `<img>` element) has no longer been hovered for a time set by `hint-hide-delay`, it is hidden.

Again, this is a total strawman proposal. Suggestions appreciated.

scottaohara:
i think i'd just expect a `:hover` and `:focus` for the selector to tie it to those actions.  As otherwise i'd read that as expecting it to show by default after the 500ms delay.

e.g.,
```html
*:hover,
*:focus {
  hint-trigger-delay=500ms;
}
```

What I don't readily have answer for is how non-interactive elements will allow users who aren't using a mouse to access their tooltips (e.g., your image example) or how these tooltips will be accessible on touch screens.

What might be useful is if caret navigation had it's own pseudo-class similar to `:hover`, or tapped into`:hover`, as that could be a means to trigger a hint on an otherwise non-focusable element.  

[...]

PR #766, opened 2023-06-08
Add explainer for `popover=hint` and hover triggering of popovers

mfreed7:
This is the first draft of a renewed explainer for two features:
     1. hover triggering of popovers
     2. `popover=hint` behaviors

This functionality was initially included in the original `popover` explainer, but [was removed](https://github.com/openui/open-ui/issues/526#issuecomment-1219845155) because of lingering uncertainty about a number of a11y and other issues. Now that the popover API has landed in WHATWG/html, this PR brings back a fresh explainer for those features.

This is intended to be a public working draft of this document, not a final explainer.


mfreed7:
@scottaohara @aleventhal PTAL and let me know what you think. I kept your names on this, but can remove them if desired. Hopefully most of what's here isn't a surprise, since it's from documents you've seen. 😄 

I wanted to get this back in the public so we can start to really discuss it again.

mfreed7:
@andrico1234 (or others) I could use some help with the build step. It gives this error:

```
3:48:50 PM: [@mdx-js/rollup] Could not parse expression with acorn: Unexpected token
3:48:50 PM: file: /opt/build/repo/site/src/pages/components/popover-hint.research.explainer.mdx
3:48:51 PM: The language "webidl" doesn't exist, falling back to plaintext.
 error   Could not parse expression with acorn: Unexpected token
```

but doesn't give a line number, and `webidl` doesn't exist in the file. I'm guessing this has to do with one of my backtick-escaped code sections, but I don't want to guess and check. Hints appreciated. It'd be amazing if this error message came with a line number.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+# `popover=hint` Behavior
+
+When a hint is shown, it behaves differently from other popover types such as `popover=auto` and `popover=manual` in a few ways:
+
+|               | `popover=auto`            | `popover=hint` | `popover=manual` |
+| ------------- | ------------------------- | -------------- | --------------   |
+| Light dismiss |      Yes                  |     Yes        |      No          |
+| Force-hides:  | Other `auto`s and `hint`s | Other `hint`s  | Nothing          |
+| Nesting:      |     Yes                   | No             | No               |
```
keithamus, line 120: Would it be right here to say there can only be 1 open `hint` on the page, many nested (but only 1 "root") `auto` and many `manual`?

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+This explainer discusses two important extensions of the Popover API:
+
+1. The ability to trigger popovers via a "hover" type interaction.
+2. The ability to show a new type of popover for "hints".
+
+These two capabilities are orthogonal but closely related, and this explainer therefore discusses them both.
+
+The motivating use case, but certainly not the only use case, for these two capabilities is in the construction of "tooltips". Typically, tooltips are transient bits of additional information related to a bit of content. This additional information is *not required* and is supplementary in nature. Usually the user accesses this tooltip by hovering, keyboard-focusing, or long-pressing a control.
+
+# "Hover" triggering
```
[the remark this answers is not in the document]
iansan5653: I agree - `hover` is misleading in that it applies only pointer compatibility. Another alternative might be `active`?

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+To facilitate these delays, two new CSS properties are added:
+
+- `popover-show-delay` - applied to the invoking element. The invoking element must be hovered, focused, or long-pressed for this long before the popover is shown. The default value for this property is `0.5s`.
+- `popover-hide-delay` - applied to the popover. The invoking element *or* the popover must be **not** hovered, focused, or long-pressed for this long before the popover is hidden. The default value for this property is `infinity`.
```
iansan5653, line 72: Does a default value of `infinity` mean the popover does not automatically hide itself by default? To me this feels unexpected considering it automatically shows itself. I'd expect something like `0.5s` to be the default here as well.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+However, when a hint is used for something more complex, such as a dialog that appears on hover, or a data table, then the user needs to be able to learn it’s there and navigate to it, so that they can use additional screen reader commands to interact with it.
+
+## A11y API support
+
+For the best screen reading experience, the implementation will need to expose different properties for a plain hint vs a rich hint (classification described above).
+
+- Plain hints: the browser will simply expose the text on the triggering element. The actual popover target element and its descendants can be invisible/ignored in the AX tree.
+    1. If no other accessible name is available, use hint’s inner text for the accessible name
+    2. If it's used as the name, then use it to compute the description, setting the described by relation to point from the trigger to the hint element
```
khiga8, line 148: > If it's used as the name

I am confused by this bit. Do we mean, if the triggering element already has an accessible name?

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+Finally, `popover=auto` popovers support "nesting": the idea that one popover is a descendant of another popover, for example in the case of nested menus. It makes sense in that example for the parent menu to stay open while the child menu opens. The same is not true for hints - it would be confusing for one hint to be nested inside another, and wouldn't make sense. Therefore, hints do not support nesting.
+
+
+# Accessibility
+
+A key distinction between types of hints is whether it may be important for a user to explore or interact with the hint. Consensus on these types of hints/tooltips can help with the resolution of [ARIA#979: Clarify the use of role=tooltip](https://github.com/w3c/aria/issues/979).
+
+- A plain hint contains only elements with a computed role of generic, text and images.
```
khiga8, line 133: Given that `aria-describedby` will not properly convey semantics, I don't think that anything other than a text string should be allowed as a "plain hint".

site/src/pages/components/popover-hint.research.explainer.mdx
scottaohara: this should probably be something like "if the triggering element already has an accessible name, then the hint becomes the accessible description..."

this is essentially supposed to mimic the current behavior of the title attribute.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+## Popover types
+
+The new hover-triggering functionality should work for all `popover` types. There are use cases, e.g. for `popover=auto`; for example, a nested menu structure where sub-menus are shown when items in the primary menu are hovered or focused.
+
+## Example
+
+```html
+<button popovertarget=menu1>Menu</button>
+<menu popover id=menu1>
```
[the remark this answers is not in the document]
scottaohara: but none of the list items or buttons have role=menuitem either, so it matches the HTML example of how to create a "menu" list of actions (buttons).

but to your point, yes, people aren't super familiar with the menu element and often misunderstand that menu !== menu.  :(

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
scottaohara: that should be exactly how this would be conveyed, and that's why these 3 types of content would be expected/allowed, since all structural semantics would be dropped in the description computation, so the flattened text string of the elements and the alternative text of the image would be used to create the description.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+When a hint is shown, it behaves differently from other popover types such as `popover=auto` and `popover=manual` in a few ways:
+
+|               | `popover=auto`            | `popover=hint` | `popover=manual` |
+| ------------- | ------------------------- | -------------- | --------------   |
+| Light dismiss |      Yes                  |     Yes        |      No          |
+| Force-hides:  | Other `auto`s and `hint`s | Other `hint`s  | Nothing          |
+| Nesting:      |     Yes                   | No             | No               |
+
+First, hints should always be light dismissable. They are transient, supplementary information, so they should not require affirmative action to close them. Clicking outside or hitting ESC will close a hint.
```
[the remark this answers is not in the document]
scottaohara: moving to button 2 should close the hint from button 1.

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
gregwhitworth: I agree with @keithamus and @iansan5653 for keyboard users the popup should still appear upon traversal to the element that has the associated popup. That said, I'm curious of @scottaohara thoughts as that may produce numerous popups while tabbing through a site.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+      Sub menu 1
+    </button>
+  </li>
+  <li>
+    <button popovertarget=submenu2 popovertargetaction=hover>
+      Sub menu 2
+    </button>
+  </li>
+</menu>
+<menu class=submenu popover id=submenu1>
```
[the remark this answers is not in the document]
scottaohara: probably should just move away from menu elements all together, as the example doesn't demonstrate role=menu, but people may assume it is one

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+For the best screen reading experience, the implementation will need to expose different properties for a plain hint vs a rich hint (classification described above).
+
+- Plain hints: the browser will simply expose the text on the triggering element. The actual popover target element and its descendants can be invisible/ignored in the AX tree.
+    1. If no other accessible name is available, use hint’s inner text for the accessible name
+    2. If it's used as the name, then use it to compute the description, setting the described by relation to point from the trigger to the hint element
+
+- Rich hints:
+    1. Add `aria-expanded` on the triggering element: the same treatment as for `popover=auto`.
+    2. Minimum role of `tooltip` on the target — see [spec PR describing minimum role](https://github.com/w3c/html-aam/pull/454).
+    3. Add `aria-details` pointing from the trigger to the popover. This allows screen readers to announce the presence of the popover and a command for navigating to it.
```
scottaohara, line 153: we may want to overtly call out that a rich hint would not contribute to the description calculation, and i'm almost certain we should not include it in the naming calculation either.

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
keithamus: I agree. Perhaps I’m misreading but the explainer doesn’t feel clear that this is the case.

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
scottaohara: it is still a sort of open question about this that I think may need to be brain stormed a bit.

For instance, in the scope of tooltips it may be better to think about introducing a user setting or alternate key to show/hide a tooltip at will.  Esc is often used to dismiss, but the only way to get a tooltip back is to navigate away and then back to the element.  That's silly and we could do better with this.  

But there are other types of UI - e.g., text/search input fields with autocomplete popups that one might want to show up immediately on focus, and it'd be good if we could support both.

site/src/pages/components/popover-hint.research.explainer.mdx
scottaohara, line 122: @keithamus this help?

```suggestion
First, hints should always be light dismissable. They are transient, supplementary information, so they should not require affirmative action to close them. Clicking or tapping outside, hitting the <kbd>ESC</kbd> key, or moving focus away from the invoking element will close a hint.
```

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+For a screen reader user, there already exists simple settings and commands to read accessible descriptions, which are just an additional piece of text associated with an object, such as used by a `title` or `aria-description` attribute.  When a hint is just used for a prettier tooltip, a screen reader can reuse this simple mechanism and act like a title was present. There is no need for the user to navigate to it.
+
+However, when a hint is used for something more complex, such as a dialog that appears on hover, or a data table, then the user needs to be able to learn it’s there and navigate to it, so that they can use additional screen reader commands to interact with it.
+
+## A11y API support
+
+For the best screen reading experience, the implementation will need to expose different properties for a plain hint vs a rich hint (classification described above).
+
+- Plain hints: the browser will simply expose the text on the triggering element. The actual popover target element and its descendants can be invisible/ignored in the AX tree.
```
scottaohara, line 146:
```suggestion
- Plain hints: the browser will expose the flattened text string comprised of the hint's text and text alternative content via the invoking element. The actual popover target element and its descendants can be invisible/ignored in the AX tree. Note: this would be similar to how `title` attribute tooltips are treated today.
```

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+    1. If no other accessible name is available, use hint’s inner text for the accessible name
```
scottaohara, line 147:
```suggestion
    1. If the invoking element lacks an accessible name, use computed text of the hint as the element's accessible name
```

site/src/pages/components/popover-hint.research.explainer.mdx
scottaohara, line 148:
```suggestion
    2. If the hint is not used as the accessible name, then the hint would be treated as the accessible description of the invoking element. A described by relation would point from the trigger to the hint element.
```

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
mfreed7: Let's move at least the bikeshedding part of this discussion (which is great, thanks all!) over to this issue: https://github.com/openui/open-ui/issues/767

Scott, we can open another issue to discuss your comments about a user setting or alternate key - does that make sense?

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
mfreed7: Interesting feedback! I had envisioned these to work with popovers (and maybe dialogs?) only. How do you envision making them more general?

As to `popover-show-delay` I'm not sure this can be done with `animation-delay` - can you prove me wrong? My concern is that the behavior wouldn't be de-bounced in that case. I.e. set `popover-show-delay` to 0.5 seconds, and have the user hover and then de-hover the element only for 0.2 seconds. Nothing should happen, but I think `animation-delay` would cause the popover to just be shown after 0.5 seconds, right?

site/src/pages/components/popover-hint.research.explainer.mdx
mfreed7: Yes, that's a great summary. I think I'll add that text to the explainer, thanks!

site/src/pages/components/popover-hint.research.explainer.mdx
mfreed7: Right - `infinity` means no auto-hide behavior. I had to do that because the property must have a value for all elements, and any popover therefore is subject to this behavior. Note that as-described, `popover-hide-delay` works on any popover, regardless of whether it was hover-triggered.

site/src/pages/components/popover-hint.research.explainer.mdx
mfreed7: Thanks - I've updated the text with your suggestion.

site/src/pages/components/popover-hint.research.explainer.mdx
mfreed7: Interesting comment! @scottaohara would you agree? I think it sounds pretty limiting to only allow pure text, and not even images.

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
mfreed7: I'm one of those people. Even though I've been told this, I consistently forget it.

Is the "correct" thing to use `<div role=menu>` then? For now, I've updated the explainer to do that, but let me know if there's something better.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+- `popover-hide-delay` - applied to the popover. The invoking element *or* the popover must be **not** hovered, focused, or long-pressed for this long before the popover is hidden. The default value for this property is `infinity`.
+
+It is important to "debounce" these delays. Said another way, the delays are interpreted as *the time an element must be continuously hovered or de-hovered*. The naive Javascript implementation of this behavior is to just call `setTimeout(() => show/hidePopover(), delayValue)`. I.e. after the invoking element is hovered, wait `delayValue`, and then show the popover. But that does not properly handle the situation where the user hovers the element, and then de-hovers it, all within the delay time. In that case, no popover should be shown. Similarly, once the popover is shown and the user de-hovers the popover briefly but re-hoveres it again quickly, the popover should not close.
+
+## Popover types
+
+The new hover-triggering functionality should work for all `popover` types. There are use cases, e.g. for `popover=auto`; for example, a nested menu structure where sub-menus are shown when items in the primary menu are hovered or focused.
```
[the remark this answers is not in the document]
mfreed7: But there are very common menu patterns that do exactly this. Google docs is one of them: the menu bar menus all have hover triggered sub-menus. Perhaps your point is that this is a bad pattern, but if people are going to do it, I'd love to provide them the "correct/best" way via a supported API, rather than having them all roll their own.

Is your concern shifting focus (presumably via the `autofocus` attribute) only? If so, maybe we could think about modifying the behavior of `autofocus` in the hover-triggered case?

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
mfreed7: I was going to say the same thing - "Hello world" does seem like the best thing to read in this case. Maybe I'm wrong.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,169 @@
...
+
+Finally, `popover=auto` popovers support "nesting": the idea that one popover is a descendant of another popover, for example in the case of nested menus. It makes sense in that example for the parent menu to stay open while the child menu opens. The same is not true for hints - it would be confusing for one hint to be nested inside another, and wouldn't make sense. Therefore, hints do not support nesting.
+
+
+# Accessibility
+
+A key distinction between types of hints is whether it may be important for a user to explore or interact with the hint. Consensus on these types of hints/tooltips can help with the resolution of [ARIA#979: Clarify the use of role=tooltip](https://github.com/w3c/aria/issues/979).
```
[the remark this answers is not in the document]
mfreed7: @scottaohara I'll leave this comment to you, as I don't have enough context.

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
mfreed7: So I think this behavior (**bluring** button 1) is handled by the `popover-hide-delay` CSS property. If hint 1 was shown because button 1 was focused (for `popover-show-delay` seconds), then it will be hidden once button 1 is blurred for `popover-hide-delay` seconds.

I've added an explicit example and a lot more text around this. LMK if that helps clarify.

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
mfreed7: I've moved to `<div role=menu>`, but yeah I'd be concerned that adding all of the details would muddy the waters. I **do** however think that the hover-triggered menu is a pretty common use case, so it's good to include. I can perhaps include some text alluding to this not being a "complete" solution?

site/src/pages/components/popover-hint.research.explainer.mdx
mfreed7: If you suggest some good wording for this, I'm happy to include it!

site/src/pages/components/popover-hint.research.explainer.mdx
mfreed7: My only concern is the last thing about moving focus away. We explicitly didn't do that for `popover=auto` because it was jarring and hard to get right. I'd say we definitely shouldn't do that for "rich hints" for all of the same reasons.

My intention was to make the "move focus away to dismiss" is handled entirely by `popover-hide-delay`. See the new text I've added, and LMK if this allays any of your concerns here.

site/src/pages/components/popover-hint.research.explainer.mdx
mfreed7: Thanks, I incorporated this.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+## Popover types
+
+The new hover-triggering functionality should work for all `popover` types. There are use cases, e.g. for `popover=auto`; for example, a nested menu structure where sub-menus are shown when items in the primary menu are hovered or focused.
+
+## Example menu
+
+```html
+<button popovertarget=menu1>Menu</button>
+<div role=menu popover id=menu1>
```
scottaohara, line 109: if this changes to role=menu, then there's a lot of extra updates that need to happen to make the markup pattern correct (role=none on all the li elements, role=menuitem on all the buttons, maybe even use tabindex=-1 to allude to the arrow key navigation)

but if we change this to a `ul`, then it's just a list like it was with the `menu` element, which means this would all just be tab key navigation and we don't need to distract with all the extra ARIA/implied keyboard functionality.  

we can go either way, it's fine.  we just need to add in all the other attributes if we go this route.

site/src/pages/components/popover-hint.research.explainer.mdx
scottaohara: no i don't agree.  that's why there would be the two concepts of a rich hint vs a standard hint.  

per the hello world example above, or even something like the following:

```
<input aria-describedby=foo>
<p id=foo><img src=... alt=warning:> this field cannot be empty</p>
```

the above is perfectly fine to convey as "warning: this field cannot be empty"

as opposed to "warning, image, this field cannot be empty"

site/src/pages/components/popover-hint.research.explainer.mdx
scottaohara: per your mention of doing addon work in another issue/PR

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+# `::tooltip` pseudo element
+
+[CSSWG Issue 8930](https://github.com/w3c/csswg-drafts/issues/8930) discusses the addition of a `::tooltip` (or similarly named) pseudo element that could be used to generate tooltips via CSS, and style them. These are intended (though see the discussion) to be Plain Hints (as described above). This proposal may be complimentary to this explainer's feature proposals, or it might replace some of them.
+
+# Monkeypatch spec
+
+{To-do: document the changes necessary to the [WHATWG/html spec](https://html.spec.whatwg.org/multipage/C#the-popover-attribute) for this proposal.}
```
dbaron, line 188: I think the use of `{}` here may be what's causing the deploy failure, see https://mdxjs.com/docs/troubleshooting-mdx/#could-not-parse-expression-with-acorn-error .  Maybe you want `\{` and `\}` ?

dbaron (commented):
It seems like this review is settling down, although there are a few open comments about the technical stuff.  One comment about trying to actually get this landed, though:

site/src/pages/components/popover-hint.research.explainer.mdx
mfreed7: Ok, thanks for the suggestion. I've changed this to `<ul>`s, and removed the `role`.

site/src/pages/components/popover-hint.research.explainer.mdx
mfreed7: Thanks!

site/src/pages/components/popover-hint.research.explainer.mdx
mfreed7: Great - thanks!

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
keithamus: https://codepen.io/keithamus/pen/yLQabJK this uses `animation-delay` to ensure that as you hover and de-hover (or focus/defocus) the `button` for less than `0.5s`, the `div` will not be shown. I believe this is properly debounced as you describe. I don't know how to get the same effect upon exit of that state; transitions don't work quite the same, I believe.

keithamus:
I'm happy to file issues around the discussion points I had, and we can land this.

dbaron (approved):
Since I think this has already gotten sufficient technical review, I've reviewed mainly for markup and site integration.  I think this looks good modulo a few comments below, although one may require a little thought and another might require a few rounds of pushing attempts to fix it and checking the deploy preview.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,195 @@
+---
+menu: Proposals
+name: Popover=hint and Hover Triggering (Explainer)
```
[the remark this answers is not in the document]
mfreed7: Ok. I went with `Popover Hint and Hover` for now. Not awesome, but at least it's short enough. Likely we should try to remove "(Explainer)" from all of them anyway, since it's quite repetitive and not informative.

@andrico1234 for further comments. But I think ok for this PR.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,195 @@
...
+path: /components/popover-hint.research.explainer
+pathToResearch: /components/popup.research
+layout: ../../layouts/ComponentLayout.astro
+---
+
+# Popover Hover Triggering and `popover=hint`
```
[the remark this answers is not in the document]
mfreed7: Oh, yeah. Plus the `popover=hint` part looks really funny.

Ok, removed. And I added a quick summary. I'm happy to add/modify in a subsequent PR.

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+- [@mfreed7](https://github.com/mfreed7), [@scottaohara](https://github.com/scottaohara), [@aleventhal](https://github.com/aleventhal)
+- June 8, 2023
```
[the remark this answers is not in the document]
mfreed7: Done. As I'm not sure what will happen to my review approval if I add files, I'll refrain from fixing the original popover explainer for right now.

site/src/pages/components/popover-hint.research.explainer.mdx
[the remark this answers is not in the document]
mfreed7: Ahh nice - I see that you get that by adding `:focus` to clear the animation if you de-hover. But you're right - not possible on the de-hover side of things.

Do you think this is worth opening a CSSWG issue to discuss? I mean about the "more general problem" question?

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -0,0 +1,195 @@
...
+
+However, when a hint is used for something more complex, such as a dialog that appears on hover, or a data table, then the user needs to be able to learn it’s there and navigate to it, so that they can use additional screen reader commands to interact with it.
+
+## A11y API support
+
+For the best screen reading experience, the implementation will need to expose different properties for a plain hint vs a rich hint (classification described above).
+
+- Plain hints: the browser will expose the flattened text string comprised of the hint's text and text alternative content via the invoking element. The actual popover target element and its descendants can be invisible/ignored in the AX tree. Note: this would be similar to how `title` attribute tooltips are treated today.
```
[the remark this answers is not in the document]
mfreed7: Markdown is cool, until it's not. I'll try. I think you're right about lining things up with the "Plain"/"Rich".

mfreed7 (commented):
Thanks for the review!

PR #774, opened 2023-06-20
rm extra -

tantek:
minor punct edit


dbaron (approved):
Was someone writing LaTeX instead of Markdown?

PR #775, opened 2023-06-20
Stop trying to use webidl syntax highlighting that doesn't exist.

dbaron:
Based on the log at
https://app.netlify.com/sites/open-ui/deploys/648757635e920e0008e01eb0 it seems like this attempt to use webidl syntax highlighting causes the warning (unattributed to any file):
    The language "webidl" doesn't exist, falling back to plaintext.

This removes the unneeded language (which doesn't cause any highlighting) and adds a closing brace as well.


dbaron:
(I noticed this while trying to understand the deploy failure in #766.)

mfreed7 (approved):
Nice! I'd seen that error also and couldn't figure out where it came from. Did you just grep for \`\`\`webidl or something?

dbaron:
Yes, I just searched for "webidl", and that seemed to make sense as the cause.

PR #776, opened 2023-06-20
replace --- with emdash

tantek:
minor punct fix, replace --- with — emdash as that looks like what was intended


dbaron (approved):
More TeX syntax in our markdown!

Issue #648, opened 2022-12-13
[selectmenu] Mechanism to use <select> as progressive enhancement?


hidde:
From: https://home.social/@ranger/109505541067518807 

In browsers that don't support `<selectmenu>`, you could do something like this to fallback to `<select>`:

```html
<selectmenu>
  <select>
  …</select>
</selectmenu>
``` 

(A bit, but probably not the same, like how you could put fallback content in a `<video>`)

Should there be a mechanism, when browsers _ do_ support `<selectmenu>`, that they ignore any nested `<select>`s and assume they are only there as a fallback? Would that even be a safe assumption to make?

brechtDR:
Creating something like this would definitely be beneficial. 

Personally, I think that when a developer chooses to use a selectmenu, it will be with the latest browsers in mind and thus could be ignored. But as I stated, that's just my opinion. 

In any case, it would make fallbacking a lot easier and could result to faster adaption and reduce of polyfills.

Sidenote though: Both should be easy to target with JS (something that targets both cases) especially for bigger filter menus.

mfreed7:
So the conclusion of the meeting above was that we generally agree that there should be a progressive enhancement mechanism for `<selectmenu>` but we're unclear what it should be. Nested `<select>` likely has some parser issues, because not all content is allowed inside `<select>`. There was a comment (that I agree with) that perhaps HTML needs a better way of feature detection, akin to CSS's `@supports`.

The action item here is to suggest a way forward that provides progressive enhancement.

josepharhar:
> Sidenote though: Both should be easy to target with JS (something that targets both cases) especially for bigger filter menus.

You should be able to do this by checking `window.HTMLSelectMenuElement`, right?

hidde:
When thinking about progressive enhancement, I think the other way around, eg what would a browser do that doesn't know what a `selectmenu` is? 

With something like 

```
<selectmenu>
  <select>
  </select>
</selectmenu>
```

I would expect a browser that doesn't know about `selectmenu` would throw that wrapping `selectmenu` tag out and/or ignore it, and then it would just render the `select` inside, with no JS needed to get that behaviour. Then a browser that _does_ know what `selectmenu` is, would throw any `select` element inside of it out, as it knows that that is a fallback. 

[...]

PR #783, opened 2023-07-17
Add section about select in selectmenu explainer

josepharhar:
This was requested in the last OpenUI meeting


site/src/pages/components/selectmenu.mdx
```diff
@@ -378,3 +379,12 @@ Like other form control elements, selectmenu supports validity checking attribut
...
+- The `<select>` popup currently has the capability to break outside the window of the browser or of an iframe. Sites likely depend on this behavior, but we would need to remove it for security reasons if `<select>` is allowed to contain arbitrary content. So again this would need to be opt-in to avoid breaking existing sites.
+
+In order to resolve these two issues, we could have an attribute opt-in on the select element, like `<select custom>`. However, this has its own issues as well: consider the case where the `<select>` is created dynamically via `document.createElement()`, when the custom attribute must be set using `setAttribute()`. When should the transition to the custom behavior be applied? Applying it at the point of `setAttribute()` would be inconsistent with how other elements like `input[type=range]` work when created dynamically; typically the transformation would occur when the element was appended to the DOM. But, this means that for a dynamically created `<select>`, the developer would have to wait until the `<select>` is appended to the DOM before calling `attachShadow()`. Creating a new tag name for the element gets around this issue.
```
josepharhar, line 390: The old explainer shows how `<selectmenu>` is better than `<select custom>` by giving this example of the author calling attachShadow() on the select element, which is not something that the current design allows for anymore. Are there any other reasons that we are going with selectmenu instead of select custom, or should this be brought back into question? @gregwhitworth @dandclark @mfreed7

dandclark: I think the parsing issue is still a problem with `<select custom>`. I'm not aware of any other cases where we make a major parsing behavior change based on the presence of an attribute on an element. This would add complexity that is probably best avoided. There are all sorts of unpleasant cases to consider, like what if an inline script that's a child of the `<select custom>` removes the attribute before the `<select>` element has closed.

josepharhar: Thanks! I replaced the attachShadow text with what you said

site/src/pages/components/selectmenu.mdx
```diff
@@ -378,3 +379,12 @@ Like other form control elements, selectmenu supports validity checking attribut
...
+
+- Changing the parsing behavior for `<select>` to allow other element children by default would likely break a lot of old sites that depend on the current behavior.
+- The `<select>` popup currently has the capability to break outside the window of the browser or of an iframe. Sites likely depend on this behavior, but we would need to remove it for security reasons if `<select>` is allowed to contain arbitrary content. So again this would need to be opt-in to avoid breaking existing sites.
+
+In order to resolve these two issues, we could have an attribute opt-in on the select element, like `<select custom>`. However, this has its own issues as well: we would have to make a major parsing change based on the presence of an attribute, which doesn't happen for other attributes/elements. What if a `<select custom>` is parsed and an inline script child of it removes its `custom` attribute before the `<select>` has been closed?
```
gregwhitworth, line 390: @josepharhar thank you for writing this up. I wouldn't end this with this question; I'd end with the emphatic statement that it is not backwards compatible.

gregwhitworth (changes requested):
I can't quite put my finger on it - but I feel like it's missing something.

site/src/pages/components/selectmenu.mdx
josepharhar: Ok, I added another paragraph to the end. What do you think?

gregwhitworth: @josepharhar How about this as I feel like the script question throws off the flow a bit:

> - The `<select>` popup currently has the capability to break outside the window of the browser or of an iframe. Sites may depend on this behavior, but we would need to remove it for security reasons if `<select>` is allowed to contain arbitrary content. 

> Given that we would be required to introduce something new into the platform (attribute, element, etc) to opt-in to the new behavior; we opted for a new element. This comes with the added benefit that over the years most developers that want customization of controls and components will not leverage `<select>` at all; akin to how developers leverage `grid` and `flex` display types for layout instead of `<table>` or `float`. Additionally, this has the potential to simplify the implementation cost since we do not have to make substantial changes to the current `<select>` parser.

@josepharhar Thoughts? @dandclark @mfreed7 if you all want to add some more weight to the parsing changes; feel free.

gregwhitworth:
@josepharhar some folks on twitter just pinged me about this so I sent them the PR to look at. Additionally, they raised progressive enhancement, so can we add in another FAQ of "Why can't the `<selectmenu>` be nested inside of the `<select>` and cover the content here: https://github.com/openui/open-ui/issues/648

site/src/pages/components/selectmenu.mdx
```diff
@@ -378,3 +379,14 @@ Like other form control elements, selectmenu supports validity checking attribut
 * [HTML spec for `checkValidity()`](https://html.spec.whatwg.org/#dom-cva-checkvalidity)
 * [HTML spec for `reportValidity()`](https://html.spec.whatwg.org/#dom-cva-reportvalidity)
 * [HTML spec for `setCustomValidity()`](https://html.spec.whatwg.org/#dom-cva-setcustomvalidity)
+
+## Why not improve `<select>`
+
+Instead of creating a new `<selectmenu>` element, we could improve the existing select element to meet the same needs. However, there are a few problems with this (borrowed from the [Custom Controls UI Explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ControlUICustomization/explainer.md):
+
+- Changing the parsing behavior for `<select>` to allow other element children by default would likely break a lot of old sites that depend on the current behavior.
```
justinfagnani, line 387: Are there any references to attempts to validate or quantify the back-compat concern here? HTTP Archive queries for `<select>` containing children other than `<option>` perhaps?

gregwhitworth: @justinfagnani this is a valid ask and no we didn't concretely look into an enumeration of potential content that would be impacted. I presume you're asking from a PE perspective as the parsing changes to support arbitrary content is non-trivial and has security issues since the current select popup escapes window bounds.

gregwhitworth: @bkardell you're a wiz with HTTP archive, any chance you can do a query to derive non option or optgroup children of a select?

site/src/pages/components/selectmenu.mdx
josepharhar: ok, i replaced the ending with your suggestion

gregwhitworth:
@josepharhar while I know the anatomy may be changed, there are failures with the deploy. Do you want to fix them or abandon this PR?

josepharhar:
I pushed a commit which should fix the build failure

site/src/pages/components/selectmenu.mdx
```diff
@@ -378,3 +379,12 @@ Like other form control elements, selectmenu supports validity checking attribut
 * [HTML spec for `checkValidity()`](https://html.spec.whatwg.org/#dom-cva-checkvalidity)
 * [HTML spec for `reportValidity()`](https://html.spec.whatwg.org/#dom-cva-reportvalidity)
 * [HTML spec for `setCustomValidity()`](https://html.spec.whatwg.org/#dom-cva-setcustomvalidity)
+
+## Why not improve `<select>`
+
+Instead of creating a new `<selectmenu>` element, we could improve the existing select element to meet the same needs. However, there are a few problems with this (borrowed from the [Custom Controls UI Explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ControlUICustomization/explainer.md)):
+
+- Changing the parsing behavior for `<select>` to allow other element children by default would likely break a lot of old sites that depend on the current behavior.
+- The `<select>` popup currently has the capability to break outside the window of the browser or of an iframe. Sites may depend on this behavior, but we would need to remove it for security reasons if `<select>` is allowed to contain arbitrary content.
```
mfreed7, line 388: Akin to the previous bullet, we would need to modify this behavior also based on the existence of an opt-in attribute. We wouldn't be able to just remove the "break outside" capability for `<select>` across the board, that would not be web compatible.

I might phrase this slightly differently - I view this "break outside" capability as the super-power of the `<select>` element. When explaining to developers which element to choose, it's clearer if there are key differences in behavior. That story would go like this:
 - If you want full styleability and customizability, you need `<selectmenu>`. This element cannot, however, break outside the window boundary.
 - If you **need** your select picker to break outside the window boundary, you need `<select>`. This element cannot, however, be fully styled.

mfreed7: This could be added after your last bullet below, instead.

mfreed7 (approved):
This LGTM to land any time, and we can iterate with followup PRs if needed.

site/src/pages/components/selectmenu.mdx
```diff
@@ -27,6 +27,7 @@ import Image from '../../components/image.astro'
   - [`disabled`, `form`, `name`, `required`, and `labels`](#disabled-form-name-required-and-labels)
   - [`type`](#type)
   - [Validity IDLs](#validity-idls)
+- [Why not improve `<select>`](#improve-select)
```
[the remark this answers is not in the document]
josepharhar: done, thanks

site/src/pages/components/selectmenu.mdx
```diff
@@ -378,3 +379,12 @@ Like other form control elements, selectmenu supports validity checking attribut
 * [HTML spec for `checkValidity()`](https://html.spec.whatwg.org/#dom-cva-checkvalidity)
 * [HTML spec for `reportValidity()`](https://html.spec.whatwg.org/#dom-cva-reportvalidity)
 * [HTML spec for `setCustomValidity()`](https://html.spec.whatwg.org/#dom-cva-setcustomvalidity)
+
+## Why not improve `<select>`
+
+Instead of creating a new `<selectmenu>` element, we could improve the existing select element to meet the same needs. However, there are a few problems with this (borrowed from the [Custom Controls UI Explainer](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/ControlUICustomization/explainer.md)):
```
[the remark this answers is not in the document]
josepharhar: done, thanks

site/src/pages/components/selectmenu.mdx
```diff
@@ -378,3 +379,12 @@ Like other form control elements, selectmenu supports validity checking attribut
...
+
+- Changing the parsing behavior for `<select>` to allow other element children by default would likely break a lot of old sites that depend on the current behavior.
```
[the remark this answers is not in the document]
josepharhar: done

site/src/pages/components/selectmenu.mdx
josepharhar: Ok, I rewrote this bullet point with your thoughts in mind. what do you think?

PR #785, opened 2023-07-29
Switch proposal

gfellerph:
This is a first draft for a proposal for standardising the `<switch>` component, submitted by recommendation of @gregwhitworth.

There is a parallel effort on standardising the switch component at https://github.com/whatwg/html/pull/9546.


site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,448 @@
...
+### Accessibility
+
+{/* - A labelling text can be assigned to the switch like any other form control through a `<label>` element, the `aria-label=""` attribute or the `aria-labelledby=""` attribute */}
+{/* - Should be assigned the implicit aria role of "switch" */}
+{/* - The current value should be reported as "on" or "off" to screen readers */}
+
+ARIA authoring practices guide for the switch pattern: https://www.w3.org/WAI/ARIA/apg/patterns/switch/
+
+<div class="question">
+  How can the `<switch>` element be assigned the implicit role of "switch"?
```
scottaohara, line 344: if it became an actual element - it would be defined in HTML AAM. and HTML AAM would point to the Core AAM definition for the switch role mappings, here - https://w3c.github.io/core-aam/#role-map-switch

site/src/pages/components/switch.explainer.mdx
```diff
@@ -334,15 +338,9 @@ Limitations:
 
 ### Accessibility
 
-{/* - A labelling text can be assigned to the switch like any other form control through a `<label>` element, the `aria-label=""` attribute or the `aria-labelledby=""` attribute */}
-{/* - Should be assigned the implicit aria role of "switch" */}
-{/* - The current value should be reported as "on" or "off" to screen readers */}
+The control can be in a checked or unchecked state. Exposure of states will be defined by AT. [^6]
```
scottaohara, line 341: Sorry, a bit of a misunderstanding here. Browsers expose the state - screen readers decide how then to “announce” the state to their users. You can drop the second sentence here

gfellerph: No worries! I was not sure how detailed this needs to be defined anyways. I'm happy to get feedback on this.

josepharhar:
Thanks for writing this up!
I'd like to see more use cases and examples to justify the track, trackslide, and thumb elements.
If it turns out that we can do everything we want without the need for those elements, then maybe it would be possible to build it on the newly proposed switch which extends the input element: https://github.com/whatwg/html/pull/9546
Otherwise, we would need to create a separate element.

I agree with mason that we can merge this now regardless

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,438 @@
...
+
+#### Properties and Attributes
+
+<div class="question">
+  If form associated, should default value of "on" be submitted if switch is on and the user did not
+  specify a value to align with the [checkbox default value
+  behavior](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value:~:text=If%20the%20value%20attribute%20was%20omitted%2C%20the%20default%20value%20for%20the%20checkbox%20is%20on%2C%20so%20the%20submitted%20data%20in%20that%20case%20would%20be%20subscribe%3Don.)?
+</div>
+
+<div class="question">
```
gregwhitworth, line 169: Can you please file issues for the open questions

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,438 @@
...
+</div>
+
+Animation variant where the thumb and the track move together to reveal the hidden side of the track. Any text displayed on the switch is optional and defined by the author through a slot or other means.
+
+![Switch anatomy with moving track](../images/switch-moving-track-anatomy.png)
+_Source: [Figma source of switch anatomy with moving track](https://www.figma.com/file/uAYj6yVcAGD2OgjwCTqRol/Switch?type=design&node-id=2%3A28&mode=design&t=OAr7EJBXAQ4AEwfI-1)_
+
+Limitations:
+
+- The thumb needs to be at least the height of the track in order to hide the boundary between active and inactive track side underneath.
```
gregwhitworth, line 223: I don't think this should be defined; why is a requirement/limitation? This is just a block within another block. A developer should be able to make them smaller, larger, etc than the track. The parts existence; is irrelevant to its visual capabilities. I'll note I may be mis-understanding your point here.

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,438 @@
...
+- It's unclear how to style the track and slider if the track content can be of arbitrary length.
```
gregwhitworth, line 224: Open an issue for the default user-agent styles of the form control as a quick idea to a solution of this would be a `minmax(min-content, <whatever the user-agent defines>`) for example.

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,448 @@
...
+  Should the thumb width be dinamically adapted to the largest track content width, if any? This
+  woudl ensure that the thumb is always able to hide active and inactive track content.
+</div>
+
+![Switch anatomy with static track](../images/switch-static-track-anatomy.png)
+_Source: [Figma source of switch anatomy with static track](https://www.figma.com/file/uAYj6yVcAGD2OgjwCTqRol/Switch?type=design&node-id=2%3A28&mode=design&t=OAr7EJBXAQ4AEwfI-1)_
+
+Limitations:
+
+- The content on both sides cannot be wider than the tumb. Otherwise, the thumb is no longer capable of hiding the content of the inactive track content.
```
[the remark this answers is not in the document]
gregwhitworth: @scottaohara IMO, this would go like so:

`<switch>On</switch>`

That said @gfellerph this highlights that you may actually want two parts for on/off. The browser will then handle styling when they are toggled. So something like this:

```
<switch>
      # --- UA Shadow <label>
            <toggled>On</toggledlabe>
            <untoggled>Off</untoggled>
      </label>
</switch>
```

I hit this same issue when sketching out `<checkbox>` as you want not only localization but also people will want to put SVGs and other content within the different states declaratively.

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,438 @@
...
+</div>
+
+Animation variant where only the thumb moves and the background of the switch changes color only. Any text displayed on the switch is optional and defined by the author through a slot or other means.
+
+![Switch anatomy with static track](../images/switch-static-track-anatomy.png)
+_Source: [Figma source of switch anatomy with static track](https://www.figma.com/file/uAYj6yVcAGD2OgjwCTqRol/Switch?type=design&node-id=2%3A28&mode=design&t=OAr7EJBXAQ4AEwfI-1)_
+
+Limitations:
+
+- The content on both sides cannot be wider than the tumb. Otherwise, the thumb is no longer capable of hiding the content of the inactive track content.
```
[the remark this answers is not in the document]
gregwhitworth: @gfellerph I would like an issue on the various naming and whether we want alignment with `<range>` since there is a lot of overlap

gregwhitworth (changes requested):
The image for the design is not rendering in the [deployment](https://deploy-preview-785--open-ui.netlify.app/components/switch.explainer/), unsure what the issue is.

site/src/pages/components/switch.explainer.mdx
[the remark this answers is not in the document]
gfellerph: I opened https://github.com/openui/open-ui/issues/804 for discussing the naming.

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,438 @@
...
+
+<div class="question">
+  If the track content on either side can be of arbitrary length, the track width needs to be `2 *
+  longest side` and the switch needs to be `track width / 2` (not accounting for thumb width,
+  paddings and borders). Is it possible to style this?
+</div>
+
+Animation variant where the thumb and the track move together to reveal the hidden side of the track. Any text displayed on the switch is optional and defined by the author through a slot or other means.
+
+![Switch anatomy with moving track](../images/switch-moving-track-anatomy.png)
```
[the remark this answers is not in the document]
gfellerph: Replaced the relative with an absolute path

gfellerph:
> Thanks for writing this up! I'd like to see more use cases and examples to justify the track, trackslide, and thumb elements. If it turns out that we can do everything we want without the need for those elements, then maybe it would be possible to build it on the newly proposed switch which extends the input element: [whatwg/html#9546](https://github.com/whatwg/html/pull/9546) Otherwise, we would need to create a separate element.
> 
> I agree with mason that we can merge this now regardless

I updated the explainer with [showcase videos](https://deploy-preview-785--open-ui.netlify.app/components/switch.explainer/#prior-artexamples) from switches with content on both track and thumb. Also, I analyzed different Design Systems to find out how widespread slotted content is ([analysis of slotted content on switches](https://deploy-preview-785--open-ui.netlify.app/components/switch.explainer/#slots)). These examples would totally be possible on an `<input type="checkbox" switch />`, using background images and `content` on pseudo elements.

But if you inspect for example the [antd switch](https://ant.design/components/switch), they used `<span>` to declare track content. Not being able to slot content into track and thumb would at least be inconvenient. If you would like to get a little more creative with JavaScript controlled animation on slotted svgs (like Adam Argyle in https://web.dev/building-a-theme-switch-component/ - not a switch but this could be on the thumb), it gets really tricky with just background images.

gregwhitworth:
@gfellerph after this week's call, there are just some issues that need to be opened and addition of usecases assuming you landed  @scottaohara feedback. Please do the above but marking as approved to not slow down progress.

gregwhitworth (approved):

site/src/pages/components/switch.explainer.mdx
gfellerph: image

This highlights the issue a bit better. In order for the animation to work, the track sides need to have the same width. One side is dependent on the others width and I was not sure how this would work.

But probably it is already too detailed for this stage, therefore I think I'm going to remove the passage.

site/src/pages/components/switch.explainer.mdx
gfellerph: @gregwhitworth, I think I'll open issues sequentially, not all at the same time, because some questions are depending on decisions from other questions.

gfellerph:
@gregwhitworth, @mfreed7, I marked this PR for ready for review (no longer a draft) with the updates in 5710d87. You already approved it, but I don't have write access on the repository. If you think this PR is ready now, please go ahead and merge it. I'll open issues for the questions and will continuously update the explainer.

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
+| `name`        | `name`         | `string`  | `""`          | The name of the component.                                                                                                                                                                                                     |
+| `id`          | `id`           | `string`  | `""`          | The unique identifier of the component.                                                                                                                                                                                        |
+| `type`        | `type`         | `string`  | `button`      | Similar to the `<button>` or `<input>`, the type property for the switch can take either "button" or "submit". "Button" is used for triggering JavaScript actions, "submit" is used to immediately submit the associated form. |
+| `readonly`    | `readonly`     | `boolean` | `false`       | Indicates that the switch is not interactive but its value should still be submitted with the form.                                                                                                                            |
```
scottaohara, line 335: in html, readonly isn't allowed on checkboxes - so it'd be odd to have that here.

This table is missing a reference to `disabled` which _should_ be listed here and is allowed on checkboxes / other form elements that don't support readonly

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
+| `disabled`  | `true`/`false`    | `false`       | When `true`, disables the control, preventing user interaction and displaying the control with a disabled appearance. | No interaction. Controlled programmatically.                                                                |
+| `focused`   | `true`/`false`    | `false`       |                                                                                                                       |                                                                                                             |
+
+### Accessibility
+
+The control can be in a checked or unchecked state. [^6]
+
+The `<switch>` element can be assigned the default role of switch as described in the [Core Accessibility API Mappings](https://w3c.github.io/core-aam/#role-map-switch). [^7]
+
+Content for thumb and either track side can be defined by the author.
```
scottaohara, line 500: since this is the accessibility section, it's important to note that the content defined in the switch will not likely be exposed without rethinking how this content can be mapped.  Yet another reason, aside from the potential for the awkward sizing of the switch, why i think it's a bad idea to put text content within the switch parts.

The switch should be named by its label (so this should also be calling out it would need to be a labellable element).  The content of the swtich, if any, really should just be on/off/a localized version of these values - and then it can just be safely ignored since the checked state will expose it as either being 'on' or 'off' via the a11y api.

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
+
+<div class="question">Feedback on this feature would be welcome.</div>
+
+For a more traditional, non-JavaScript approach, it's necessary to receive the state of the switch on the server. In order to facilitate the directness of the action, a switch can be of type "submit". Like a submit button, toggling the switch will immediately submit the associated form to the server. This enables creating a HTML/CSS only settings page while staying true to user expectations of how the component should behave.
+
+#### Use with Assistive Technology
+
+{/* _Are there any details about shadow dom, focus delegation, or aria attributes that need special attention?_ */}
+
+The switch element can be associated with the existing `role="switch"`.
```
scottaohara, line 533: rather than commenting on the type bits above, i'll just repeat here that while a switch generally performs an immediate action due to the scripting associated with it, it is not really thought of as a type of submit button, and i've already outlined in a previous comment how submit buttons / form submissions are expected to behave - which does not match the expected UX of a switch.

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
+No security implications are expected.
+
+## Related issues
+
+- [switch: What customizability for animations should be supported?](https://github.com/openui/open-ui/issues/816)
+- [[switch]: Naming of the component and it's parts](https://github.com/openui/open-ui/issues/804)
+
+## Resources
+
+{/* _Any related resource links such as web standards, discussion threads, diagrams, etc._ */}
```
scottaohara, line 554: there should be links to thinks like the ARIA Authoring practices switch component page -https://www.w3.org/WAI/ARIA/apg/patterns/switch

arguably the switch definition in the aria spec.

there are a number of a11y-related switch articles / demos that exist as well which could be linked to, but i think just linking to the ARIA resources would probably be good enough

scottaohara (commented):
some suggested changes / missing bits and questions about some stuff i'm wondering about / have concerns without better understanding

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
+- `drag&drop`: the switch can be toggled by dragging the thumb
+
+#### Form association
+
+<div class="question">
+  How should form validation work? Is there a way to describe that the valid state is "off" instead
+  of the more intuitive "on"?
+</div>
+
+A `<switch>` can be part of a form and has a mechanism (`type="submit"`) to immediately submit the form it's part of. The value of a switch is reported to the form target as either "on" or "off".
```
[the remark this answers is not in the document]
gfellerph: It's an attempt to provide this idea of "switches trigger an immediate action" to non-javascript use cases. A simple example could be toggling dark mode server side on a html only page

```html
<form action="/index.html?mode=dark">
  <label for="dark-mode">Turn dark mode on</label>
  <switch id="dark-mode" type="submit"></switch>
</form>
```

I don't know if this is a good idea, it's something that I'd open a ticket for so it can be discussed.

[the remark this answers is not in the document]
keithamus: Auto-submission is an anti-pattern IMO. I don't think we want to encourage it. If you think it is necessary please do raise an issue, but I would object to this for the reasons Scott has already mentioned.

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
+## Design
+
+### User interaction
+
+From a user standpoint, the following interactions should be supported
+
+- Click: toggles the switch
+- Drag & drop: toggles the switch in the direction of the movement, the thumb follows the pointer
+- Swipe: toggles the switch in the direction of the swiping movement, the thumb follows the touch interaction
+- <kbd>Spacebar</kbd>: toggles the switch which has keyboard focus
```
keithamus, line 305: I think `Enter` should also work here.

site/src/pages/components/switch.explainer.mdx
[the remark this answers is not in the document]
scottaohara: Thanks for the reply. But yeh, agree with Keith. Tbh, in a no js scenario I’d rather someone used a noscript that contained a form consisting of a checkbox and a submit button.   Don’t blur the lines with switch acting like a checkbox and a submit where type=submit could be used with or without JS

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
+
+### Input type checkbox with switch attribute
+
+There is a [switch proposal at WHATWG](https://github.com/whatwg/html/pull/9546) (also see the original issue at https://github.com/whatwg/html/issues/4180) that would re-use the input element (`<input type="checkbox" switch>`) to implement a switch component. This explainer here is focusing on a different approach, using a standalone `<switch>` element. See the [comparison section](#comparison) for a list of differences between the switch attribute proposal and the switch element explained here.
+
+### Background
+
+{/* _Relevant historical or background information, related existing issues, etc._ */}
+
+The history of light switches date back to the 80’s, and they were used for just the On-Off purpose. The technological advancements of the day allow performing numerous tasks with the help of these light switches, including the adjustment of brightness of lamps.
```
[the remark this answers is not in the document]
gfellerph: I guess so, but the original source says just 80's (https://www.behance.net/gallery/70302039/History-Of-Toggle-UIUX-Case-Study-UI-Samples). Would you update it?

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
+
+The first ever light switch worked on a technology called the quick-break mechanism. Invented by John Henry Holmes in 1884, there used to be formation of a residue in the previous arcing technologies, thus reducing the life of the switch.
+
+The word "toggle" is a reference to a kind of mechanism or joint consisting of two arms, which are almost in line with each other, connected with an elbow-like pivot. However, the phrase "toggle switch" is applied to a switch with a short handle. [^2] <sup>and</sup> [^3]
+
+<div style="max-width: 400px; margin-inline: auto">
+  ![Internal components of a toggle
+  switch](https://upload.wikimedia.org/wikipedia/commons/thumb/3/35/Light_switch_inside_explained.jpg/440px-Light_switch_inside_explained.jpg)
```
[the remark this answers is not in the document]
gfellerph: No, thanks for the heads up.

site/src/pages/components/switch.explainer.mdx
gfellerph: I did not include it because it does not work on the checkbox either. Do you want to add it anyways? I initially thought the enter key would toggle a checkbox. A quick google search did not give me the "why" it does not work, there are many people asking how to do that with JavaScript, so it's definitely something that people are building on top of the standard spacebar.

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
+
+<div class="question">
+  If form associated, should default value of "on" be submitted if switch is on and the user did not
+  specify a value to align with the [checkbox default value
+  behavior](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value:~:text=If%20the%20value%20attribute%20was%20omitted%2C%20the%20default%20value%20for%20the%20checkbox%20is%20on%2C%20so%20the%20submitted%20data%20in%20that%20case%20would%20be%20subscribe%3Don.)?
+</div>
+
+<div class="question">
+  Should a custom "activate initially" attribute be used with the switch (`toggled`) or should the checkbox `checked` attribute be used as a standard?
+  `<switch toggled>` seems more natural while `<switch checked>` is already an established standard
```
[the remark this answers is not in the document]
gfellerph: I changed explainer to use `checked`.

site/src/pages/components/switch.explainer.mdx
gfellerph: I changed the section title to "edge cases" and updated the sample images to point out a possible implementation issue, which is closer to what I initially wanted to point out here.

site/src/pages/components/switch.explainer.mdx
scottaohara: if the checkbox were used within a form, Enter would cause it to submit the form.  if a switch can be form associated, it too should mirror checkbox and other form controls that do not use enter to activate/toggle, but instead submit the form.

if a switch is not form associated, then arguably it could work with either space enter or both.  APG notes that switch operates with Spacebar and optionally with Enter as well.

site/src/pages/components/switch.explainer.mdx
gfellerph: I opened https://github.com/openui/open-ui/issues/959 to gather some ideas or solutions. The more I think about it, the more I think it's not a good idea to add content, except for images. A checkmark on an "on" switch could help communicate that it's "on" without just using color while avoiding most of the issues. I'm just not sure how that could be implemented.

lukewarlow:
Is this ready to land as an initial explainer and then add to it in follow ups?

lukewarlow:
That makes sense. I'm able to merge so once you're happy with it give me a ping and I can get it landed :)

site/src/pages/components/switch.explainer.mdx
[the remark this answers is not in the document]
gfellerph: Added this to the anatomy section, thanks for the input.

gfellerph:
> Thanks for writing this up! I'd like to see more use cases and examples to justify the track, trackslide, and thumb elements. If it turns out that we can do everything we want without the need for those elements, then maybe it would be possible to build it on the newly proposed switch which extends the input element: [whatwg/html#9546](https://github.com/whatwg/html/pull/9546) Otherwise, we would need to create a separate element.
> 
> I agree with mason that we can merge this now regardless

@josepharhar I added a section that compares the two proposals (https://deploy-preview-785--open-ui.netlify.app/components/switch.explainer/#comparison-switch-attribute-vs-switch-element).

site/src/pages/components/switch.explainer.mdx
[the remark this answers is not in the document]
gfellerph: I removed the part about the switch being able to submit forms.

site/src/pages/components/switch.explainer.mdx
gfellerph: I removed readonly and added disabled.

site/src/pages/components/switch.explainer.mdx
gfellerph: Resolved on #979 and #978.

site/src/pages/components/switch.explainer.mdx
gfellerph: Added the links for the APG switch as well as the aria spec for `role="switch"`.

site/src/pages/components/switch.explainer.mdx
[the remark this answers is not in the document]
gfellerph: I'll leave it like in the original

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
+- It requires an immediate action, no confirmation is needed for the state to take effect
+
+### Features
+
+{/* _A list of the key features unique to this component._ */}
+
+#### Interaction
+
+- `click`/`touch`: the switch can be toggled by clicking it
+- `drag&drop`: the switch can be toggled by dragging the thumb
```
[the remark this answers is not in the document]
gfellerph: I added space, I'll not add enter because of https://github.com/openui/open-ui/pull/785#discussion_r1404657511

site/src/pages/components/switch.explainer.mdx
gfellerph: Removed that section from the explainer

site/src/pages/components/switch.explainer.mdx
```diff
@@ -0,0 +1,574 @@
...
```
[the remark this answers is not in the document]
gfellerph: I removed that interaction for now and added a question to the explainer and will open an issue so it can be discussed.

mfreed7:
Seems like this proposal should get landed. Nothing stops further iteration on the explainer, but with it here in PR form, nobody can see it.

mfreed7:
> Seems like this proposal should get landed. Nothing stops further iteration on the explainer, but with it here in PR form, nobody can see it.

@gfellerph I'm happy to click the button to do that - just give me the go-ahead.

Issue #702, opened 2023-03-31
[select] Don't reuse slot="" and ::part(); behavior="" is also strange


domenic:
slot="" and ::part() are used for custom elements. They are a separate namespace for authors to use.

Native HTML elements get customized via normal child elements and pseudo-elements:

- `<summary>` for `<details>`, `<option>` and `<optgroup>` for `<select>`, `<legend>` for `<fieldset>`, ...
- `::marker` for `<details>`, `::backdrop` for `<dialog>`, `::placeholder` for `<input>`, etc. (This is maybe related to #645.)

Similarly, your use of a semi-generic `behavior=""` attribute, which only has meaning within a certain parent context but can be used on any arbitrary element, is not idiomatic to HTML either. Is there any way to replace that with an idiomatic child-element pattern?

josepharhar:
Thanks for your feedback! I see what you mean, especially with your details and summary example.

What if we replaced all of the slot and behavior attributes with new elements?
- `<div popover=auto slot=listbox behavior=listbox>` would become `<listbox>`
- `<button slot=button behavior=button>` would become `<selectmenubutton>` or something. I don't think we can necessarily just use `<button>` because you might want to use `<button>` elements in other parts of the selectmenu.
- `<span slot=selected-value behavior=selected-value>` would become `<selectedvalue>`
- `<span slot=marker behavior=marker>` would become `<marker>`, or maybe we could replace it entirely with a pseudo-element...?

This would align better with the HTML spec, but may also reduce the customizability of selectmenu, for example by not providing a way to opt out of behaviors - not sure how important this actually is for customizability though.

I'm likely missing historical context on this. I'd like to talk about it at the next meeting

domenic:
> What if we replaced all of the slot and behavior attributes with new elements?

I think this would be ideal. I look forward to hearing back what others say at the meeting, but I think this should be the starting position, and any deviations from it should be considered quite carefully. (It feels like instead the current starting position was some sort of custom element library.)

gregwhitworth:
> This would align better with the HTML spec, but may also reduce the customizability of selectmenu

I'll need to think about this a bit and run through a variety of use cases to determine the implications of this recommendation as this is actually my key concern and why we landed where we did. Looping in @jh3y since he's probably built the most `selectmenu` demos and would be able to determine if this would break his customizations or not.

mfreed7:
> slot="" and ::part() are used for custom elements. They are a separate namespace for authors to use.

[...]

PR #798, opened 2023-08-15
Use new selectmenu anatomy in explainer

josepharhar:
This new anatomy was proposed and agreed on here:
https://github.com/openui/open-ui/issues/702

### DIFF
--- a/site/src/components/select-anatomy.jsx
+++ b/site/src/components/select-anatomy.jsx
@@ -6,13 +6,8 @@ const SelectAnatomy = () => {
     <AnatomyWrapper>
       <Host name="select">
         <Slot name="button">
-          <Part name="button">
-            <Slot name="selected-value">
-              <Part name="selected-value">Currently selected value</Part>
-            </Slot>
-            <Slot name="marker">
-              <Part name="marker">Dropdown indicator (e.g. icon)</Part>
-            </Slot>
+          <Part name="button type=selectlist">
+            <Part name="selectedvalue">Currently selected value</Part>
           </Part>
         </Slot>
         <Slot name="listbox">
--- a/site/src/pages/components/selectlist.mdx
+++ b/site/src/pages/components/selectlist.mdx
@@ -67,8 +67,6 @@ Selectlist supports a variety of pseudo-elements to target its different parts:
 
 Each of these pseudo-elements will target the default part if the author doesn't provide their own replacement. If the author provides their own, these pseudo-elements will target the author provided part.
 
-One way to style the control to match your requirements is to use the CSS `::part()` pseudo-element to select the different parts within the control's anatomy that you wish to style.
-
 ```html
 <style>
   .my-select-list::button {
--- a/site/src/pages/components/selectlist.mdx
+++ b/site/src/pages/components/selectlist.mdx
@@ -65,6 +65,8 @@ Selectlist supports a variety of pseudo-elements to target its different parts:
 - `::marker` - Targets the dropdown arrow of the default supplied button.
 - `::selectedvalue` - Targets the `<selectedvalue>` element.
 
+Each of these pseudo-elements will target the default part if the author doesn't provide their own replacement. If the author provides their own, these pseudo-elements will target the author provided part.
+
 One way to style the control to match your requirements is to use the CSS `::part()` pseudo-element to select the different parts within the control's anatomy that you wish to style.
 
 ```html
--- a/site/src/pages/components/selectlist.mdx
+++ b/site/src/pages/components/selectlist.mdx
@@ -46,10 +46,10 @@ Because the various parts of the selectlist can be styled, it's important to und
 <SelectAnatomy />
 
 - `<selectlist>` - The root element that contains the button and listbox.
-- `slot=button` - The button element that triggers the visibility of the listbox. If this part is not provided by the author, then selectlist will automatically create one.
-- `slot=selected-value` - The element that displays the value of the currently selected option. If this part is not provided by the author, then selectlist will automatically create one. Note that this part does not necessarily have to be placed inside the `slot=button` part (read on for more information about how to slot your own markup).
-- `slot=marker` - Something that indicates this button opens a listbox, e.g. an icon. If this part is not provided by the author, then selectlist will automatically create one.
-- `slot=listbox` - The wrapper that contains the `<option>`(s) and `<optgroup>`(s). If this part is not provided by the author, then selectlist will automatically create one.
+- button (slot) - The portion of the element which is always visible and in flow with the page, unlike the listbox. It should contain a button to open the listbox. If this part is not provided by the author, then selectlist will automatically create one. All child elements of the selectlist, except `<listbox>`, `<option>`s, and `<optgroup>`s will be slotted into this slot.
+- `<button type=selectlist>` - The button which opens the listbox when clicked. The `type=selectlist` attribute indicates to the browser that this button should open the listbox.
+- `<selectedvalue>` - The element which contains the text of the currently selected option. Every time that the user selects an option, the browser will replace the text content of this element with the text content of the selected option.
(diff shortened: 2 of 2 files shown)


josepharhar:
@mfreed7 want to review this?

site/src/pages/components/selectlist.mdx
```diff
@@ -46,10 +46,10 @@ Because the various parts of the selectlist can be styled, it's important to und
...
-- `slot=selected-value` - The element that displays the value of the currently selected option. If this part is not provided by the author, then selectlist will automatically create one. Note that this part does not necessarily have to be placed inside the `slot=button` part (read on for more information about how to slot your own markup).
-- `slot=marker` - Something that indicates this button opens a listbox, e.g. an icon. If this part is not provided by the author, then selectlist will automatically create one.
-- `slot=listbox` - The wrapper that contains the `<option>`(s) and `<optgroup>`(s). If this part is not provided by the author, then selectlist will automatically create one.
+- button (slot) - The portion of the element which is always visible and in flow with the page, unlike the listbox. It should contain a button to open the listbox. If this part is not provided by the author, then selectlist will automatically create one. All child elements of the selectlist, except `<listbox>`, `<option>`s, and `<optgroup>`s will be slotted into this slot.
```
gregwhitworth, line 49: @josepharhar I propose not using the term slot anywhere and this anatomy should not have any slots and thus not have the "toggle".

gregwhitworth: Additionally, I don't get this statement: 

> The portion of the element which is always visible and in flow with the page, unlike the listbox.

It may not be visible and it may not be in flow (eg: I should be able to position this out of flow). I think a more clear definition would be that this will be the element that invokes the listbox as a popover via the popover API or something to that affect

site/src/pages/components/selectlist.mdx
```diff
@@ -46,10 +46,10 @@ Because the various parts of the selectlist can be styled, it's important to und
...
+- `<button type=selectlist>` - The button which opens the listbox when clicked. The `type=selectlist` attribute indicates to the browser that this button should open the listbox.
+- `<selectedvalue>` - The element which contains the text of the currently selected option. Every time that the user selects an option, the browser will replace the text content of this element with the text content of the selected option.
+- `<listbox>` - The wrapper that contains the `<option>`(s) and `<optgroup>`(s). If this part was not provided by the author, then selectlist will automatically create one.
```
gregwhitworth, line 52: Wrap selectlist with code quotes

site/src/pages/components/selectlist.mdx
```diff
@@ -59,18 +59,24 @@ Selectlist provides a variety of tools to help with styling, including pseudo-se
 
 ### Styling parts of the control
 
+Selectlist supports a variety of pseudo-elements to target its different parts:
+- `::button` - Targets the button slot.
+- `::listbox` - Targets the `<listbox>`.
+- `::marker` - Targets the dropdown arrow of the default supplied button.
```
gregwhitworth, line 65: Are these psuedo elements only available if the author doesn't provide their own element? At which point I should be able to simply style them as elements, correct?

site/src/pages/components/selectlist.mdx
```diff
@@ -59,18 +59,24 @@ Selectlist provides a variety of tools to help with styling, including pseudo-se
...
+- `::selectedvalue` - Targets the `<selectedvalue>` element.
+
 One way to style the control to match your requirements is to use the CSS `::part()` pseudo-element to select the different parts within the control's anatomy that you wish to style.
```
gregwhitworth, line 68: This is no longer the case and should be removed

gregwhitworth (changes requested):
Following signoff from @mfreed7 can we also get @domenic and possibly Anne to give it a review?

site/src/components/select-anatomy.jsx
```diff
@@ -6,13 +6,8 @@ const SelectAnatomy = () => {
...
         <Slot name="button">
-          <Part name="button">
-            <Slot name="selected-value">
-              <Part name="selected-value">Currently selected value</Part>
-            </Slot>
-            <Slot name="marker">
-              <Part name="marker">Dropdown indicator (e.g. icon)</Part>
-            </Slot>
+          <Part name="button type=selectlist">
+            <Part name="selectedvalue">Currently selected value</Part>
```
lukewarlow, line 10: Should the part name here be selectedvalue or just value?

Reason I ask is virtually all inputs will have some form of "value" part. But `selectedvalue` only really makes sense for elements with multiple choices? Saying that most elements can be linked up to a datalist so maybe `selectedvalue` is fine for all of them?

site/src/pages/components/selectlist.mdx
josepharhar: > I propose not using the term slot anywhere and this anatomy should not have any slots and thus not have the "toggle".

When I talked to domenic earlier, he didn't sound very opposed to defining the user-agent shadowroot of selectlist in the HTML spec because we already sort of do so for the details element and mention "slots": https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements

For this reason, I feel comfortable going forward using this terminology until we encounter further issues. Is there any other reason you don't want to use the word slot?

> this will be the element that invokes the listbox as a popover

But this is exactly what `<button type=selectmenu>` is. I could get rid of the "button (slot)" section entirely, but I think its worth spelling out since we are explicitly choosing to have a separate button slot and button type=selectmenu in order to support split buttons

> It may not be visible and it may not be in flow (eg: I should be able to position this out of flow).

Good point, I rewrote the first sentence

site/src/pages/components/selectlist.mdx
josepharhar: done for the highlighted section

site/src/pages/components/selectlist.mdx
josepharhar: They should target whatever is rendered, whether thats the default/fallback content or the author supplied one. I added a sentence to clarify

site/src/pages/components/selectlist.mdx
josepharhar: whoops, good catch. i removed this sentence

site/src/components/select-anatomy.jsx
josepharhar: The new `<selectedvalue>` element is only intended to work for `<selectlist>`, at least for now, and it is an element with multiple choices.

Do you think the element's tag name should be `<value>` instead?

josepharhar:
> can we also get @domenic and possibly Anne to give it a review?

imo they should review it on the public site after this and my remaining patches are merged to the explainer

site/src/components/select-anatomy.jsx
lukewarlow: If it were to be value then it likely could be reused for other inputs in future?

josepharhar: Good point. I opened an issue to discuss this further: https://github.com/openui/open-ui/issues/808

site/src/pages/components/selectlist.mdx
mfreed7: +1 we should mention how this works via slots. @domenic ’s point was about the developer side of this - that you shouldn’t need to use the `slot` attribute. But I think it’s important (for interop, spec ease, and stylability) that we define the structure of the shadow root, and define the algorithm used to slot in content.

mfreed7:
> > can we also get @domenic and possibly Anne to give it a review?
> 
> imo they should review it on the public site after this and my remaining patches are merged to the explainer

+100 I really don’t like that we have so much review happening on these explainers. The feedback is awesome, and it’s not that I don’t want all of the great input! But the history and conversation on a PR is so much harder to follow later, as compared to an issue. I’d really prefer we do a very basic review on these and then land them. Fundamental (or even small) GitHub issues should then be raised to discuss changes, and the explainer gets updated after we resolve those issues. That way we have a paper trail for how we got to where we ended up.

mfreed7 (approved):

gregwhitworth (approved):

PR #805, opened 2023-08-19
Update file anatomy to not use slots

lukewarlow:
As #702 decided to not reuse slots this removes the slots from the file anatomy.

The file editors draft also needs updating to not refer to the slots.

### DIFF
--- a/site/src/components/file-anatomy.jsx
+++ b/site/src/components/file-anatomy.jsx
@@ -3,14 +3,10 @@ import { AnatomyWrapper, Host, Part, Slot } from './anatomy-components'
 
 const FileAnatomy = () => {
   return (
-    <AnatomyWrapper>
+    <AnatomyWrapper showToggle={false}>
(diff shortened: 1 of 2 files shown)


site/src/components/file-anatomy.jsx
```diff
@@ -3,14 +3,10 @@ import { AnatomyWrapper, Host, Part, Slot } from './anatomy-components'
...
-    <AnatomyWrapper>
+    <AnatomyWrapper showToggle={false}>
       <Host name="openui-file">
-        <Part name="file-selector-button">
-          <Slot name="button" />
-        </Part>
-        <Part name="label">
-          <Slot name="label" />
-        </Part>
+        <Part name="file-selector-button" />
```
lukewarlow, line 8: Should this be `button type=file-selector` to match selectlist? If so should anatomy show pseudos? Or should they be part of the explainer instead?

Issue #789, opened 2023-08-10
[process] Scribe documentation


una:
There is a series of commands for scribing a meeting. It would be useful to have a short `scribe.md` file in this repo to enumerate these commands and act as a guide, including:

- Starting a meeting (a set of commands to start, set title, etc.)
- Setting a new topic
- Posting the notes to the issue
- Ending a meeting

etc.

dbaron:
I think perhaps the shortest version is something like (where I'm using `${variable}` as the syntax for something that needs to be filled in):

* starting a meeting
  * required steps:
    1. `/invite Zakim`
    2. `Zakim, start meeting`
    3. `ScribeNick: ${username}`
  * recommended steps: add additional lines of metadata as described in "Step 2" of [scribe.perl's documentation](https://w3c.github.io/scribe2/scribedoc.html)
* new topic:
  * if it's logged to a github issue, do `/me github-bot, topic ${url}`
  * if it's not logged to github, or the above gives an error, use `Topic: ${title_of_topic}`
* ending meeting:
  1. `/me github-bot, end topic`
  2. `/me RRSAgent, draft minutes v2`

hidde:
> if it's logged to a github issue, do /me github-bot, topic ${url}

TIL! to confirm, is this: 

```
Topic: ${topic}
github: ${url} 
```

the same as this:

```
/me github-bot, topic ${url}
```

?

> /me RRSAgent, draft minutes v2

I believe v2 is the default now, so we could probably leave that off?

hidde:
Ok, have made a PR (https://github.com/openui/open-ui/pull/810) for this: 

- the “starting the meeting”/“ending the meeting”/Zakim bit was already covered in [chair-meeting.md](https://github.com/openui/open-ui/blob/main/meetings/telecon/chair-meeting.md)
- added “setting a topic” and posting notes to GitHub to both [meetings.md](https://github.com/openui/open-ui/blob/main/meetings/telecon/meetings.md) and [chair-meeting.md](https://github.com/openui/open-ui/blob/main/meetings/telecon/chair-meeting.md) (usually the chair sets the topic and adds the GitHub link but good to know the mechanics for those who scribe too)

PR #810, opened 2023-08-22
Improve scribe docs (re #789)

hidde:
this is to try and address #789


dbaron (commented):
Instead of always recommending separate `Topic:` and `github:` commands, I might bias towards recommending using `/me github-bot, topic <url>` for github issues which will let `github-bot` set the `Topic:` and also set the github issue, with a single command.

hidde:
Thanks, @dbaron, have updated!

gregwhitworth (approved):

PR #813, opened 2023-08-23
Fix Selectlist links

YummyBacon5:
The old Selectmenu page incorrectly links to the new Selectlist page
The Selectlist page has a few incorrect table of content links


mfreed7 (approved):
Thanks for the fixes!

PR #828, opened 2023-09-07
Add mention of legend element

### LINKED ISSUE #822: [Selectlist] Allow `<label>` as the first child of `<optgroup>`, as an alternative to the `label` attribute
# Issue

The only way to display a label with an `optgroup` element is with the `label` attribute.

But this would not allow for very much customisability. Since no other HTML elements would be shown, if induced with the attribute.

Also that there is currently [no way to style an `optgroup`'s label](https://open-ui.org/components/selectlist/#styling-parts-of-the-control).

Both of these issues would cause authors to create their own custom groups, which is what `selectlist` is supposed to fix.

# Possible Solution

Add a `::group-label` pseudo-element selector that targets the group's label.

Allow a `label` element to be the first child of an `optgroup` element. This would allow for authors to include their own elements.

Use cases for the latter:
Would allow authors to include more semantic meaning in a group's name. For example, an `abbr` element.
Would allow for better styling. For example, `span` elements to style individual words.

scottaohara:
- fixes a `type=selectmenu` typo to `selectlist` in last example
- adds `<legend>` element to list of select anatomy. 
- example of basic use of a `<legend>` element, noting it replaces the `label` attribute as a way to name an `<optgroup>`

closes #822


site/src/pages/components/selectlist.mdx
```diff
@@ -52,6 +52,7 @@ Because the various parts of the selectlist can be styled, it's important to und
 - `<selectedvalue>` - The element which contains the text of the currently selected option. Every time that the user selects an option, the browser will replace the text content of this element with the text content of the selected option.
 - `<listbox>` - The wrapper that contains the `<option>`(s) and `<optgroup>`(s). If this part was not provided by the author, then `<selectlist>` will automatically create one.
 - `<optgroup>` - Optional element which groups `<option>`(s) together with a label.
+- `<legend>` - Used to provide a label to an `<optgroup>` element. It must be the first child of the `<optgroup>`.
```
mfreed7, line 55: Must it be the first child? I know it's likely bad to have more than one of them, but I wonder if it's ok to put it anywhere you like?

mfreed7 (approved):
Very nice addition - thanks! Only one small comment, take it or leave it.

site/src/pages/components/selectlist.mdx
scottaohara: MUST it be from a browser implementation pov?  i mean, probably not?  i'll defer to you on that.  automagical stuff with details > summary, fieldset > legend, table > caption and how those are treated (re-positioned) if not the first items vs how they're rendered if not the first instance of the specific elements.

from an author / introduce the thing you're grouping with the group label, rather than randomly have that in the middle of your group - thus making it visually awkward to where the group beings POV... i think it's a pretty darn good author must, unless there's going to be automagical re-positioning of this element similar to the others?

YummyBacon5:
I think that the `<legend>` element should be included on selectlist's anatomy

site/src/pages/components/selectlist.mdx
scottaohara: from another angle, even for instances where maybe someone wants to be a bit more creative with how they're laying out their label for a group of options, shouldn't this be possible with modern CSS to visually reposition without needing to also futz with the actual DOM ordering?

mfreed7 (approved):

site/src/pages/components/selectlist.mdx
```diff
@@ -139,7 +140,28 @@ The default behavior of the `<selectlist>` control mimics the behavior of the `<
 </selectlist>
 ```
 
-When doing so, the default button, `<selectedoption>`, and `<listbox>` are created for you.
+When doing so, the default button, `<selectedvalue>`, and `<listbox>` are created for you. As with the current `<select>` element, the initial `<option>` will be returned as the `<selectedvalue>`.
```
lukewarlow, line 143: Isn't it `<selectoption>` rather than `<selectedvalue>`?

scottaohara: oh that's going to trip me up forever.  just like saying "hide" and "show" instead of "open" and "close". thank you

scottaohara, line 143:
```suggestion
When doing so, the default button, `<selectedoption>`, and `<listbox>` are created for you. As with the current `<select>` element, the initial `<option>` will be returned as the `<selectedoption>`.
```

PR #831, opened 2023-09-14
Add page 2 of the contributors on the homepage

brechtDR:
Updated the homepage to add the second page of the Open UI contributors on GitHub, as i noticed there are some people in there who are frequently involved with open-ui but not necessarily opening a lot of pull requests.

Screenshot:
Screenshot 2023-09-14 at 07 52 44

### DIFF
--- a/site/src/components/contributors.astro
+++ b/site/src/components/contributors.astro
@@ -1,12 +1,26 @@
 ---
-const CONTRIBUTORS_URL = `https://api.github.com/repos/openui/open-ui/contributors`
-const CONTRIBUTORS = await (await fetch(CONTRIBUTORS_URL)).json()
-// This is needed if call to CONTRIBUTORS_URL fails for some reasons.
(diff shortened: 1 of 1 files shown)


site/src/components/contributors.astro
```diff
@@ -1,12 +1,26 @@
...
+  const response = await fetch(`${CONTRIBUTORS_URL}?page=${page}`);
+  const contributors = await response.json();
+  CONTRIBUTORS.push(...contributors);
+};
+
+// Fetch contributors from page 1
+await fetchContributors(1);
+
+// Fetch contributors from page 2
+await fetchContributors(2); // Add more fetchContributors calls for additional pages as needed
```
gregwhitworth, line 16: do you happen to know if there is a page 3, is there any way to make it so that it fetches ALL of them and then we iterate over that?

gregwhitworth (approved):

Issue #808, opened 2023-08-19
Naming of the selected value element


josepharhar:
In [this explainer PR](https://github.com/openui/open-ui/pull/798), I added `<selectedvalue>` as the new element name which is intended to replace `behavior=selected-value` as per this issue: https://github.com/openui/open-ui/issues/702
The reason I chose `<selectedvalue>` is simply because it matches `behavior=selected-value`.

@lukewarlow pointed out that perhaps if we named it `<value>` it could be used in a wider variety of future use cases: https://github.com/openui/open-ui/pull/798#discussion_r1299176644

Anyone have thoughts about what it should be called?

josepharhar:
@una suggested that `<selectedoption>` would be a good name as well

josepharhar:
Yeah I suppose that `<selectedoptions>` would make more sense than `<selectedoption>` for multi-select.
I'm not actively working on multi-select yet and plan to do that after we get single-select done. The "select-v2" label for issues in this repo contain multi-select issues.

If you have any ideas for a term that would unambiguously be for singular or multiple options, I'm all ears.

jp-larose:
Why is this a tag, and not a property on the `<selectlist>` tag?  It could simply be the `value` property, matching the pattern of other form elements.

josepharhar:
> Why is this a tag, and not a property on the `<selectlist>` tag? It could simply be the `value` property, matching the pattern of other form elements.

I'm not sure I understand what the API shape would look like.
Here is an example of how selectedoption would be used:
```html
<selectlist>
  <button type=selectlist>
    <span>selected option:</span>
    <selectedoption></selectedoption>
  </button>
  <listbox>
    <option>one</option>
    <option>two</option>
  </listbox>
</selectlist>
```
How would you rewrite this without the selectedoption tag?

jp-larose:
> > Why is this a tag, and not a property on the `<selectlist>` tag? It could simply be the `value` property, matching the pattern of other form elements.
> 
> I'm not sure I understand what the API shape would look like. Here is an example of how selectedoption would be used:

[snip]

> How would you rewrite this without the selectedoption tag?

I guess my initial thought was that the form would simply submit text, in which case this would be sufficient:
```
<selectlist value="one">
  <button type=selectlist>
    <span>selected option:</span>
  </button>
  <listbox>
    <option>one</option>
    <option>two</option>
  </listbox>
</selectlist>
```

[...]

PR #833, opened 2023-09-14
Rename selectedvalue to selectedoption

josepharhar:
This is being discussed here:
https://github.com/openui/open-ui/issues/808

I renamed it to <selectedoption> for now in the chromium prototype, so the explainer should match.


mfreed7 (approved):
LGTM

PR #850, opened 2023-09-28
Remove lodash

lukewarlow:
This PR removes lodash from the site.

Most of what it's doing can be trivially replaced by vanilla JS.


gregwhitworth (approved):
There's a merge conflict that needs to be resolved; thanks for the update.

PR #851, opened 2023-09-28
Anatomy schema improvements

### LINKED ISSUE #845: Improve anatomy JSON Scheme and Display
Currently the anatomy that can be defined for a component in a design system is rather limited. Just an array of names with a description.

It's also displayed rather simply

image

It would be good to be able to represent the structure in an anatomy like we can in our anatomy proposals. It would also be good to display the anatomies in a way more akin to the concepts where the various design systems have their anatomies displayed and labelled with where the anatomy comes from.

This would make it far easier to turn these anatomies into a proposal with confidence that we cover X% of design systems.

lukewarlow:
Fixes #845 

Legacy Anatomies:

image

New Antomies:

image

### DIFF
--- a/site/src/components/anatomy.jsx
+++ b/site/src/components/anatomy.jsx
@@ -1,12 +1,29 @@
 import React from 'react'
 import './anatomy.css'
 
-import { anatomiesByComponent } from '../sources'
+import { componentsByName } from '../sources'
 
 const Anatomy = ({ component }) => {
-  const anatomy = anatomiesByComponent[component]
+  const anatomies = componentsByName[component]
+    .filter(
+      (component) =>
+        !!component.anatomy && (!Array.isArray(component.anatomy) || component.anatomy.length > 0),
+    )
+    .map((component) => {
+      if (Array.isArray(component.anatomy)) {
+        return {
+          sourceName: component.sourceName,
+          anatomy: component.anatomy,
+        }
+      }
 
-  if (Object.keys(anatomy).length === 0) {
+      return {
+        sourceName: component.sourceName,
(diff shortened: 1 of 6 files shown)


lukewarlow:
Blocked by #850

lukewarlow:
Follow up issue to work out the UI improvements https://github.com/openui/open-ui/issues/874

lukewarlow:
See https://github.com/openui/open-ui/pull/885 for usage of this new schema to see if there's any core requirements missing. For me it's the "slot" concept that I think does need to exist. It largely doesn't exist in browsers atm (aside from default slot) but it does get uses for `<selectlist>` and for `<details>`. I also imagine it's a much bigger part in other design systems. 

My gut feeling is that the current approach doesn't demonstrate this level of flexibility in the anatomies.

lukewarlow:
@gregwhitworth I've been looking into using that custom element manifest for our definitions and I think it's definitely viable but it would be more a thing of importing defintions of specific things and incorporating our own spin on it. (The manifest can't be used for nesting of structure as a wholesale thing for example).

Gonna require me to sit down and give it some thought how best to incorporate it all so for I think for now we should try and get this and the browser anatomies merged so we've got something we can use for now. Then I can make a follow up that adjusts it to utilitise the custom element manifest definitions where possible.

lukewarlow:
@gregwhitworth are you able to approve this when you've got the time. Then we can land the basic browser data as a follow up.

I will separately take a look into a slightly bigger refactor using the custom element schema but that will require a bit more work and I don't want to block having the browser data available.

PR #855, opened 2023-10-01
Remove eager selection of select option with arrow keys

### LINKED ISSUE #742: `<selectmenu>` eagerly makes selections
I believe the selection should be made when the option is chosen, not while it's just being perused:
1. would prevent the width jank as shown in the video
2. would prevent the `oninput` event from firing for each keyboard navigation
3. would align it's behavior to that of a mouse

<details>
<summary>see video</summary>

Kapture 2023-05-02 at 12 21 28

</details>

[codepen](https://codepen.io/argyleink/pen/wvYrZEV/ab71f1b613db0487f64ff8c6919b0173) for the above video

lukewarlow:
Fixes #742

### DIFF
--- a/site/src/pages/components/customizableselect.mdx
+++ b/site/src/pages/components/customizableselect.mdx
@@ -655,8 +655,8 @@ When the listbox is open:
 * Escape closes the listbox without changing the selected value.
 * Spacebar is used for typeahead.
 * Enter changes the selected value to the currently focused option and closes the listbox.
-* The up arrow key moves focus backwards in the list of options and changes the selected value to the newly selected option.
-* The down arrow key moves focus forwards in the list of options and changes the selected value to the newly selected option.
+* The up arrow key moves focus backwards in the list of options.
+* The down arrow key moves focus forwards in the list of options.
 * The left and right arrow keys do nothing.
 
 ## `multiple` and `size` attributes


scottaohara:
I agree that a change shouldn’t fire until the user has purposefully chosen an option, but with current select elements the displayed option in the triggering element does change. And that can be useful to keep.

mfreed7:
> I agree that a change shouldn’t fire until the user has purposefully chosen an option, but with current select elements the displayed option in the triggering element does change. And that can be useful to keep.

Does that typically happen for userland select menus? I feel (nothing more than a feeling) like I haven't seen that behavior anywhere but the builtin `<select>`. If so, perhaps we should think about what's right?

josepharhar:
> I agree that a change shouldn’t fire until the user has purposefully chosen an option, but with current select elements the displayed option in the triggering element does change. And that can be useful to keep.

We resolved in the issue in the commit message to not do this, right?

lukewarlow:
I feel like the preview should show the actually selected option. If it was to change with navigation but then disappear with escape (or clickoutside) I would find that confusing personally. But like mason that's not based on anything more than intuition and not having seen it in many custom implementations.

mfreed7:
> > I agree that a change shouldn’t fire until the user has purposefully chosen an option, but with current select elements the displayed option in the triggering element does change. And that can be useful to keep.

I think it's also important to point out that the above statement - that the displayed option in the triggering element **does** change - is a Windows-specific observation. It is not true, for example, on a Mac. For one, the popover covers the in-page element. But if you're careful, you can observe what's underneath it (e.g. via sharing your screen in a Meet, because that does not share the popover) and you can see that it does not change as you use arrow keys to change options in the popover.

Certainly a `change` event should not be fired until a new option is selected, however that's done.

josepharhar:
Yeah I thought we basically agreed to do the mac behavior. This is what I implemented in the chromium prototype if you look in canary: https://jsfiddle.net/xhsofq2z/4/

scottaohara:
> > I agree that a change shouldn’t fire until the user has purposefully chosen an option, but with current select elements the displayed option in the triggering element does change. And that can be useful to keep.
> 
> Does that typically happen for userland select menus? I feel (nothing more than a feeling) like I haven't seen that behavior anywhere but the builtin `<select>`. If so, perhaps we should think about what's right?

i'd submit you don't typically see this in most custom implementations similarly to why you don't see the ability to start typing chracters and have selection/focus move to the option that matches the typed characters - i'd submit it's an overlooked detail (which can be useful if the selected vs focused option has poor visual indicators) rather than a purposeful omission.

Regardless, I'm not going to strongly advocate that we do this.  I just didn't think that the decision to not have selection follow focus _had_ to mean this behavior was off the table too.  But if people don't see the value in it, then we can move on.

---

Separately though, @josepharhar, i noticed in [the demo](https://jsfiddle.net/xhsofq2z/4/) you posted some oddities.  rather than get into those here, i'll post a new issue

gregwhitworth:
So what is the next step to get this PR unblocked? @scottaohara said he won't push back here, the change aligns with the resolution in the issue. I'm going to approve and the issue can be re-opened if someone wants to change it.

gregwhitworth (approved):

gregwhitworth:
@josepharhar @lukewarlow friendly ping on this

lukewarlow:
I was concerned about https://github.com/openui/open-ui/issues/1087 and this contradicting each other but it seems that the desire is still to not do selection follow focus so this change is correct.

Correct me if I'm wrong @josepharhar ?

lukewarlow:
Given the issue I just mentioned though I think this probably does need some wider discussions. Specifically where do we draw the line between "platform convention" and actually having a good behaviour 😅 

Either way I think we should remove this text as it's not currently the plan and won't always be correct regardless.

tantek (approved):

josepharhar:
> I was concerned about #1087 and this contradicting each other but it seems that the desire is still to not do selection follow focus so this change is correct.
> 
> Correct me if I'm wrong @josepharhar ?

This is correct, I approve this PR

Issue #854, opened 2023-09-29
Is someone willing to pull together research on tooltip nestability?


mfreed7:
The [OpenUI research page for tooltips](https://open-ui.org/components/tooltip.research/) does not include information on whether tooltips are allowed to be nested. On the WHATWG html issue for `popover=hint`, there's an [open question](https://github.com/whatwg/html/issues/9776#issuecomment-1741574741) about whether tooltips are most commonly allowed to be nested, or whether they're restricted to be one tooltip at a time.

Does anyone have the time to take a look into the design systems listed on the research page (or bring in more systems) and see if they allow nesting?

Additionally, do any a11y folks have input on whether nested tooltips are "ok"? I have been assuming not, but I realize (from the WHATWG conversation) that that's just an assumption that I can't justify with any data. @scottaohara @aleventhal

brechtDR:
I'll try to kickstart this

scottaohara:
another quick note in how the nesting of tooltips can lead to issues.

a standard tooltip is generally expected to provide an accessible description for the content that invokes it.  

so `<button title=blerp>edit</button>` is commonly exposed as "edit, button. blerb".

Let's modify that to create a custom tooltip that produces the same output

```
<button aria-describedby=t>edit</button> 
<div role=tooltip id=t>blerb</div>
```

if we add nested tooltips to this
```
<button aria-describedby=t>edit</button> 
<div role=tooltip id=t>
  blerb
  <button aria-describedby=t2>term</button>
  <div role=tooltip id=t2>
     content here
     <button aria-describedby=t3>nest again</button>
     <div role=tooltip id=t3>more stuff here</div>
  </div>
</div>
```

the above results in the initial "edit" button being exposed as such:
>edit, button. blerb term content here nest again more stuff here

Well that sucks.

But, can't we update the browser to not expose the tooltips which haven't been rendered yet?

Sure - but then when they are rendered that would fire a description change event for each nested instance.  So the initial button's description would change because of the new content added into the tooltip, causing a re-announcement of that initial content and everything nested within (see previous quote) - and then any further nested tooltip would be re-announced.

_but wouldn't this be covered by converting these to rich tooltips_?

yes, if the association from the invoking element to the popover revised from a description to a details a11y relationship - then the previous problems go away.  but new issues get introduced:
[...]

PR #860, opened 2023-10-05
Tooltip: Add tooltip nesting research

brechtDR:
Based on the following issue: https://github.com/openui/open-ui/issues/854

I tried to kickstart this research by creating a table.
I don't have access to each of those systems, but by reading documentation and creating quick code demo's I came up with the following. 

If there are some people that can help, or update those that I wasn't able to test, that would be welcome.

Table of first commit:

| UI Library              | Can nested?        | Closes other tooltips?                         | Closes other Non-tooltip popovers? |
| ----------------------- | ------------------ | ---------------------------------------------- | ---------------------------------- |
| Ant Design              | No                 | Yes (default, but can be changed)              | No                                 |
| Atlaskit                | No                 | Yes                                            | No                                 |
| Bootstrap               | No                 | Yes, but stays open when clicked               | No                                 |
| Carbon design system    | No                 | Yes                                            | No                                 |
| Evergreen               | No                 | Yes                                            | No                                 |
| UI Fabrik               | No                 | Yes                                            | No                                 |
| FAST                    | Yes                | Yes (but not when nested, previous stays open) | No                                 |
| KoliBri                 | Unknown            | To be investigated                             | No                                 |
| Lightning Design System | To be investigated | To be investigated                             | No                                 |
| Material Components Web | Yes                | Yes                                            | No                                 |
| Semantic UI             | Yes                | Not the parent when nested                     | No                                 |
| Stardust UI             | No                 | can be prevented                               | No                                 |
| Boosted                 | No                 | Not when other tooltip is clicked              | No                                 |

### DIFF
--- a/site/src/pages/components/tooltip.research.mdx
+++ b/site/src/pages/components/tooltip.research.mdx
@@ -20,3 +20,21 @@ import Concepts from '../../components/concepts'
 ## Concepts
 
 <Concepts client:load component="Tooltip" />
+
+## Supports nesting
+
+| UI Library              | Can nested?        | Closes other tooltips?                         | Closes other Non-tooltip popovers? |
+| ----------------------- | ------------------ | ---------------------------------------------- | ---------------------------------- |
+| Ant Design              | No                 | Yes (default, but can be changed)              | No                                 |
+| Atlaskit                | No                 | Yes                                            | No                                 |
+| Bootstrap               | No                 | No                                             | No                                 |
+| Carbon design system    | No                 | Yes                                            | No                                 |
+| Evergreen               | No                 | Yes                                            | No                                 |
+| UI Fabrik               | No                 | Yes                                            | No                                 |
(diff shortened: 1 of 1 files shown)


brechtDR:
> Basically, I'd like to know whether this behavior is accidental or intentional. I think mentioning it in documentation shows that it's intentional, while just having it work that way might be accidental.

Yes, I will see what i can do, didn't keep it all while testing. my bad

brechtDR:
@mfreed7 I updated this a bit.

The ones that I wasn't able to do are set to "to be investigated", maybe you have an idea who we can tag for those, or how we continue from here?

When documentation only has textual content, I added this as info. Links to documentation or code examples where i could draw the conclusion were added to the library name.

All tooltips do not close other popovers that are not tooltips, so this was added as a general notice below the table.

mfreed7 (approved):

PR #864, opened 2023-10-06
add invokers explainer

keithamus:
This ports the Invokers explainer from https://github.com/keithamus/invoker-buttons-proposal/ to here.

Refs https://github.com/keithamus/invoker-buttons-proposal/issues/29

/cc @mfreed7

PR #866, opened 2023-10-06
Replace popovertargetaction=interest with interesttarget

lukewarlow:
Now that #864 is merged we can replace the old style `popovertargetaction="interest"` system for hover triggered popovers for the new `interesttarget` attribute.

### DIFF
--- a/site/src/pages/components/popover-hint.research.explainer.mdx
+++ b/site/src/pages/components/popover-hint.research.explainer.mdx
@@ -17,7 +17,7 @@ A followup proposal to the [original Popover proposal](https://open-ui.org/compo
 
 - [Background](#background)
 - ["Hover" triggering](#hover-triggering)
-  - [Add the `interest` value](#add-the-interest-value)
+  - [Interest triggering](#interest-triggering)
   - [CSS values to control delay](#css-values-to-control-delay)
   - [Example buttons](#example-buttons)
   - [Popover types](#popover-types)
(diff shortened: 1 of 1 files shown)


keithamus (approved):
👍 LGTM

site/src/pages/components/popover-hint.research.explainer.mdx
```diff
@@ -108,12 +110,12 @@ The new hover-triggering functionality should work for all `popover` types. Ther
 <button popovertarget=menu1>Menu</button>
 <ul popover id=menu1>
   <li>
-    <button popovertarget=submenu1 popovertargetaction=interest>
+    <button interesttarget=submenu1>
```
scottaohara, line 113: is this enough though?  for the previous one with the button and the hint, that makes sense because the hint is the focus/hover popover to the primary action of the button.

but here, a submenu button would probably also need the popovertarget or invokertarget attribute for click/tap opening

lukewarlow: So in terms of replacing the existing functionality I think what I've done is correct. It would previously trigger on hover (Etc) and it now triggers on hover (Etc).

As for whether this is desirable is another question?

I for one agree it probably should be invoke not interest? I also don't know if it should be both?

Potentially worth making as a follow up issue? (assuming I'm correct and it previously would only trigger for hover not both hover and click)

scottaohara: yah fair point about just updating what was there @lukewarlow 

i guess i kinda glossed over that with the original, assuming popovertarget pointing to the auto would work with click/tap/keypress, and just also happen to open on hover/focus with the popovertargetaction=hint.
but sure, new issue i suppose.

lukewarlow: Tangential but your comment made me think https://github.com/openui/open-ui/issues/872

lukewarlow:
Closing in favour of https://github.com/openui/open-ui/pull/899

Issue #797, opened 2023-08-14
Enable Astro "View Transitions" in the Open UI site


andrico1234:
https://docs.astro.build/en/guides/view-transitions/

Astro have recently released experimental support for view transitions. 

In the context of the Open UI site, enabling View Transitions would allow users to navigate across the site without a full page reload, which they call SPA mode (though I don't yet know exactly what the consequences of this kind of site navigation would have on the Open UI site as a whole).

The framework offers a few simple transitions out of the box, fallback for browser that don't support the View Transitions API, and a `prefers-reduced-motion` mode that prevents all animation if the setting is enabled.

What do you all think? I can get a push a small demo to the preview server and see how it fares against the current routing

gregwhitworth:
I'm not opposed to this but this needs someone to tackle it and until that point I'll close this.

PR #876, opened 2023-10-10
Upgrade to Astro v3

lukewarlow:
This upgrades the Astro framework to v3. This should lead to a much faster dev environment along with unlocking niceties such as built in view transitions should we want to use them (see #797 )

See https://docs.astro.build/en/guides/upgrade-to/v2/ for v1 to v2

See https://docs.astro.build/en/guides/upgrade-to/v3/ for v2 to v3

The only major thing this does is increase the minimum node requirement to node 18.

It also might change the shape of the build output not sure on that one.


site/src/pages/404.jsx
```diff
@@ -1,14 +0,0 @@
-import React from 'react'
```
andrico1234, line 1: Does Astro add its own default 404 page ?

site/src/pages/contribute.astro
```diff
@@ -0,0 +1,3 @@
+---
+return Astro.redirect('/get-involved');
```
andrico1234, line 2: Much nicer! Great spot

andrico1234:
Code wise, everything looks great! I'll give the preview deploy a test over the next few days. Thanks for this!

site/src/pages/404.jsx
lukewarlow: Astro no longer supports JSX components. And I checked and netlify provided its own 404 so I don't believe this one is ever seen

andrico1234 (approved):

andrico1234:
Thanks for this! I've approved. It looks like you need a Code Owner to approve as well

mfreed7:
> Thanks for this! I've approved. It looks like you need a Code Owner to approve as well

I'll approve based on this! (I'm no Astro expert.)

mfreed7 (approved):

PR #882, opened 2023-10-13
Add use cases to selectlist explainer

josepharhar:


josepharhar:
@andrico1234 @lukewarlow I'm not sure if I added CSS the right way, I feel like there's probably an actual way I'm supposed to do it instead of adding style attributes like I did in this PR.

@mfreed7 is this what you had in mind for use cases?

lukewarlow:
I *think* if you give them classes and then make a new sheet under site/styles astro might just handle it all magically for you. I could be wrong though.

mfreed7 (approved):

PR #887, opened 2023-10-15
Update selectlist anatomy

lukewarlow:
Update selectlist anatomy

- Change selectedvalue to selectedoption
- Add marker part
- Add optional legend part

### DIFF
--- a/site/src/components/select-anatomy.jsx
+++ b/site/src/components/select-anatomy.jsx
@@ -7,13 +7,15 @@ const SelectAnatomy = () => {
       <Host name="select">
         <Slot name="button">
           <Part name="button type=selectlist">
-            <Part name="selectedvalue">Currently selected value</Part>
+            <Part name="selectedoption">Currently selected option</Part>
+            <Part name="marker"></Part>
           </Part>
         </Slot>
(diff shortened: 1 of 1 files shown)


sudheer-gowrigari (approved):

site/src/components/select-anatomy.jsx
```diff
@@ -7,13 +7,15 @@ const SelectAnatomy = () => {
       <Host name="select">
         <Slot name="button">
           <Part name="button type=selectlist">
-            <Part name="selectedvalue">Currently selected value</Part>
+            <Part name="selectedoption">Currently selected option</Part>
+            <Part name="marker"></Part>
```
gregwhitworth, line 11: Not worth blocking on - but do you think we should represent these as pseudos? eg: `::marker`?

gregwhitworth (approved):

site/src/components/select-anatomy.jsx
lukewarlow: Yes the explainer itself mentions some of the pseudos. Hopefully the new anatomy schema will allow us to make clear what's a pseudo and what's an element proper.

PR #888, opened 2023-10-15
Correct file structure to use button element

lukewarlow:
Currently it's using divs but it should use the actual elements like checkbox does for its label.

### DIFF
--- a/site/src/pages/components/file.mdx
+++ b/site/src/pages/components/file.mdx
@@ -98,9 +98,9 @@ button invokes the file selection prompt.
   <button part="file-selector-button">
     <slot name="file-selector-button"></slot>
   </button>
-  <label part="label">
+  <span part="label">
     <slot name="label"></slot>
-  </label>
+  </span>
 </host>
 ```
 
--- a/site/src/pages/components/file.mdx
+++ b/site/src/pages/components/file.mdx
@@ -95,12 +95,12 @@ button invokes the file selection prompt.
 
 ```html
 <host>
-  <div part="file-selector-button">
+  <button part="file-selector-button">
     <slot name="file-selector-button"></slot>
-  </div>
-  <div part="label">
+  </button>
+  <label part="label">
     <slot name="label"></slot>
-  </div>
+  </label>
 </host>
 ```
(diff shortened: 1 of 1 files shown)


site/src/pages/components/file.mdx
```diff
@@ -95,12 +95,12 @@ button invokes the file selection prompt.
 
 ```html
 <host>
-  <div part="file-selector-button">
+  <button part="file-selector-button">
     <slot name="file-selector-button"></slot>
-  </div>
-  <div part="label">
+  </button>
+  <label part="label">
```
scottaohara, line 101: this seems odd.  is this really its 'label', or is it more a description of the current state of the control.  e.g., no file chosen, x file(s) chosen?  If that latter, this probably shouldn't be a label element then.

lukewarlow: As per https://gist.github.com/lukewarlow/2eeb75ba6d3039cd10d850c19162976b It's a `<label>` in Firefox. A `<span>` in Blink and not represented in the Dom structure in WebKit. 

Having said that I'm happy to change it to `<span>` (or leave it be for now) if that's considered more appropriate.

scottaohara: yeh, i don't really see the value in what firefox has there.  I'm actually rather [befuddled by chrome's implementation](https://github.com/w3c/html-aam/issues/421) - this actually reminded me i have bugs to file

lukewarlow: I've changed it to be a span to match the DOM structure of Chrome (and it's an inline element so it matches the actual rendering unlike a div)

lukewarlow:
It's worth noting that the slots section actually explicitly calls this out as a label element. @scottaohara should I change that to refer to a span?

image

scottaohara:
yeh, for the fallback content.  maybe the slot name should be updated to help clarify this is more of a prompt/description.  a label serves a different purpose, and this elemement _should_ be labelable as well.

lukewarlow:
Closing in favour of https://github.com/openui/open-ui/pull/913

PR #890, opened 2023-10-16
Fix select research to selectlist links

lukewarlow:
Fix select research to proposal link and add one the other way.

Also add a select component page to point to selectlist.

### DIFF
--- a/site/src/pages/components/select.mdx
+++ b/site/src/pages/components/select.mdx
@@ -1,0 +1,6 @@
+---
+name: Select Element (Explainer)
+layout: ../../layouts/ComponentLayout.astro
+---
+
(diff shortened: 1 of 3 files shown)


site/src/pages/components/select.mdx
```diff
@@ -0,0 +1,6 @@
+---
+name: Select Element (Explainer)
+layout: ../../layouts/ComponentLayout.astro
+---
+
+See selectlist explainer. [See the explainer here](./selectlist/).
```
lukewarlow, line 6: ideally this (and selectmeny) would just be a redirect but I don't know if that's easy to do till the Astro v3 PR gets merged.

PR #902, opened 2023-10-23
Improve selectlist use case UI on mobile

lukewarlow:
The images appear squashed an distorted on mobile currently. This fixes that.

| Before | After |
| ------  | ----- |
| image | image|


gregwhitworth (approved):
Those images are quite large on desktop. Not worth blocking PR but you might want to set a max-width or height.

lukewarlow:
I quite like the size of them on desktop? But I only have 1080p monitors. I'll merge it as is for now and we can maybe do a follow up if there is an issue

PR #912, opened 2023-10-27
Update invokeaction defaults for select and input

### LINKED ISSUE #911: [invokers] Should input and select auto action exist?
Currently the defaults for invokers on input and select are showPicker which is also the auto value.

image

~~While I was implementing this @josepharhar mentioned it might be better for this to be a toggle by default?~~

~~Having auto as a togglePicker would better match others autos too so I tend to agree with Joey.~~

~~I propose a new togglePicker and hidePicker action to go with showPicker and auto would map to togglePicker instead of showPicker.~~

After some thought and discussion I propose removing auto for now.

lukewarlow:
Remove auto action

Closes #911

### DIFF
--- a/site/src/pages/components/invokers.explainer.mdx
+++ b/site/src/pages/components/invokers.explainer.mdx
@@ -380,9 +380,7 @@ that this list is ordered and higher rules take precedence:
 | `<details>`     | `'toggle'`            | If the `<details>` is `open`, then close it, otherwise open it                       |
 | `<details>`     | `'open'`              | If the `<details>` is not `open`, then open it                                       |
 | `<details>`     | `'close'`             | If the `<details>` is `open`, then close it                                          |
-| `<select>`      | `'auto'`              | Open the `<select>` option picker UI                                                 |
-| `<select>`      | `'showPicker'`        | Open the `<select>` option picker UI                                                 |
-| `<input>`       | `'auto'`              | Call `.showPicker()` on the invokee                                                  |
+| `<select>`      | `'showPicker'`        | Call `.showPicker()` on the invokee                                                  |
 | `<input>`       | `'showPicker'`        | Call `.showPicker()` on the invokee                                                  |
 | `<video>`       | `'auto'`              | Toggle the `.playing` value                                                          |
 | `<video>`       | `'playpause'`         | Toggle the `.playing` value                                                          |


site/src/pages/components/invokers.explainer.mdx
```diff
@@ -380,10 +380,14 @@ that this list is ordered and higher rules take precedence:
...
 | `<details>`     | `'close'`             | If the `<details>` is `open`, then close it                                          |
-| `<select>`      | `'auto'`              | Open the `<select>` option picker UI                                                 |
-| `<select>`      | `'showPicker'`        | Open the `<select>` option picker UI                                                 |
-| `<input>`       | `'auto'`              | Call `.showPicker()` on the invokee                                                  |
-| `<input>`       | `'showPicker'`        | Call `.showPicker()` on the invokee                                                  |
+| `<select>`      | `'auto'`              | If the `<select>` is `open`, then close it, otherwise open it                        |
+| `<select>`      | `'togglePicker'`      | If the `<select>` is `open`, then close it, otherwise open it                        |
+| `<select>`      | `'showPicker'`        | If the `<select>` is not `open`, then open it                                        |
+| `<select>`      | `'closePicker'`       | If the `<select>` is `open`, then close it                                           |
```
lukewarlow, line 386: Do we need an explicit close? Most (all?) will close when clicked outside, aside from file input which is a blocking modal anyway?

I'm thinking add it for completeness?

keithamus (approved):
LGTM

site/src/pages/components/invokers.explainer.mdx
lukewarlow, line 386: Should this be closePicker or hidePicker?

It's hidePopover but close dialog and details?

lukewarlow: Give it's show and not openPicker I'm leaning towards this actually being hidePicker?

lukewarlow:
Hmm actually the more I think about this the less sure I am this makes sense.

The only popover=manual type picker is the color input in Firefox. Everything else closes on clickoutside OR is a modal dialog (file picker).

So does having a toggle and hide action actually make sense.

@keithamus what do you think about for now just removing the auto action? Then if we decide we want these in future (before shipping) we can add them easy enough?

keithamus (approved):

mfreed7:
> Hmm actually the more I think about this the less sure I am this makes sense.
> 
> The only popover=manual type picker is the color input in Firefox. Everything else closes on clickoutside OR is a modal dialog (file picker).
> 
> So does having a toggle and hide action actually make sense.
> 
> @keithamus what do you think about for now just removing the auto action? Then if we decide we want these in future (before shipping) we can add them easy enough?
> 
> Implementing the toggle is slightly difficult because the mousedown on the button leads to the picker closing which then reopens on mouseup from the invoke.

I think this (plus the conversation on the crbug) indicates that we need to include invoketarget invokers in the popover light dismiss calculation algorithm. Currently that algorithm exempts `<button popovertarget>` elements that invoked a popover. It feels like it needs to do the same for other similar things that use `invoketarget`.

lukewarlow:
Yup definitely.

It seems NearestTargetPopoverForInvoker is the culprit. I've added information to the invoke popovers chrome bug.

lukewarlow:
@mfreed7 no rush but when you're next taking a look through these PRs could you approve this one so the explainer matches the chrome implementation (and the draft spec https://github.com/whatwg/html/pull/9912)

mfreed7 (approved):
Certainly!

Issue #558, opened 2022-06-29
[selectmenu] Accessibility of option labels


dandclark:
In some cases it will be useful for authors to include non-interactive content in a `<selectmenu>`'s listbox for the purpose of labeling groups of options. The native `<select>` offers this today with `<optgroup>` and its `label` attribute: [MDN `<optgroup>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup).
 
![image](https://user-images.githubusercontent.com/1688716/176494324-adc9689d-330b-4b56-8091-fa0381e211bf.png)

It is very likely that developers will want to add optgroup labels like this with `<selectmenu>`, and they will want them to be styleable. To meet the goal of accessibility-by-default, these labels must be accessible without depending on devs to write ARIA.
 
The `<optgroup label="…">` approach used by `<select>` won't work because there is no way to target styles to these.
 
One way of approaching this problem is to use the `<legend>`* element for optgroup captions in the following way:
 
```html
<selectmenu>
  <optgroup>
    <legend>Fruits</legend>
    <option>Apple</option>
    <option>Pear</option>
    <option>Banana</option>
  </optgroup>
</selectmenu>
```

The developer can target the `<legend>` element with CSS to style it in any way they choose.
 
The platform will apply the text content of the `<legend>` as the accessible label of the `<optgroup>`, so that the above example will be exposed to accessible tech as if it had the following ARIA:
 
```html
<selectmenu>
  <optgroup role="group" aria-labelledby="fruit-legend">
    <legend role="caption" id="fruit-legend">Fruits</legend>
    <option>Apple</option>
    <option>Pear</option>
    <option>Banana</option>
  </optgroup>
</selectmenu>
```
 
"Error" cases would be handled in the following ways:
- If an `<optgroup>` is missing a `<legend>` element, this will result in a console warning. As an intervention in this case, the platform applies any non-interactive** text content inside an `<optgroup>` before the first `<option>` element as the `<optgroup>` label.
- If a `<legend>` is found in the listbox outside any `<optgroup>`, this will result in a console warning.
- If text content** is found in the listbox outside any `<optgroup>` or `<option>`, this will result in a console warning.
- An `<optgroup>` inside another `<optgroup>` emits a console warning. In the future we could consider adding more useful support for these and removing the warning.
[...]

PR #918, opened 2023-10-30
Improve usecases in selectlist explainer

josepharhar:
In response to this feedback: https://github.com/whatwg/html/issues/9799#issuecomment-1775422449


site/src/pages/components/selectlist.mdx
```diff
@@ -70,23 +70,23 @@ selectlist::button {
...
+This example adds "rich content" inside option elements in the form of images next to the options. This is a common pattern on the web which isn't possible in the existing `<select>` element because `<select>`'s `<option>`s only render text content.
 
 <div className="code-image-container">
 <div>
 ```html
 <selectlist>
   <option>
-    <img src="usa.jpg">
-    USA
+    <img src="swallow.jpg">
```
mfreed7, line 80: Do these new images exist in the repo? I don't see them but maybe I missed it.

site/src/pages/components/selectlist.mdx
```diff
@@ -165,18 +165,40 @@ This example has a `<listbox>` which enables it to put arbitrary content into th
...
-    <div>Heading</div>
-    <div>First two options:</div>
-    <option>one</option>
-    <option>two</option>
-    <div>Second two options:</div>
-    <option>three</option>
-    <option>four</option>
+    <div class=container>
+      <div>
+        <div role=group>1-2</div>
```
mfreed7, line 170: Is this a "good" example? I would think we should use `<optgroup>` instead of a handful of divs with "role=group", no?

site/src/pages/components/selectlist.mdx
```diff
@@ -165,18 +165,40 @@ This example has a `<listbox>` which enables it to put arbitrary content into th
...
+        <option>seven</option>
+        <option>eight</option>
+      </div>
+    </div>
   </listbox>
 </selectlist>
 <style>
-div {
-  font-weight: bold;
+.container {
```
mfreed7, line 193: Note: if you change to `<optgroup>` you can still do exactly this in CSS.

site/src/pages/components/selectlist.mdx
scottaohara: don't forget `alt=""` or `alt`

site/src/pages/components/selectlist.mdx
scottaohara: yes.  these should be `optgroup` and either use the `label` attribute or the proposed `legend` element to provide the groups the accName.

Otherwise, if using divs, we need to correct the markup.

so these should be either:

```
<optgroup>
  <legend>name for group</legend>
  <option>...</option>
   ....
</optgroup>
```

or

```
<optgroup label='name for group'>
  <option>...</option>
   ...
</optgroup>
```

or

```
<div role=group aria-labelledby=f>
  <div id=f>name for group</div>
  <option>...</option>
   ....
</optgroup>
```

scottaohara:
i may be worth noting specifically that just using emoji wouldn't be appropriate for many graphic + text label use cases, because emojis _are exposed_ as content.

so for instance,
```
<option>:gb: united kingdom</option>
```
would be result in this option having the accessible name of "united kingdom flag united kingdom" and that's silly.  There's no good way to mark that emoji as decorative with the current select > option parsing rules either.  cause even if someone did `<span>:gb:</span> ...` that span would be parsed out and we'd still get "united kingdom flag united kingdom".

lukewarlow:
Not to mention flag emojis don't actually render on all platforms.

That does beg the question should aria-hidden on Scott's mentioned span have an impact on how it's parsed? Like when you're producing a value to be submitted should aria-hidden stop it being included?

scottaohara:
> That does beg the question should aria-hidden on Scott's mentioned span have an impact on how it's parsed? Like when you're producing a value to be submitted should aria-hidden stop it being included? (Aria-hidden on the basis it's a good indication something is decorative)

ARIA shouldn't change the behavior of HTML.  It's only function is to modify how HTML is presented by the browser's accessibility tree.  If someone needs to have their option submit a particular value, they should specify the `value` attribute on that option, and let what's rendered as the visible/programmatic name of the option be just that.

site/src/pages/components/selectlist.mdx
josepharhar: I feel like using optgroup here would just be asking for people to say that `<select>` can already handle this use case.
`<optgroup>` has a user agent shadowroot and assigns the actual text with role=group from an attribute, which is not very robust, and `<selectlist>` doesn't have that limitation since youre allowed to put `<div role=group>`s in it.

But yeah this example does not actually do anything that optgroup can't do...

mfreed7: I definitely think an example using `<optgroup>` can showcase things the `<select>` can't do. For example, it can contain a fully-styled `<legend>` element (per [the resolution](https://github.com/openui/open-ui/issues/558#issuecomment-1198490800)):

```
<selectlist>
  <optgroup>
    <legend><img src=cool_group_icon>Fancy group</legend>
    <option>...
```

site/src/pages/components/selectlist.mdx
josepharhar: ok i changed them to optgroups

site/src/pages/components/selectlist.mdx
josepharhar: I switched back to country flags and added alt text.
The images don't exist in the repo, but they don't need to because they aren't actually downloaded or rendered anywhere on the website - the only exist in the demo which I made locally just to create the adjacent screenshot.

josepharhar:
> it may be worth noting specifically that just using emoji wouldn't be appropriate for many graphic + text label use cases, because emojis are exposed as content.

> Not to mention flag emojis don't actually render on all platforms.

Thanks! I'll go back to country flags then, they looked nicer than the animals anyway

mfreed7 (approved):

PR #923, opened 2023-10-31
Add X to the community links

### LINKED ISSUE #914: [Website] Open UI's X account should be linked
Apparently there's an X (Twitter) account for this CG.

It would make sense for this to be linked in the nav bar along with the other social medias

https://x.com/openuicg

YummyBacon5:
Closes #914

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -51,10 +51,10 @@
     "rehype-autolink-headings": "^6.1.1"
   },
   "dependencies": {
-    "@fortawesome/fontawesome-svg-core": "^1.2.36",
-    "@fortawesome/free-brands-svg-icons": "^5.15.4",
-    "@fortawesome/free-solid-svg-icons": "^5.15.4",
-    "@fortawesome/react-fontawesome": "^0.1.15",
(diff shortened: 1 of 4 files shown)


lukewarlow:
The yarn.lock changes seem quite big? The repository URL seems to have changed?

Can you also make sure the yarn lock doesn't revert the astro and lodash changes too.

Overall this looks good though.

YummyBacon5:
I think that I've now fixed the yarn.lock file. I think it was so different since I used npm instead of yarn to install the dependencies?

And lodash is still removed

gregwhitworth:
Goodness, am I the only one that can't stand the X logo with another `X` next to it? I'm tempted to request that we remove the words from the other icons and increase spacing. Thoughts @lukewarlow @una @andrico1234  ??

lukewarlow:
I agree the double X is odd. I'd be tempted to either drop the icons (instead of the words), or to just simply use the twitter icon and label it twitter instead.

YummyBacon5:
I don't mind the double X. But for now, I have removed the text and have increased the size of the icons

Issue #920, opened 2023-10-31
[dialog] Declarative indicator for dialog modality


lukewarlow:
The invokers proposal is a brilliant addition and really helps solve some of the developer ergonomic issues with dialogs (along with much more).

However, it misses one key case as I mentioned in https://github.com/whatwg/html/issues/3567, you can't have a modal dialog load open with just HTML, you need JS to call showModal() or a user to click an invoker button.

My use case is that I sometimes have modal dialogs that are linked to a client side router where the URL controls their visibility. I currently call .showModal() on component mount. But if I could instead just do this declaratively that would remove the requirement for client JS execution (think SSR scenarios)

`<dialog open>` works for non-modals but there's no equivalent for modal dialogs.

lukewarlow:
Happy to bikeshed etc but my proposal would be a new `modal` attribute. It wouldn't change anything by default it just acts as an indicator when used alongside the open attribute.

My thoughts on what `<dialog modal>` would do when called with JS.

`show()` - Throw an exception (I'm not fully set on this so could probably be persuaded either way).

`showModal()` - Work as expected.

This could also simplify the invoker proposal. Currently we have a [toggleModal](https://open-ui.org/components/invokers.explainer/#defaults) action, and Keith has added an additional `toggle` action inside of his Chromium CL.

If we added a new `modal` attribute, we could just have a single `toggle` action that would just work as expected?

Depending on the behaviour of `show()` JS function, we could even just have a `show` action and forget `showModal`?

YummyBacon5:
> If we added a new modal attribute

Could `<dialog open="modal">` not work?

lukewarlow:
> Could <dialog open="modal"> not work?

It would potentially for the open as modal on page load. Potentially as I'm not sure on changing boolean attribute to take values. I guess it was done with hidden so could be fine?

It wouldn't, however, additionally remove ambiguity around invoker actions though.

keithamus:
> Potentially as I'm not sure on changing boolean attribute to take values. I guess it was done with hidden so could be fine?

Yes this is fine as it will become an enumerated attribute where the invalid/empty states map to <whatever the current behaviour is>. The problem is that the current behaviour will need a name.

lukewarlow:
Naming things is hard, could that be another reason to use a new attribute for the modality instead. Then we get potential invoker benefits *and* we don't have to think of naming <whatever the current behaviour is>

mfreed7:
[...]

PR #936, opened 2023-11-06
add toggle action to invokers dialog

keithamus:
When prototyping/implementing the behaviour for invokers on `<dialog>`, I noticed that `toggle` wasn't an action available, only `toggleModal`.

For completeness and compatibility with non-modal dialogs, I went ahead and implemented `toggle`, which begs the question: should it exist?

Rather than raising an issue I thought I'd raise a PR for us to speculate on.

/cc @lukewarlow @mfreed7 @scottaohara

### DIFF
--- a/site/src/pages/components/invokers.explainer.mdx
+++ b/site/src/pages/components/invokers.explainer.mdx
@@ -373,6 +373,7 @@ that this list is ordered and higher rules take precedence:
 | `<* popover>`   | `'showPopover'`       | Call `.showPopover()` on the invokee                                                 |
 | `<dialog>`      | `'auto'`              | If the `<dialog>` is not `open`, call `showModal()`, otherwise cancel the dialog     |
 | `<dialog>`      | `'toggleModal'`       | If the `<dialog>` is not `open`, call `showModal()`, otherwise cancel the dialog     |
+| `<dialog>`      | `'toggle'`            | If the `<dialog>` is not `open`, call `show()`, otherwise cancel the dialog          |
 | `<dialog>`      | `'cancel'`            | If the `<dialog>` is `open`, cancel the dialog                                       |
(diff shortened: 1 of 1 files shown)


lukewarlow:
I think it should exist for completeness. Unless we make it so invokers ONLY work for modal dialogs? Which could be an option?

Also gonna link #920

site/src/pages/components/invokers.explainer.mdx
```diff
@@ -373,6 +373,7 @@ that this list is ordered and higher rules take precedence:
 | `<* popover>`   | `'showPopover'`       | Call `.showPopover()` on the invokee                                                 |
 | `<dialog>`      | `'auto'`              | If the `<dialog>` is not `open`, call `showModal()`, otherwise cancel the dialog     |
 | `<dialog>`      | `'toggleModal'`       | If the `<dialog>` is not `open`, call `showModal()`, otherwise cancel the dialog     |
+| `<dialog>`      | `'toggle'`            | If the `<dialog>` is not `open`, call `show()`, otherwise cancel the dialog          |
```
lukewarlow, line 376:
```suggestion
| `<dialog>`      | `'toggle'`            | If the `<dialog>` is not `open`, call `show()`, otherwise close the dialog          |
```

lukewarlow (changes requested):

site/src/pages/components/invokers.explainer.mdx
lukewarlow: See https://github.com/openui/open-ui/pull/951 for changing the others

gregwhitworth:
@keithamus this PR is almost a year old and I presume outdated? What do you want to do with it?

keithamus:
This can be closed as the current explainer includes this.

PR #943, opened 2023-11-10
Change `alt` attributes to be empty

YummyBacon5:
For these images, the `alt` attribute should be empty because these images are decorational. Since all the info is already conveyed through the option's text.

Currently, a screen reader would announce "flag of USA USA" as the accessible name, which is verbose.


gregwhitworth:
Agree with the complete redundancy here; but I'm curious if we should add an alt of "flag" to the image? This would result with an AT reading of "Image ... Flag ... USA, etc". I think I prefer your approach but would like @josepharhar to weigh in since he's primarily making the `selectlist` changes

scottaohara:
These should have alt values. Flag isn’t useful here unless the intent is to choose a flag

scottaohara (approved):

lukewarlow:
Not a blocker here but we should add label elements to each of the selectlists

coliff (approved):

YummyBacon5:
Hey, please can this PR be merged now? Since it has been months

lukewarlow:
@mfreed7 or @gregwhitworth would be good if one of you could approve so we can merge :)

mfreed7:
> These shouldnt have alt values. Flag isn’t useful here unless the intent is to choose a flag

Based on this reply, this LGTM.

mfreed7 (approved):

YummyBacon5:
It has been more months. This change does improve the accessible name. I guess I'm just not trusted enough to make these changes, lol.

Is there anything which I can change to get this merged?
@gregwhitworth, @dbaron, @josepharhar

mfreed7:
> This branch has conflicts that must be resolved
Use the [web editor](https://github.com/openui/open-ui/pull/943/conflicts) or the  to resolve conflicts.
Conflicting files
site/src/pages/components/selectlist.mdx

Once you fix that, it [already has the needed approvals](https://github.com/openui/open-ui/pull/943#issuecomment-2076114140) to land.

YummyBacon5:
Done

PR #955, opened 2023-11-16
Updating Combox explainer with key decisions

sudheer-gowrigari:
Updating Combox explainer with key decisions section


sudheer-gowrigari:
Will be added as part of updating combobox anatomy PR

PR #967, opened 2023-12-03
Add stepUp and stepDown actions for input elements

spookylukey:
See:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepUp
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepDown

The idea is that with this in place, we could easily create more visually appealing (and accessible) versions of `<input type=number>` without any javascript.

For example, the number widgets in this shopping basket from https://abilitynet.org.uk/training/how-use-screen-reader-accessibility-testing?utm_source=Intro-Screen-Readers&utm_medium=Factsheet-Highlights&utm_campaign=2023-Training-Promo

Screenshot 2023-12-03 at 14-15-42 How to use a screen reader for accessibility testing AbilityNet

It also requires some CSS to hide the native spinners - https://stackoverflow.com/questions/3790935/can-i-hide-the-html5-number-input-s-spin-box

### DIFF
--- a/site/src/pages/components/invokers.explainer.mdx
+++ b/site/src/pages/components/invokers.explainer.mdx
@@ -382,6 +382,8 @@ that this list is ordered and higher rules take precedence:
 | `<details>`     | `'close'`             | If the `<details>` is `open`, then close it                                          |
 | `<select>`      | `'showPicker'`        | Call `.showPicker()` on the invokee                                                  |
 | `<input>`       | `'showPicker'`        | Call `.showPicker()` on the invokee                                                  |
+| `<input>`       | `'stepUp'`            | Call `.stepUp()` on the invokee                                                      |
+| `<input>`       | `'stepDown'`          | Call `.stepDown()` on the invokee                                                    |
 | `<video>`       | `'playpause'`         | Toggle the `.playing` value                                                          |
 | `<video>`       | `'pause'`             | If `.playing` is `true`, set it to `false`                                           |
 | `<video>`       | `'play'`              | If `.playing` is `false`, set it to `true`                                           |


site/src/pages/components/invokers.explainer.mdx
```diff
@@ -382,6 +382,8 @@ that this list is ordered and higher rules take precedence:
 | `<details>`     | `'close'`             | If the `<details>` is `open`, then close it                                          |
 | `<select>`      | `'showPicker'`        | Call `.showPicker()` on the invokee                                                  |
 | `<input>`       | `'showPicker'`        | Call `.showPicker()` on the invokee                                                  |
+| `<input>`       | `'stepUp'`            | Call `.stepUp()` on the invokee                                                      |
+| `<input>`       | `'stepDown'`          | Call `.stepDown()` on the invokee                                                    |
```
lukewarlow, line 386: Can we change these to make clear it's just aimed at number inputs?

lukewarlow:
Have you got any thoughts on what the aria -mapping for these would look like.

Afaik the default number input stepper buttons aren't keyboard navigable and so probably don't have any inherent aria relationship.

site/src/pages/components/invokers.explainer.mdx
lukewarlow: Ah if type="number" won't fit without just removing spaces maybe leave for now. We can adjust it once other prs that refactor this make it in.

gregwhitworth:
@lukewarlow correct that most do not allow you to tab into shadow DOM and thus you can use the up or down keys to increment the value. However, there is still value here from a usecase perspective. These would be buttons although I probably wouldn't use accessible name that results in step. I would expect something along the lines of `increment` or `decrement` for the button title respectively.

Additionally, I akin to what we're doing with others; I would prefer to introduce a new element that contains these parts that have the wiring already in place as I'm unsure how we would retrofit them onto number. Technically speaking this is already possible today but the invoker to increment/decrement doesn't exist.

Another usecase that is akin to this are carousels where they often have related buttons that invoke a `next` and `previous` "card". I'm not implying they should have the same invoker but the a11y paradigms will be of similar nature.

gregwhitworth:
@spookylukey can you please complete the IPR contribution requirements

ryantownsend:
Came to this repo to make this suggestion – so just here adding a +1 to having a mechanism for incrementing/decrementing number inputs. I'm manually patching this behaviour in myself at the moment.

lukewarlow:
I'm gonna go ahead and close this and replace it by https://github.com/openui/open-ui/pull/1029 it's effectively the same and I've included you as the commit co-author @spookylukey just deals with the IPR problems as I'd like to get these included.

One difference is for now I've specified this just for number input, which is I think where the most utility comes from. If there's other input types where this would be useful raise an issue and we can definitely consider them.

PR #997, opened 2024-02-22
Update code owners for invokers to include Keith and I (Luke)

lukewarlow:
We can update codeowners more as time goes on but as a first attempt I think it makes sense for me and Keith to "own" invokers (including interest) in addition to the 3 main codeowners?


lukewarlow:
Ah I thnk @keithamus needs write access to be a codeowner (which makes sense) @gregwhitworth would that be okay with you?

mfreed7 (approved):
LGTM, modulo the question about @keithamus having write access. Sounds fine to me, but @gregwhitworth any issues?

gregwhitworth:
@mfreed7 nope - let me update them now

Issue #839, opened 2023-09-20
[invokers] Can we add interesttarget support to more elements


lukewarlow:
See https://github.com/keithamus/invoker-buttons-proposal/issues/14 for more details and that repo in general for full details.

TLDR: `interesttarget` attribute currently proposed to be supported on buttons and links. Could we add support to `<area>`, `<input>`, `<textarea>`, `<select>`, `<summary>`. (`selectlist` in future would be on this list) along with potentially other elements?

Are there extra elements we can/should add this too?

lukewarlow:
To clarify is there a specific reason against `<area>`?

lukewarlow:
See https://github.com/openui/open-ui/issues/857 for a request for interest support on abbr elements.

lukewarlow:
Just while I remember I thought I'd not it down one use case for interesttarget (at least the focus aspect) on inputs in general is the `<input list="foo">` type scenario where you want to be able to trigger a popover of suggestions while focused on an input. Though potentially this is better off being a platform editable combobox type setup and we don't expose the capability beyond that at least initially.

Much the same way that selectlist will have focusgroup behaviour (afaik) but we don't expose focusgroup yet.

adactio:
Could `interestTarget` be used to provide a declarative option for some JavaScript APIs? I'm thinking specifcally of JavaScript APIs that require a user interaction (and so most demos involve the creation of a button element anyway) and that don't currently have an equivalent HTML element.

For example:

* The Fullscreen API
* The Geolocation API
* The Web Share API

For that last one, the default behaviour could be sharing the current URL. I've been [collecting use cases](https://github.com/adactio/share-button-type/issues/1) to inform [a proposal for `button type="share"`](https://github.com/adactio/share-button-type/issues/1) but `button interestarget="share"` would achieve the same goal.

Likewise, `button interesttartget="requestfullscreen"` and `button interesttarget="exitfullscreen` could be directly equivalent to `document.requestfullscreen` and `document.exitfullscreen` but without the need to create a button with JavaScript (or write any JavaScript at all).

I'm not sure how the `button interesttarget="geolocation"` would work other than perhaps populating an `input` element using `invoketarget`.

I realise that this issue was asking specifically for HTML elements, not JavaScript APIs, so I apologize if this is off-topic.

lukewarlow:
@adactio could you open a new issue for that discussion?

[...]

PR #1000, opened 2024-02-25
Update interest invokers introduction

lukewarlow:
To show that submit buttons and area elements are supported

### DIFF
--- a/site/src/pages/components/interest-invokers.explainer.mdx
+++ b/site/src/pages/components/interest-invokers.explainer.mdx
@@ -23,11 +23,11 @@ layout: ../../layouts/ComponentLayout.astro
 
 Following [invokers](../invokers.explainer), adding an `interesttarget`
 attribute to interactive elements: starting with `<button>`, `<input
-type="button">`/`<input type="reset">`, and `<a>` (perhaps expanding to
-`<area>`, `<input>`, `<textarea>`, `<select>`, `<summary>` or maybe more, see
+type="button">`/`<input type="reset">`/`<input type="submit">`, and `<a>` (perhaps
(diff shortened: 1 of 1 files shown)


site/src/pages/components/interest-invokers.explainer.mdx
```diff
@@ -23,11 +23,12 @@ layout: ../../layouts/ComponentLayout.astro
 
 Following [invokers](../invokers.explainer), adding an `interesttarget`
 attribute to interactive elements: starting with `<button>`, `<input
-type="button">`/`<input type="reset">`, and `<a>` (perhaps expanding to
-`<area>`, `<input>`, `<textarea>`, `<select>`, `<summary>` or maybe more, see
+type="button">`/`<input type="reset">`/`<input type="submit">`, `<a>`
+ and `<area>` (perhaps expanding to
```
lukewarlow, line 27: Area is included further down the page so this just fixes that here. Submit buttons are similarly mentioned as supported so make that explicit here.

keithamus:
I think in https://github.com/openui/open-ui/issues/839#issuecomment-1739848561 we only resolved to support Input/Button/Anchor right?

> * `RESOLVED: support interesttarget on buttons (<input type=button> or <button>) and links (<a>) only, and explore expanding to other element types in the future.`

lukewarlow:
Ah correct hmm then I guess this should be updated the other way. 

I'm not keen on it being on anchor but not area personally though. If it's not an element we should be using we should deprecate it.

lukewarlow:
Fixed that and made #1003

keithamus (approved):

lukewarlow:
I apparently messed up the code owners change? Could you see what I did wrong?

mfreed7 (approved):

PR #1004, opened 2024-02-26
Fix dialog toggle action in invokers

lukewarlow:
Also fix missing usages of button value for returnValue of dialog.

### DIFF
--- a/site/src/pages/components/invokers.explainer.mdx
+++ b/site/src/pages/components/invokers.explainer.mdx
@@ -281,7 +281,7 @@ handled. Note that this list is ordered and higher rules take precedence:
 | `<* popover>`   | `'togglePopover'`     | Call `.togglePopover()` on the invokee                                               |
 | `<* popover>`   | `'hidePopover'`       | Call `.hidePopover()` on the invokee                                                 |
 | `<* popover>`   | `'showPopover'`       | Call `.showPopover()` on the invokee                                                 |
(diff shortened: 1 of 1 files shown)


site/src/pages/components/invokers.explainer.mdx
```diff
@@ -294,7 +294,7 @@ Invokers, the names and exact semantics may be subject to change:
 | `<details>`     | `'toggle'`            | If the `<details>` is `open`, then close it, otherwise open it                       |
 | `<details>`     | `'open'`              | If the `<details>` is not `open`, then open it                                       |
 | `<details>`     | `'close'`             | If the `<details>` is `open`, then close it                                          |
-| `<dialog>`      | `'toggle'`            | If the `<dialog>` is `open`, cancel the dialog                                       |
+| `<dialog>`      | `'toggle'`            | If the `<dialog>` is `open`, then close it, otherwise call `showModal()`                                      |
```
lukewarlow, line 297: I can't remember what we decided for cancel vs close I believe it should be close and showModal right?

site/src/pages/components/invokers.explainer.mdx
```diff
@@ -294,7 +294,7 @@ Invokers, the names and exact semantics may be subject to change:
 | `<details>`     | `'toggle'`            | If the `<details>` is `open`, then close it, otherwise open it                       |
```
lukewarlow, line 294: Did we decided if we were keeping auto values?

For example the dialog closedby attribute will have a missing value default of "auto" without it needing an associated keyword?

keithamus (approved):

PR #1006, opened 2024-02-26
Change invokeaction `'auto'` keyword to `auto` state

lukewarlow:
- Remove explicit `'auto'` keyword, include mention of `auto` state when value is missing.

It's been discussed before that having `'auto'` and corresponding action keywords (e.g. `'toggle'`) was odd. The idea of auto is useful you don't want to add the invokeaction attribute for something that's an obvious default, like toggle is for a popover today.

After some recent discussions around the dialog `closedby` attribute proposal it was pointed out you can map the missing value to a state without an associated keyword.

This is the OpenUI side to doing that for invokeaction.


keithamus (approved):

site/src/pages/components/invokers.explainer.mdx
```diff
@@ -275,22 +275,24 @@ additional behaviours alongside the event dispatch. In addition the following
 table represents how built-in invocations on specific element types are
 handled. Note that this list is ordered and higher rules take precedence:
 
+When the `invokeaction` attribute is missing it will default to an `auto` state.
+
 | Invokee Element | `action` hint         | Behaviour                                                                            |
 | :-------------- | :-------------------- | :----------------------------------------------------------------------------------- |
-| `<* popover>`   | `'auto'`              | Call `.togglePopover()` on the invokee                                               |
+| `<* popover>`   | `auto`                | Same as `'togglePopover'`                                                            |
```
keithamus, line 282: Non blocking but this could be `(auto)` instead of `` `auto` `` for extra clarity, perhaps.
```suggestion
| `<* popover>`   | (auto)                | Same as `'togglePopover'`                                                            |
```

lukewarlow: Ah good shout

keithamus (approved):

PR #1010, opened 2024-02-27
[focusgroup] Add mention and example for Shadow DOM behavior

### LINKED ISSUE #521: [focusgroup] Should it support shadow trees?
Copied from https://github.com/MicrosoftEdge/MSEdgeExplainers/issues/556

@benbeaudry filed originally on 2022-02-23

> I'm implementing a prototype of this explainer in Chromium and I'm wondering whether we should allow elements in shadow trees to be part of a focusgroup or not. I couldn't find any mention in the explainer, but I don't see why we shouldn't also include elements hidden within a shadow tree.
>
>WDYT? If you agree, it might be a good idea to mention it in the explainer.

@travisleithead replied

> Good point. Insofar as a shadow dom is used to wrap the internals of a logical component, it might makes sense for the host element of that component to declare a focusgroup such that descendents of the focusgroup within the component can take advantage of it. Definitely needs to be described in the explainer/spec.

> Perhaps we should piggyback on the existing `delegatesFocus` flag. If present, then the author of the Shadow Root intends for focus to transition to some sub-component. Therefore, if a focusgroup definition is placed on the host element, then the focusable children (inside the component) should attempt to "link-up" to a cross-shadow-root focusgroup definition in their ancestor "flat tree". Otherwise (the `delegatesFocus` flag is absent), only elements in that light tree (e.g., those exposed through slots) would be candidates. This may require some clever defintion of the work "children" (e.g., shadow-inclusive children?).

@benbeaudry replied

> The "delegatesFocus" flag that already exist is definitely something we _could_ leverage with the focusgroup attribute. However, after some digging, it seems like this flag is only used to redirect the focus event from the ShadowRoot to the first focusable element within the ShadowDOM when the ShadowRoot isn't focusable.
>
> In our situation, the use case would be a bit different: the focus would move from an element outside of the ShadowDOM to an element inside of it on an arrow key press without trying to first go to the ShadowRoot. In my opinion, the delegatesFocus flag is a totally different feature and it doesn't make much sense to piggyback on it because of the different use cases.
>
> I am by no means an expert on the delegatesFocus flag or on ShadowDOMs, so my opinion is to be taken with a grain of salt.
>
> That being said, for the moment, I've decided to allow a focusgroup within a ShadowDOM to extend one that is declared outside of that ShadowDOM in the initial implementation of the focusgroup feature (in Chromium-based browsers). I'll keep monitoring this thread to bring the appropriate changes to the Chromium implementation if needed.

travisleithead:
This change:
* Makes mention in various places that children/descendants are now "shadow-inclusive".
* Adds an example of a scenario where Shadow DOM children will participate in a focusgroup
* Adds a "Design decisions" section to document this decision and links it to the primary issue.

Fixes #521

### DIFF
--- a/site/src/pages/components/focusgroup.explainer.mdx
+++ b/site/src/pages/components/focusgroup.explainer.mdx
@@ -105,6 +105,8 @@ focus tracking are unchanged with this proposal.
 
 1. (Child opt-in) Group a set of focusable child elements under a single focusgroup.
 2. (Descendent opt-in) Focusable elements deeply nested can participate in a single focusgroup.
+3. (Cross Shadow DOM) Focusable elements nested inside a Shadow DOM are discoverable and focusable
+   when their Shadow Host or an ancestor element declares a focusgroup (by default).
 3. (Wrap) Focusgroup can be configured to have wrap-around focus semantics.
 4. (Horizontal/vertical) A focusgroup can be configured to respond to either horizontal navigation
    keys or vertical keys or both (to trivially reserve one axis of arrow key behavior for
@@ -139,7 +141,7 @@ Focusgroups consist of a **focusgroup definition** that establish **focusgroup c
 items. Focusgroup items are the elements that actually participate in the focusgroup (among the possible
 focusgroup candidates).
 
-When a linear focusgroup definition is associated with an element, all of that element's direct children
+When a linear focusgroup definition is associated with an element, all of that element's shadow-inclusive direct children
 become focusgroup candidates. Focusgroup candidates become focusgroup items if they are focusable, e.g.,
 implicitly focusable elements like `<button>`, or explicitly made focusable via `tabindex` or some
 other mechanism (e.g., `contenteditable`).
@@ -169,7 +171,7 @@ a reserved name that corresponds to the same focusgroup implied by the HTML `foc
 ```
 
 For the `parent` element which includes the focusgroup definition, the elements `one`, `two`, and `three`
-(and any other children of `parent` that may be added or removed) are focusgroup candidates and because
+(and any other shadow-inclusive children of `parent` that may be added or removed) are focusgroup candidates and because
 each are focusable, they also become focusgroup items. When one of the focusgroup items is focused, then
 the user can move focus sequentially according to DOM order among all the focusgroup items using the arrow
 keys (up/right moves focus forward, down/left moves focus backwards). Note that only elements `one` and
@@ -221,7 +223,7 @@ The CSS-specified value of 'nowrap' would override the HTML attribute value's "w
 similarly for 'both' overriding "horizontal".
 
 There is no change to the way Tab navigation works with `tabindex` nor the Tab ordering behavior. To
-use a focusgroup, focus must enter that element's focusable children somehow: for accessibility
+use a focusgroup, focus must enter that element's focusable shadow-inclusive children somehow: for accessibility
 and keyboard-only scenarios, the Tab key is typically used--programmatic calls to `element.focus()`
 or user clicks with a pointing device are alternatives.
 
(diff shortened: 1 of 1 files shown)


site/src/pages/components/focusgroup.explainer.mdx
```diff
@@ -105,6 +105,8 @@ focus tracking are unchanged with this proposal.
 
 1. (Child opt-in) Group a set of focusable child elements under a single focusgroup.
 2. (Descendent opt-in) Focusable elements deeply nested can participate in a single focusgroup.
+3. (Cross Shadow DOM) Focusable elements nested inside a Shadow DOM are discoverable and focusable
```
gregwhitworth, line 108: is "by default" needed in this statement?

site/src/pages/components/focusgroup.explainer.mdx
```diff
@@ -221,7 +223,7 @@ The CSS-specified value of 'nowrap' would override the HTML attribute value's "w
 similarly for 'both' overriding "horizontal".
 
 There is no change to the way Tab navigation works with `tabindex` nor the Tab ordering behavior. To
-use a focusgroup, focus must enter that element's focusable children somehow: for accessibility
+use a focusgroup, focus must enter that element's focusable shadow-inclusive children somehow: for accessibility
```
gregwhitworth, line 226: The statement of "somehow" feels odd to me given how normative this text reads. Unsure where WHATWG defines this but I recommend adjusting this to possibly something along the lines of "it enters the elements focusable shadow-inclusive children by traversing to the first child in based on standard tab ordering behavior (link out to WHATWG definition/algo)"

site/src/pages/components/focusgroup.explainer.mdx
```diff
@@ -270,6 +272,22 @@ from "selection/toggle" state.
 A future extension of this proposal may allow focus entering a focusgroup to jump to the currently
 toggled (or the toggle with the highest value) if any toggles are set in the focusgroup.
 
+Focusgroup definitions apply by default across Shadow DOM boundaries in order to make it easy for component
```
gregwhitworth, line 275: again I would remove "by default" as this is showing how people can leverage standards that exist to make it so that their shadow children are not focusable.

site/src/pages/components/focusgroup.explainer.mdx
```diff
@@ -310,9 +328,9 @@ this purpose.
 
 ### 6.4. Expanding and connecting linear focusgroups together (`extend`)
 
-By default, a focusgroup definition's focusgroup candidates are its direct children. There are two
+By default, a focusgroup definition's focusgroup candidates are its direct shadow-inclusive children. There are two
```
gregwhitworth, line 331: Remove "by default"

site/src/pages/components/focusgroup.explainer.mdx
```diff
@@ -1223,7 +1248,7 @@ navigation modes, while a user not working with an AT (or ATs designed for visua
 would enable the full spatial navigation model. An opt-in for spatial navigation would certainly
 be a requirement and could be an extension to `focusgroup` (e.g., `focusgroup=spatial`).
 
-## 10. Open Questions
+## 11. Open Questions
```
gregwhitworth, line 1251: Can you have these link out to respective issues if they don't already exist?

gregwhitworth (approved):
Primarily nit changes, thanks for submitting.

site/src/pages/components/focusgroup.explainer.mdx
travisleithead: That was pre-existing bad prose... but since I'm here I'm going to clean it up for the time-being. Note, this paragraph will shortly be obsolete because the feature IS changing to modify how Tab navigation works in certain scenarios. :)

site/src/pages/components/focusgroup.explainer.mdx
travisleithead: Done.

site/src/pages/components/focusgroup.explainer.mdx
travisleithead: Done.

site/src/pages/components/focusgroup.explainer.mdx
travisleithead: Yep!

PR #1029, opened 2024-04-05
Add stepUp and stepDown actions for number inputs

### LINKED ISSUE #968: [invokers] stepUp and stepDown actions for number inputs
See https://github.com/openui/open-ui/pull/967 for a PR by @spookylukey

>See:

> - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number
> - https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepUp
> - https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/stepDown

> The idea is that with this in place, we could easily create more visually appealing (and accessible) versions of `<input type=number>` without any javascript.

> For example, the number widgets in this shopping basket from https://abilitynet.org.uk/training/how-use-screen-reader-accessibility-testing?utm_source=Intro-Screen-Readers&utm_medium=Factsheet-Highlights&utm_campaign=2023-Training-Promo

> Screenshot 2023-12-03 at 14-15-42 How to use a screen reader for accessibility testing AbilityNet

> It also requires some CSS to hide the native spinners - https://stackoverflow.com/questions/3790935/can-i-hide-the-html5-number-input-s-spin-box

Opening the issue to have a place to discuss and agenda+ if necessary.

lukewarlow:
Fixes #968

Closes #967 

Definitely needs some thought with regards to aria mappings that perhaps @scottaohara could help with. But think it's a good idea to get these listed as future so we do address them.

### DIFF
--- a/site/src/pages/components/invokers.explainer.mdx
+++ b/site/src/pages/components/invokers.explainer.mdx
@@ -311,6 +311,8 @@ Invokers, the names and exact semantics may be subject to change:
 | `<*>`           | `'toggleFullscreen'`  | If the element is fullscreen, then exit, otherwise request to enter                  |
 | `<*>`           | `'requestFullscreen'` | Request the element to enter into 'fullscreen' mode                                  |
 | `<*>`           | `'exitFullscreen'`    | Request the element to exit 'fullscreen' mode                                        |
+| `<input type=number>` | `'stepUp'`      | Call `.stepUp()` on the invokee                                                      |
+| `<input type=number>` | `'stepDown'`    | Call `.stepDown()` on the invokee                                                    |
 
 > Note: The above tables are an attempt at wide coverage, but ideas are welcome. Please submit a PR if you have one!
 


site/src/pages/components/invokers.explainer.mdx
```diff
@@ -311,6 +311,8 @@ Invokers, the names and exact semantics may be subject to change:
 | `<*>`           | `'toggleFullscreen'`  | If the element is fullscreen, then exit, otherwise request to enter                  |
 | `<*>`           | `'requestFullscreen'` | Request the element to enter into 'fullscreen' mode                                  |
 | `<*>`           | `'exitFullscreen'`    | Request the element to exit 'fullscreen' mode                                        |
+| `<input type=number>` | `'stepUp'`      | Call `.stepUp()` on the invokee                                                      |
```
lukewarlow, line 314: stepUp and stepDown take an argument which is multiplied by the step attribute. Should invokers handle that somehow? Thinking maybe the button value attribute can be used as the argument?

We have precedence for the dialog returnValue?

mfreed7: I think it'd be best to avoid getting more complicated on things like this. The number already has a step, so the most natural argument is `1` - step by 1.

mfreed7 (approved):

Issue #863, opened 2023-10-05
selectlist: Should the "checked" option have a checkmark next to it?


josepharhar:
I touched on this while discussing https://github.com/openui/open-ui/issues/827

In order to indicate the currently selected option inside the listbox (I supposed it should also already be indicated in the button), should the currently selected/checked option in a single select selectlist have a checkbox rendered next to it?

For multi select, we would of course need this. For single select, maybe not?

The MacOS native select picker has a checkmark to the left of the selected option and also renders on top of the button which would otherwise indicate what the selected option is. The aria practices single select example has a checkmark to the right of the option: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

If we did this, how would it be implemented? Just a user-agent style thats like `option:checked::after { content: '✓' }`? Or perhaps a new pseudo-element like `::checkmark`? Should it be an SVG? Or be some sort of native checkbox thing? Should it be to the right of the option like the aria-practices example? Or should it be on the left? If it's on the left, should there be padding there for every option?

Whatever the case is, I feel like this should be easy for the developer to override/replace/remove, right @una? Would the `content` property in `option:checked::after { content: '✓' }` be easy enough to override? Should there be an easy way to move it from the left side of the option to the right side or vice versa?

@scottaohara @smhigley

lukewarlow:
My gut feeling on this is don't bother providing a check by default at all. Instead just provide UA styles of the following (or something similar). This way the default is usable but it's ultimately up to the authors to make it nicer. Because I feel like adding a checkmark is just going to create more work for ourselves making a good way to style it (which will mostly just be removing it)?

```css
option:checked {
  color: selecteditemtext;
  background-color: selecteditem;
}
```

lukewarlow:
Happy to make a new issue for this but seeing as it's slightly related.

selectlist {
  accent-color: blue;
}

Does not seem to impact on any of the default UA styles. 

Should it?

Or should the style be using AccentColor and AccentColorText (I think probably not raw but potentially modified?)

josepharhar:
> adding a checkmark is just going to create more work for ourselves making a good way to style it

Agreed, in the OP I almost said "I have so many questions that I don't feel like this can be completed before we finish single select and move on to multi select".

> Does not seem to impact on any of the default UA styles (e.g hover state)

Are these things already stylable via other CSS properties?
[...]

PR #1059, opened 2024-05-29
Add checkmark to select explainer

josepharhar:
Fixes https://github.com/openui/open-ui/issues/863


lukewarlow:
I'm interested to here the reasoning for why ::before is easy but ::marker is hard?

mfreed7 (approved):

josepharhar:
As it's currently implemented in chromium, setting display isn't allowed on ::marker I think. There also might be restrictions on which elements you can put ::marker on and how its rendered or what it renders. ::before on the other hand you can do whatever you want.

I think it would be easier to create a new pseudo element than reuse ::marker, but I think that's another discussion we can have later.

PR #1078, opened 2024-07-30
rewrite accessibility section on invokers

keithamus:
This rewrites the a11y section on invokers to be far more comprehensive, detailing the exact intentions we have when it comes to applying various ARIA states to the button, as well as focus fixups.

### DIFF
--- a/site/src/pages/components/invokers.explainer.mdx
+++ b/site/src/pages/components/invokers.explainer.mdx
@@ -318,7 +318,7 @@ equivalent will be exposed to Assistive Technologies.
 
 Buttons with these popover commands will implicitly receive
 `aria-details=IDREF`, where `IDREF` matches that of the `commandfor` attribute,
-while the popover is in the showing state.
+under certain conditions while the popover is in the showing state.
 
 <details>
 
@@ -326,15 +326,15 @@ while the popover is in the showing state.
 
 Some Assistive Technologes allow users to press a keyboard shortcut to navigate
 to the element pointed to by `aria-details`. This is useful in scenarios where
-the next focusable control is _not_ the popover, such as if there are buttons
+the popover is _not_ the invoking element's next accessibility sibling, such as if there are buttons
 or other controls between the invoking button and the popover.
 
 </details>
 
 <br/><br/>
 
-Browsers should omit `aria-details` based on heuristics where it is redundant,
-such as the contents of the popover being the next focusable control, or while
+Browsers are expected to omit `aria-details` based on heuristics where it is redundant,
+such as the contents of the popover being the next accessibility sibling to the invoking element, or while
 the popover's contents are not traversable, such as when it is hidden, or not
 in the DOM.
 
@@ -352,8 +352,8 @@ state, such that the button always reflects the state of expansion.
 
 Some Assistive Technologies will announce the state of expansion while the
 button is focussed, for example a `<button aria-expanded=false>File</button>`
-may be announced in a screen reader as "File, button, collapsed", while
-`<button aria-expanded=true>File</button>` may be announced in a screen reader
+may be announced by a screen reader as "File, button, collapsed", while
+`<button aria-expanded=true>File</button>` may be announced by a screen reader
 as `File, button, expanded`. This is helpful information for users who require
 an alternative of the visual confirmation of the popover being rendered on
 screen.
--- a/site/src/pages/components/invokers.explainer.mdx
+++ b/site/src/pages/components/invokers.explainer.mdx
@@ -454,8 +454,8 @@ possible to navigate to the invoking button.
 ##### aria-controls
 
 Similar to `aria-details`, `aria-controls` is not useful for a dialog that
-is closed, and non-navigable for a dialog that is open as modal, so it would be
-useless to add.
+is closed, and non-navigable for a dialog that is open as modal, so it would 
+provide no benefit to add.
 
 ##### aria-pressed
 
@@ -472,13 +472,13 @@ to provide it an implicit `aria-expanded=true` state also.
 
 ##### aria-haspopup=dialog
 
-When the `commandfor=` element is a `<dialog>`, adding aria-haspopup=dialog is
-a potentially viable attribute to add. Some Assistive Technologies will announce
+When the `commandfor=` element is a `<dialog>` it might seem `aria-haspopup=dialog` 
+would be a potentially viable attribute to add. Some Assistive Technologies will announce
 that the button will open a dialog, for example
 `<button aria-haspopup=dialog>Delete</button>` may be announced by a screen
 reader as "Delete, button, opens dialog". However, buttons that open dialogs
-rarely come with additional visual treatment, and so the lack of a visual
-analog means that using this attribute likely provides surplus information
+rarely come with additional visual treatment, and so the [lack of a visual
+analog means that using this attribute likely provides surplus information](https://w3c.github.io/aria/#h-note-50)
 to users of assistive technologies that isn't otherwise provided. Additionally,
 the _intent_ for `aria-haspopup` is more focused towards combobox popovers,
 where the type of popover may be ambigious. As command buttons can be used for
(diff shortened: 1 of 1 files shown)


mfreed7 (approved):
Wow, awesome new section! This looks good to me, but I'm not the right person to judge it. I'll stamp this approved, because I'm ok landing it and putting up PRs to improve it over time. But @scottaohara and @aleventhal please take a look and also @ mention anyone else that should have a look.

scottaohara:
i'll review this later today or tomorrow.  thanks @keithamus

site/src/pages/components/invokers.explainer.mdx
```diff
@@ -326,6 +302,302 @@ Invokers, the names and exact semantics may be subject to change:
...
+<summary>Other considerations not explicitly proposed.</summary>
+
+##### aria-pressed
+
+Given elements will have `aria-expanded`, adding `aria-pressed` would be
+confusing or redundant, and as such won't be proposed for these buttons.
+
+##### aria-controls
+
+While `aria-controls` attempts to establish a similar style of relationship to
```
scottaohara, line 429: this isn't really accurate - re: parent/child relationship.  i'm not sure why this is mentioned at all if it's not being used / was never implemented for popovers?

site/src/pages/components/invokers.explainer.mdx
```diff
@@ -326,6 +302,302 @@ Invokers, the names and exact semantics may be subject to change:
...
+`aria-details`, it is largely intented for parent-child relationships. Buttons
+that open popovers do not necessarily represent a parent-child relationship.
+
+</details>
+
+<br/><br/>
+
+#### Buttons with `command=show-modal`
+
+Buttons with this dialog commands will implicitly receive `aria-haspopup=dialog`.
```
scottaohara, line 439: same comments as before.  the intent for the haspopup was more to help users know what types of popups were invoked from a combobox.  i really am trying to get ARIA back on track for that being the primary purpose of this attribute, i really dont want to see this get used for buttons and further cement a practice this wasn't originally intended for.

site/src/pages/components/invokers.explainer.mdx
```diff
@@ -326,6 +302,302 @@ Invokers, the names and exact semantics may be subject to change:
...
+
+Exposing an `aria-details` relationship between the button and the dialog is
+redundant. These relationships are useful only when the related element is
+open and present on the page and the invoking button is still navigable. A
+modal dialog renders the rest of the page inert meaning that it won't be
+possible to navigate to the invoking button.
+
+##### aria-pressed
+
+The `aria-pressed` state is the incorrect type of association for elements
```
scottaohara, line 495: it's weird this says aria-expanded would be the correct state to expose, but in the next section also talks about how the state is redundant.

site/src/pages/components/invokers.explainer.mdx
```diff
@@ -326,6 +302,302 @@ Invokers, the names and exact semantics may be subject to change:
...
+
+Similar to `aria-details`, the `aria-expanded` state redundant, given the
+dialog is opened as modal. For the user to have navigated to a button that
+opens a modal dialog, the dialog must not be open. If the dialog is open then
+the button will be inert, therefore non-navigable, therefore it is redundant
+to provide it an implicit `aria-expanded=true` state also.
+
+##### aria-controls
+
+While `aria-controls` attempts to establish a similar style of relationship to
```
scottaohara, line 509: similar as before, the description here is not accurate, but also just doesn't seem worth mentioning if the only reason to do so is to say "it's not used"

scottaohara (commented):
let me know if there are questions with my comments

mfreed7:
> let me know if there are questions with my comments


@keithamus is the plan still to update this with @scottaohara 's comments before landing? (I hope so.)

scottaohara:
we are going to need the HTML AAM PR for this as well, so whether fixed here or in that PR (and then this gets updated to just say 'go look at html aam' instead) is fine by me.  

but from what i understand, we are going to need that HTML AAM PR to land at the same / close to the same time as the HTML PR.  so we need to get that going / address the above either way.

site/src/pages/components/invokers.explainer.mdx
keithamus: I tried to add these extra bits as somewhat of an FAQ to field questions of "why hasn't this particular property been implemented", to have some written down rationale that people can refer to when those questions arise.

scottaohara: ok, that makes sense.  

so then it's more of a way this is described then - since it's not commonly a parent/child relationship - but more of an association between related elements

keithamus: Would it be better to point out instead that `aria-details` has better support, and applying both is redundant?

scottaohara: yah that could work.  so something like

aria-controls is not used due to lack of support among screen readers, and even if it was supported it would be redundant to the association already created due to the implicit aria-details relationship already created.

keithamus:
@scottaohara I believe I've addressed your review commentary. If you would kindly give it another pass that would be much appreciated.

site/src/pages/components/invokers.explainer.mdx
```diff
@@ -490,11 +451,16 @@ open and present on the page and the invoking button is still navigable. A
 modal dialog renders the rest of the page inert meaning that it won't be
 possible to navigate to the invoking button.
 
+##### aria-controls
+
+Similar to `aria-details`, `aria-controls` is not useful for a dialog that
+is closed, and non-navigable for a dialog that is open as modal, so it would be
+useless to add.
```
scottaohara, line 458: wording nit to not be redundant in saying it 'is not useful' and then 'it would be useless...."

```suggestion
is closed, and non-navigable for a dialog that is open as modal, so it would 
provide no benefit to add.
```

site/src/pages/components/invokers.explainer.mdx
```diff
@@ -504,11 +470,19 @@ opens a modal dialog, the dialog must not be open. If the dialog is open then
...
 to provide it an implicit `aria-expanded=true` state also.
 
-##### aria-controls
+##### aria-haspopup=dialog
 
-While `aria-controls` attempts to establish a similar style of relationship to
-`aria-details`, it is largely intented for parent-child relationships. Buttons
-that open modal dialogs would not represent a parent-child relationship.
+When the `commandfor=` element is a `<dialog>`, adding aria-haspopup=dialog is
+a potentially viable attribute to add. Some Assistive Technologies will announce
```
scottaohara, line 476:
```suggestion
When the `commandfor=` element is a `<dialog>`, `aria-haspopup=dialog` might seem
a potentially viable attribute to add. Some Assistive Technologies will announce
```

site/src/pages/components/invokers.explainer.mdx
```diff
@@ -504,11 +470,19 @@ opens a modal dialog, the dialog must not be open. If the dialog is open then
...
+that the button will open a dialog, for example
+`<button aria-haspopup=dialog>Delete</button>` may be announced by a screen
+reader as "Delete, button, opens dialog". However, buttons that open dialogs
+rarely come with additional visual treatment, and so the lack of a visual
+analog means that using this attribute likely provides surplus information
```
scottaohara, line 481:
```suggestion
rarely come with additional visual treatment, and so the [lack of a visual
analog means that using this attribute likely provides surplus information](https://w3c.github.io/aria/#h-note-50)
```

scottaohara (approved):
approving with minor additional suggestions.  take them as you will.  thank you!

keithamus:
Thanks @scottaohara! I'll merge and get to work on the AAM PRs!

scottaohara:
> Thanks @scottaohara! I'll merge and get to work on the AAM PRs!

excellent. thank you.  just give me a holler if you have any questions when making that PR.

PR #1089, opened 2024-09-09
Update customizable select syntax

josepharhar:
This PR updates the syntax based on recent standards discussions.

- Rename from ::select-fallback-datalist to ::picker(select)
- ::picker(select){appearance:base-select} requirement
- Removal of ::select-fallback-button

I also noticed that the split button image had a 404 because it was in the wrong directory so I moved it to the correct directory where all the other images are.


site/src/pages/components/customizableselect.mdx
```diff
@@ -245,6 +263,9 @@ This example has a complex structure wrapping the `<option>`s which is all slott
 .container > div {
   background-color: lightgray;
 }
+select, select::picker(select) {
```
mfreed7, line 266: I think **all** of these should be simplified:

```
select, ::picker(select) {
```

(the `::picker(select)` can only ever match on a `<select>` element.)

josepharhar: I simplified all the code examples, but I kept it as select::picker(select) in some of the explanatory text because I think it's helpful to know that it applies to the select element

PR #1101, opened 2024-09-27
Add select explainer redirect

josepharhar:
css tricks has an article which is pointing to a 404 on our site here: https://css-tricks.com/the-selectmenu-element/#:~:text=Based%20on%20the%C2%A0Open%20UI%E2%80%99s%C2%A0%3Cselect%3E%C2%A0proposal

This patch adds a page with a link to the updated explainer to make sure we don't continue to 404.


site/src/pages/components/select.mdx
```diff
@@ -0,0 +1,6 @@
+---
+name: Select Element (Explainer)
+layout: ../../layouts/ComponentLayout.astro
+---
+
+This explainer has been renamed to customzable select. [See the new explainer here](../customizableselect/).
```
lukewarlow, line 6:
```suggestion
This explainer has been renamed to customizable select. [See the new explainer here](../customizableselect/).
```

mfreed7:
This is "ok" like this, but perhaps better is a server redirect?

See this URL for an example:

https://open-ui.org/components/popup.research.explainer/

It automatically redirects (without an interstitial and extra click) over to

https://open-ui.org/components/popover.research.explainer/

josepharhar:
> This is "ok" like this, but perhaps better is a server redirect?
> 
> See this URL for an example:
> 
> https://open-ui.org/components/popup.research.explainer/
> 
> It automatically redirects (without an interstitial and extra click) over to
> 
> https://open-ui.org/components/popover.research.explainer/

Yes that is much better, I found a configuration file with redirects in it, and I replaced the selectlist and selectmenu pages with redirects.

When I ran it locally it didn't look like it was working though. Maybe I'm missing something?

mfreed7:
> > This is "ok" like this, but perhaps better is a server redirect?
> > See this URL for an example:
> > https://open-ui.org/components/popup.research.explainer/
> > It automatically redirects (without an interstitial and extra click) over to
> > https://open-ui.org/components/popover.research.explainer/
> 
> Yes that is much better, I found a configuration file with redirects in it, and I replaced the selectlist and selectmenu pages with redirects.
> 
> When I ran it locally it didn't look like it was working though. Maybe I'm missing something?

Interesting. Looking at the source files, it looks like someone changed how this was working since the last time I touched it also. I wonder who the expert would be?

Side-thing, I also notice as I look back at the popup/popover files, there are still old versions of the "popup" explainer lingering in the source tree. That's likely bad too?

tantek (approved):

josepharhar:
> I  wonder who the expert would be?

Probably @andrico1234 

> Side-thing, I also notice as I look back at the popup/popover files, there are still old versions of the "popup" explainer lingering in the source tree. That's likely bad too?

Are you talking about this? I figured it was needed to make the redirect work but we could try deleting it and see what happens? https://github.com/openui/open-ui/blob/main/site/src/pages/components/popup.research.explainer.js

mfreed7:
> > I  wonder who the expert would be?
> 
> Probably @andrico1234
> 
> > Side-thing, I also notice as I look back at the popup/popover files, there are still old versions of the "popup" explainer lingering in the source tree. That's likely bad too?
> 
> Are you talking about this? I figured it was needed to make the redirect work but we could try deleting it and see what happens? https://github.com/openui/open-ui/blob/main/site/src/pages/components/popup.research.explainer.js

No, not that one. I do think that's needed for the redirect? I meant the other popup.*.mdx files. But after reviewing them, I remember that they're all "historical" versions that are kept for posterity. No issue, I think.

josepharhar:
Ok cool. Anything else I can do to get this merged? I still don't have merging permissions.

site/src/pages/components/select.mdx
josepharhar: this is gone

PR #1105, opened 2024-10-10
Answer question of directionality

### LINKED ISSUE #1098: [switch]: Does the "on" position change between rtl and ltr?
This question comes up in [the globalization section of the switch explainer](https://open-ui.org/components/switch.explainer/#globalization).

In a `ltr` environment, the "on" position of the switch is usually on the right side of the switch. Is it the same for a `rtl` environment or should the "on" position be on the left side of the switch?

gfellerph:
Closes #1098 which resolved to make the switch adapt to flow direction of scripts and render the "on" state on the left side for `rtl` and on the right side for `ltr`.

### DIFF
--- a/site/src/pages/components/switch.explainer.mdx
+++ b/site/src/pages/components/switch.explainer.mdx
@@ -525,8 +525,15 @@ The switch element can be associated with the existing `role="switch"`.
 ### Globalization
 
 <div class="question">
-  Is thumb in the "on" position on the left in a `rtl` directional environment?
+  How should the switch behave in writing modes other than `horizontal-tb`?
 </div>
+
+The switch should adapt to the current text direction[^11].
+
+| Writing mode  | Direction     | "on" position |
+| ------------- | ------------- | ------------- |
+| horizontal-tb | Right to left | left side     |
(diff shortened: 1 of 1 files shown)


site/src/pages/components/switch.explainer.mdx
```diff
@@ -525,9 +525,16 @@ The switch element can be associated with the existing `role="switch"`.
 ### Globalization
 
 <div class="question">
-  Is thumb in the "on" position on the left in a `rtl` directional environment?
+  How should the switch behave in writing modes other than `horizontal-tb`?
```
gregwhitworth, line 528: Since you have an answer to the question, please remove the question

gregwhitworth: That said, we didn't resolve on the vertical scenario so there should probably be a new question elsewhere (maybe replace this text) and open a new issue and link to that one.

gregwhitworth (approved):
I left a comment or two and I'll assume you'll make those changes prior to landing the PR. Additionally, for your open questions you should open issues around them and link to them. That doesn't need to be a part of this PR unless you'd like it to be. Thank you for your work on this.

Issue #1104, opened 2024-09-30
Link delegation to descendant


noamr:
As per spec, the `a` element cannot nest, you cannot put links inside links.
This is important from an a11y perspective and for other reasons.
However, many websites today (e.g. Reddit, Rotten Tomatoes) implement this kind of link "nesting" on their own with somewhat verbose javascript.

Proposing to discuss finding a web platform solution for this that would take accessibility and ergonomics into account.
This solution might look like link delegation
```html
<section link="item1-link">
   <link id="item1-link" href="details?item=1">
</section>
```

In the above example, the `link` attribute of the `section` element "delegates" the link to an `a` (or `area`) descendant with the given ID. UAs should implement this by making the section's default click action delegate the click to the `item1-link` element, and they may support it in other UI such as right-clicking the `section`.

SzymonNadbrzezny:
Could someone explain to me what link nesting/link delegation would be used for? I'm looking currently at Reddit and I'm wondering whether the example of post applies, since they link to post and link from post's description are overlaping?

I personally understand the neatness of this behaviour from user perspective, but it's also a bit annoying for me when I'm trying to enter post and I'm redirected or just have image opened up.

noamr:
> Could someone explain to me what link nesting/link delegation would be used for? I'm looking currently at Reddit and I'm wondering whether the example of post applies, since they link to post and link from post's description are overlaping?
> 
> I personally understand the neatness of this behaviour from user perspective, but it's also a bit annoying for me when I'm trying to enter post and I'm redirected or just have image opened up.
> 
> EDIT: I'm your average user as far as any accesibility is concerned, but might be also biased since I'm not used to that design choice.

Check for example [walmart](https://www.walmart.ca/en/shop/summer-clearance/5396675859953?catId=10017&icid=deals_page_other_upto_50_off_clothing_small_round_hsk_95680_73W9TSAALR
). You can click on the whole item to go into the item, or click on a particular color to go to the item with that color already selected. They achieve this with an `<a>` element on top of the regular content (using `z-index`).
This is not an uncommon pattern on the web, and using `z-index` tricks for this seems suboptimal...

gfellerph:
@noamr beat me to it, but I think card components (wich can mean a lot of things) would generally be a use case for this. A lot of times you have markup like this:

```html
<div class="card">
  <img src="..." />
  <h3>Title</h3>
  <p>Description</p>
  <a href="">Read more</a>
</div>
```

[...]

PR #1109, opened 2024-10-15
Create link-area-delegation-explainer.md

noamr:
See #1104

### DIFF
--- a/link-area-delegation-explainer.md
+++ b/link-area-delegation-explainer.md
@@ -18,7 +18,7 @@ The whole card is clickable and would lead to the Reddit post, but the buttons (
 
 ## Existing ways to achieve this
 Today, achieving this is done using two alternatives, each with its own tradeoffs:
-1. Stacking: put the card *behind* the buttons, so that it catches clicks that are not handled by the buttons.
+1. Stacking: put the link *behind* the buttons, so that it catches clicks that are not handled by the buttons.
 2. Capturing: have the card capture clicks, and delegate them to the main link only if the original target is not one of the buttons.\
 
 These are imperfect and brittle ways to achieve this kind of experience, however this became a common enough pattern in the web, so perhaps it is time it became a first class citizen.
--- a/link-area-delegation-explainer.md
+++ b/link-area-delegation-explainer.md
@@ -39,7 +39,7 @@ This can mean two things in terms of UI:
 2. The card's context menu behavior is augmented with the link descendant's context menu behavior (e.g. "Open in a new tab")
 3. Potentially, the UA can expand the hit-testing area for some of the descendants, to prioritize clicking them vs. clicking on the card.
 4. This feature has no effect on keyboard navigation.
-5. At first, this feature is not exposed to the accessibility tree, and should have no effect on assistive technologies. We can research in the future whether this card/link relationship can be expressed in these technologies in a useful way.
+6. At first, this feature is not exposed to the accessibility tree, and should have no effect on assistive technologies. We can research in the future whether this card/link relationship can be expressed in these technologies in a useful way.
 
 ### Link delegation attributes
 Expressing link area delegation in HTML can look something like this:
@@ -117,6 +117,24 @@ Something like:
 
 This has nice ergonomics because the selection of link area can be done outside of the main structure of the page using CSS selectors.
 However, something about this feels limiting, as it is constrained to click areas in advance, and doesn't lend itself to future UA enhancements of this experience, such as context menus.
+
+### Use invokers
+One interesting alternative from @jaffathecake is to use [invokers](https://open-ui.org/components/invokers.explainer/) for this, e.g.:
+```html
+<section class="card" commandfor="card123-title" command="click">
+  <a href="/post?id=123" id="card123-title">Post Title</a>
+  <img>
+  <button>Join</button>
+  <button>Share</button>
+</section>
+```
+
+This has the nice effect of being more general purpose, e.g. delegating any click rather than just links, but perhaps the special-purpose link delegation has some other added value.
+
+## Open questions
+- Do link area ancestors have a default UA styles (probably not?)
+- Can you nest link areas? (probably?)
+- Do link areas also delegate hover? 
 
 ## Summary
 Link area delegation is a common pattern on the web. This proposes graduating it from a userland CSS/JS issue to a first class web platform primitive.
--- a/link-area-delegation-explainer.md
+++ b/link-area-delegation-explainer.md
@@ -7,7 +7,7 @@ We are currently incubating and exploring a new idea. Chime in!
 Historically and for very good reasons, interactive elements in the web cannot include other interactive elements, and specifically nested links.
 This is a well established [HTML ARIA guideline]([https://w3c.github.io/html-aria/](https://w3c.github.io/html-aria/#allowed-descendants-of-aria-roles)).
 
-Despite this, many websites, [Walmart](https://walmart.ca), [Reddit](https://reddit.com), [Rotten Tomatoes](https://rottentomatoes.com), to name a few prominent ones),
+Despite this, many websites, [Walmart](https://walmart.ca), [Reddit](https://reddit.com), [Rotten Tomatoes](https://rottentomatoes.com), to name a few prominent ones,
 use CSS and JS to express something that feels like nested links. 
 
 ## The card pattern
--- a/link-area-delegation-explainer.md
+++ b/link-area-delegation-explainer.md
@@ -1,0 +1,123 @@
+# Link Area Delegation
+
+## State of this explainer
+We are currently incubating and exploring a new idea. Chime in!
+
+## Overview
+Historically and for very good reasons, interactive elements in the web cannot include other interactive elements, and specifically nested links.
+This is a well established [HTML ARIA guideline]([https://w3c.github.io/html-aria/](https://w3c.github.io/html-aria/#allowed-descendants-of-aria-roles)).
+
+Despite this, many websites, [Walmart](https://walmart.ca), [Reddit](https://reddit.com), [Rotten Tomatoes](https://rottentomatoes.com), to name a few prominent ones),
+use CSS and JS to express something that feels like nested links. 
+
+## The card pattern
+The common pattern for nested links is usually a clickable "card" with internal clickable links, like this:
+<img src="https://github.com/user-attachments/assets/27e5076d-bc98-4fb8-a429-e8462ff9024c" width=400>
+
+The whole card is clickable and would lead to the Reddit post, but the buttons (Download, comments, join etc) are also clickable.
+
+## Existing ways to achieve this
+Today, achieving this is done using two alternatives, each with its own tradeoffs:
+1. Stacking: put the card *behind* the buttons, so that it catches clicks that are not handled by the buttons.
+2. Capturing: have the card capture clicks, and delegate them to the main link only if the original target is not one of the buttons.\
+
+These are imperfect and brittle ways to achieve this kind of experience, however this became a common enough pattern in the web, so perhaps it is time it became a first class citizen.
+
+## Don't panic (a11y concerns)
+The ARIA guidelines discussed earlier are still valid, so this needs to be treaded carefully. However, it doesn't seem like the websites that introduce this pattern break ARIA in any way.
+That's because this pattern doesn't modify the interactive structure of the page (AKA semantics, AKA the accessibility tree), it only gives cards enhanced behavior when they are clicked,
+allowing their clicks to be handled as if its target was the main link.
+
+One accessibility issue that can be caused by using this (and any of the existing methods) is creating clickable targets that are too close together (see [these guidelines](https://www.w3.org/WAI/perspective-videos/controls/)).
+This is indeed a concern, but it's debatable whether enabling this feature in HTML makes a difference for this concern vs. the current approaches.
+The contrary might be true, where giving this information to the UA might give it a chance to intervene when the clickable nested target is too small.
+
+## Link Area Delegation
+The proposed mental model of this problem is "the card is delegating its interactive area to one of its link descendants".
+This can mean two things in terms of UI:
+1. The default click action for the card is the default click action of the link descendant
+2. The card's context menu behavior is augmented with the link descendant's context menu behavior (e.g. "Open in a new tab")
+3. Potentially, the UA can expand the hit-testing area for some of the descendants, to prioritize clicking them vs. clicking on the card.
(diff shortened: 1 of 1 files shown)


link-area-delegation-explainer.md
```diff
@@ -0,0 +1,123 @@
+# Link Area Delegation
+
+## State of this explainer
+We are currently incubating and exploring a new idea. Chime in!
+
+## Overview
+Historically and for very good reasons, interactive elements in the web cannot include other interactive elements, and specifically nested links.
+This is a well established [HTML ARIA guideline]([https://w3c.github.io/html-aria/](https://w3c.github.io/html-aria/#allowed-descendants-of-aria-roles)).
+
+Despite this, many websites, [Walmart](https://walmart.ca), [Reddit](https://reddit.com), [Rotten Tomatoes](https://rottentomatoes.com), to name a few prominent ones),
```
jakearchibald, line 10:
```suggestion
Despite this, many websites, [Walmart](https://walmart.ca), [Reddit](https://reddit.com), [Rotten Tomatoes](https://rottentomatoes.com), to name a few prominent ones,
```

or

```suggestion
Despite this, many websites ([Walmart](https://walmart.ca), [Reddit](https://reddit.com), [Rotten Tomatoes](https://rottentomatoes.com), to name a few prominent ones)
```

link-area-delegation-explainer.md
```diff
@@ -0,0 +1,123 @@
...
+
+## The card pattern
+The common pattern for nested links is usually a clickable "card" with internal clickable links, like this:
+<img src="https://github.com/user-attachments/assets/27e5076d-bc98-4fb8-a429-e8462ff9024c" width=400>
+
+The whole card is clickable and would lead to the Reddit post, but the buttons (Download, comments, join etc) are also clickable.
+
+## Existing ways to achieve this
+Today, achieving this is done using two alternatives, each with its own tradeoffs:
+1. Stacking: put the card *behind* the buttons, so that it catches clicks that are not handled by the buttons.
```
jakearchibald, line 21:
```suggestion
1. Stacking: put the link *behind* the buttons, so that it catches clicks that are not handled by the buttons.
```

?

jakearchibald (commented):
Two questions I had after reading this (I'll let you judge if they should be in the explainer).

- What happens with nested `linkarea` elements?
- Are there any default UA styles?
- Does hovering the `linkarea` trigger hovering on the related link?

This seems very similar to `command` and `commandFor`. Could it just be:

```html
<section class="card" command="click" commandFor="card123-title">
  <a href="/post?id=123" id="card123-title">Post Title</a>
  <img>
  <button>Join</button>
  <button>Share</button>
</section>
```

In that way, you could also make the default action of a table row to toggle a checkbox within the first table cell.

noamr:
> Two questions I had after reading this (I'll let you judge if they should be in the explainer).

You mean three questions? :)
> 
> * What happens with nested `linkarea` elements?
Good question. I think it bubbles in such a way that the nearest one to the target wins. Will add it to the explainer!

> * Are there any default UA styles?
Probably not ATM

> * Does hovering the `linkarea` trigger hovering on the related link?
Will add as an open question

> 
> This seems very similar to `command` and `commandFor`. Could it just be:
> 
> ```
> <section class="card" command="click" commandFor="card123-title">
>   <a href="/post?id=123" id="card123-title">Post Title</a>
>   
>   <button>Join</button>
>   <button>Share</button>
> </section>
> ```
> 
> In that way, you could also make the default action of a table row to toggle a checkbox within the first table cell.
That's a nice thought! I wonder if `commandfor` handles all the use cases here, and if this is just about clicks

jakearchibald:
> You mean three questions? :)

Haha I thought of the third along the way

> > * Are there any default UA styles?
> 
> Probably not ATM

You'll struggle to add them later. But fwiw I don't think it needs any.

> That's an interesting thought! I wonder if `commandfor` handles all the use cases here, and if this is just about clicks

Even if it does, avoiding the need for a unique ID per table row (or whatever) is a big DX win.

Even if things don't go the `commandfor` route, being able to target buttons/checkboxes is still useful.

mfreed7:
> > > * Are there any default UA styles?
> > 
> > 
> > Probably not ATM
> 
> You'll struggle to add them later. But fwiw I don't think it needs any.

I agree with both points - hard to do later, but also, I don't think you need/want them. Feels like this doesn't change how the element looks, just how it behaves, at least by default.

> > That's an interesting thought! I wonder if `commandfor` handles all the use cases here, and if this is just about clicks

At the moment, `command`/`commandfor` are only supported on buttons. This would explicitly be for **not**-buttons, since you can't nest a link inside a button. It feels like the shape of the API is similar, but perhaps this is different enough that it'll be less confusing to have a dedicated/different attribute?

> Even if things don't go the `commandfor` route, being able to target buttons/checkboxes is still useful.

This is a good point - basically it seems like maybe you can target things that can be click targets? That raises the (most difficult) question about the name of the attribute - in that case `link` doesn't make sense anymore.

-----

I'll approve this so it can be a landed explainer. These are all great questions to raise as issues against it.

mfreed7 (approved):

jakearchibald:
> At the moment, `command`/`commandfor` are only supported on buttons. This would explicitly be for **not**-buttons, since you can't nest a link inside a button. It feels like the shape of the API is similar, but perhaps this is different enough that it'll be less confusing to have a dedicated/different attribute?

`command`/`commandfor` could easily be added to other elements, if that's the only issue.

`command`/`commandfor` is a ready to declaratively state a link between an activator and activatee. That seems to be what we're talking about with this link feature.

link-area-delegation-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+  <img>
+  <button>Join</button>
+  <button>Share</button>
+</section>
+```
+
+In the above example, the `section` element has a `linkarea` attribute, which would delegate its click to the first descendant with a `defaultlink` attribute, in this case the first `a` element.
+
+We could have different similar alternatives to this with different ergonomic tradeoffs, e.g. using ID references:
+```html
```
nmn, line 58: I think we should make this the primary proposal.

The `linkarea`/`defaultlink` approach is convenient but it creates the potential for additional edge-cases. For example, if two links within the `linkarea` claim to be the `defaultlink`.

link-area-delegation-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+For example, a use case for link area delegation is making table rows clickable.
+
+```html
+<table>
+  <tr linkarea>
+     <td>1.</td>
+     <td><a href="/details?id=1" defaultlink>Title</a>
+  </tr>
+</table>
+```
```
nmn, line 94: This use-case is a good reason to *not* limit this feature to link.

For example, the use-case might be to allow clicking the `checkbox` by clicking anywhere on a `tr`, which is not a link, but an input.

Perhaps the attribute should be `pointer-delegate`?

link-area-delegation-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+  <img>
+  <button>Join</button>
+  <button>Share</button>
+</section>
+```
+
+```css
+.card { pointer-area: contain; }
+.card a:nth-child(1) { pointer-area: expand; }
+```
```
nmn, line 116: I like the idea of using CSS for this, but I would want to overload the `pointer-events` property for this.

```css
.card { pointer-events: delegate(--delegate-name); }
.card a:nth-child(1) { pointer-delegate-name: --delegate-name); }
```

---

Thinking about this further, anchor positioning might also be a workaround that makes the CSS solution work in more situations. It probably worth adding that to the discussion.

link-area-delegation-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Use invokers
+One interesting alternative from @jaffathecake is to use [invokers](https://open-ui.org/components/invokers.explainer/) for this, e.g.:
+```html
+<section class="card" commandfor="card123-title" command="click">
+  <a href="/post?id=123" id="card123-title">Post Title</a>
+  <img>
+  <button>Join</button>
+  <button>Share</button>
+</section>
```
nmn, line 129: I love this idea! I would make this small tweak:

```suggestion
<section class="card" commandfor="card123-title" commands="click contextmenu">
  <a href="/post?id=123" id="card123-title">Post Title</a>
  <img>
  <button>Join</button>
  <button>Share</button>
</section>
```

link-area-delegation-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+This has the nice effect of being more general purpose, e.g. delegating any click rather than just links, but perhaps the special-purpose link delegation has some other added value.
+
+## Open questions
+- Do link area ancestors have a default UA styles (probably not?)
+- Can you nest link areas? (probably?)
```
nmn, line 136: You should be able nest link area but there should not be an intermediate linkarea between a link area and it's target link.

OK:
```html
<section class="card" commandfor="card123-title" commands="click contextmenu">
  <a href="/post?id=123" id="card123-title">Post Title</a>
  <img>
  <div class="card" commandfor="card123-img" commands="click contextmenu">
    <a href="/post?id=123" id="card123-img">
      <img>
    </a>
    <button>edit</button>
  </div>

  <button>Join</button>
  <button>Share</button>
</section>
```


NOT OK:
```html
<section class="card" commandfor="card123-title" commands="click contextmenu">  
  <img>
  <div class="card" commandfor="card123-img" commands="click contextmenu">
    <a href="/post?id=123" id="card123-title">Post Title</a>
    <a href="/post?id=123" id="card123-img">
      <img>
    </a>
    <button>edit</button>
  </div>
  <button>Join</button>
  <button>Share</button>
</section>
```

link-area-delegation-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+- Do link areas also delegate hover?
```
nmn, line 137: The `invokers` API should *allow* for this, but we don't have a use-case for this. We usually want to style the card's hover state separately from the link within anyway.

nmn (commented):
Did a quick edit pass and left some comments along the way. (I promised to help!)

---

The CSS proposal made me think about anchor positioning and how that makes the status quo a bit better than what we've discussed so far. It should let us avoid the problem of needing `position: static` on intermediate elements.

link-area-delegation-explainer.md
jakearchibald: I think it would be better to make the default command "auto", which is some sensible activation default. This could be toggle for dialog/popover, and default link activation for links. 

Right click behaviour is a good callout though. Also things like command click / middle click.

link-area-delegation-explainer.md
noamr: Let's add this for a discussion later rather than in the PR for the initial explainer?

link-area-delegation-explainer.md
noamr: Added a note

link-area-delegation-explainer.md
noamr: I think both of those approaches can co-exist. Anyway I'm rewording to be less opinionated about which is the primary proposal.

link-area-delegation-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+The contrary might be true, where giving this information to the UA might give it a chance to intervene when the clickable nested target is too small.
+
+## Link Area Delegation
+The proposed mental model of this problem is "the card is delegating its interactive area to one of its link descendants".
+This can mean two things in terms of UI:
+1. The default click action for the card is the default click action of the link descendant
+2. The card's context menu behavior is augmented with the link descendant's context menu behavior (e.g. "Open in a new tab")
+3. Potentially, the UA can expand the hit-testing area for some of the descendants, to prioritize clicking them vs. clicking on the card.
+4. This feature has no effect on keyboard navigation.
+6. At first, this feature is not exposed to the accessibility tree, and should have no effect on assistive technologies. We can research in the future whether this card/link relationship can be expressed in these technologies in a useful way.
```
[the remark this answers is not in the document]
noamr: Done

link-area-delegation-explainer.md
```diff
@@ -0,0 +1,141 @@
...
+<img src="https://github.com/user-attachments/assets/27e5076d-bc98-4fb8-a429-e8462ff9024c" width=400>
+
+The whole card is clickable and would lead to the Reddit post, but the buttons (Download, comments, join etc) are also clickable.
+
+## Existing ways to achieve this
+Today, achieving this is done using two alternatives, each with its own tradeoffs:
+1. Stacking: put the link *behind* the buttons, so that it catches clicks that are not handled by the buttons.
+2. Capturing: have the card capture clicks, and delegate them to the main link only if the original target is not one of the buttons.\
+
+These are imperfect and brittle ways to achieve this kind of experience, however this became a common enough pattern in the web, so perhaps it is time it became a first class citizen.
```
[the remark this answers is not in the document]
noamr: Fair enough!!

link-area-delegation-explainer.md
noamr: Added this to the q

noamr: Added this as a note

link-area-delegation-explainer.md
noamr: Let's add this to the discussion once the PR is merged

noamr:
> Did a quick edit pass and left some comments along the way. (I promised to help!)
> 
> The CSS proposal made me think about anchor positioning and how that makes the status quo a bit better than what we've discussed so far. It should let us avoid the problem of needing `position: static` on intermediate elements. The focus ring issues remain, of course.

Thanks! I think I accepted all your edits? I left some comments unresolved so that it can be a discussion after the PR is merged, it's hard to have this entire discussion on the initial merge.

link-area-delegation-explainer.md
scottaohara: it might be good to fill out those examples a bit more, or potentially provide a visual as well.  looking at just the markup, i'm having a hard time seeing these as use cases that'd necessitate this feature.

noamr:
@jakearchibald @nmn care to follow up on your great comments with PRs on the explainer? That would be easier to manage... otherwise I'll try to incorporate them later on.

scottaohara:
oh boo.  i was doing a pretty big review / suggestions to this. guess i'll just make a pr now

scottaohara:
also it appears this explainer was created in the root folder... my PR will also merge it to the correct location.

noamr:
> also it appears this explainer was created in the root folder... my PR will also merge it to the correct location.

Go for it! That would be great.
Thanks for participating, sorry for the in-flight collision.

PR #1111, opened 2024-10-16
Link delegation explainer - review updates

scottaohara:
Follow on to https://github.com/openui/open-ui/pull/1109

This PR updates _a lot_ of the explainer.

Additional accessibility callouts, use cases, questions, and clarifications have been made.

Moves the file from the root folder to the correct components folder.

cc @noamr, @nmn, @mfreed7


scottaohara:
things from my abandoned review that i didn't address in this PR:

the reddit example is a good one - though the _need_ for this with the reddit example is maybe not the best.  The Walmart example should probably be better called out - particularly the product listing example. But it might also be worth noting that Walmart's treatment is not the best (use of duplicate links, both exposed to screen readers) - but it being problematic is a positive for why a standardization of this feature would be useful.   The rotten tomatoes citation.... i honestly have no idea which pattern on that site would fit into the CSS or JS examples.... but i did see some examples of entire sections of content wrapped by hyperlinks - which prompted me to include that example in this explainer.

site/src/components/link-area-delegation-explainer.md
```diff
@@ -0,0 +1,178 @@
...
+    <p>With this Reddit example, the whole card is clickable and doing so and would navigate a user to the Reddit post. But there are also other interactive elements (up/down vote, comments, join, the sub-reddit link, etc.) are also clickable. These appear to be "nested" within the clickable card, but due to work arounds, this is not the case.</p>
+  </figcaption>
+</figure>
+
+### Existing ways to achieve clickable containers
+Today, making a container/card clickable can be achieved by a few different approaches, each with its own tradeoffs:
+1. Using the `a` element to wrap the content that needs to be clickable.  The issue with this is:
+  a. doesn't support nested interactive elements without creating accessibility issues.
+  b. using an `a` element to wrap content is not supported in certain contexts - e.g., such as trying to wrap a row within a table.  The HTML parser will remove such hyperlinks from the rendered table markup.
+  c. even without nested interactive elements, the accessible name of the hyperlink can become overly verbose, or may not be exposed as all depending on the element the hyperlink wraps.
```
scottaohara, line 30: i decided not to expand on this point beyond what i wrote here.  but maybe others might be curious / it might be a necessary rebuttal to any future feedback of "HTML already partially allows for this by wrapping content in a hyperlink".

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,191 @@
+---
+menu: Active Proposals
+name: Link Area Delegation (Explainer)
+##path: /components/...
```
scottaohara, line 4: per commit message, is this necessary?

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,191 @@
...
+- Feedback: https://github.com/openui/open-ui/issues/
+
+{/* START doctoc generated TOC please keep comment here to allow auto update */}
+{/* DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE */}
+
+## State of this explainer
+We are currently incubating and exploring a new idea. Chime in!
+
+## Overview
+Historically and for very good reasons, interactive elements on the web cannot be nested within other interactive elements, including hyperlinks.
```
nmn, line 20: This sentence is a bit awkward to read. Is this better?

```suggestion
Historically and for very good reasons, interactive elements on the web cannot be nested within other interactive elements. Examples of interactive elements include links and buttons.
```

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,191 @@
...
+Interactive elements not allowing descendent interactive elements is a well established rule in HTML for various interactive elements - where descendent interactive elements or elements with a `tabindex` attribute are specifically called out as invalid (see [the `a` element's content model](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element). Additionally, the [ARIA in HTML guideline]([https://w3c.github.io/html-aria/](https://w3c.github.io/html-aria/#allowed-descendants-of-aria-roles)) specification provides further guidance to discourage the use of interactive elements as descendent of ARIA roles which are themselves considered "widget" (interactive) roles.
```
nmn, line 21:
```suggestion
This restriction on nested interactive elements is a well established rule in HTML - where descendent interactive elements or elements with a `tabindex` attribute within other interactive elements are specifically called out as invalid (see [the `a` element's content model](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element). Additionally, the [ARIA in HTML guideline]([https://w3c.github.io/html-aria/](https://w3c.github.io/html-aria/#allowed-descendants-of-aria-roles)) further discourages the use of interactive elements as descendent of ARIA roles which are themselves considered "widget" (interactive) roles.
```

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,191 @@
...
+
+Despite this long-standing rule, there are websites which ignore it. In doing so, these sites end up with UI which will be difficult, if not blocking, for people using assistive technology, such as screen readers. On the flip-side, there are also websites such as [Walmart](https://walmart.ca), [Reddit](https://reddit.com), [Rotten Tomatoes](https://rottentomatoes.com), to name a few prominent ones, that use specific markup patterns and CSS or JS to present content that visually appears to have nested targets, as the larger container appears to be clickable - but inspecting the markup patterns would show otherwise.
```
nmn, line 23:
```suggestion
Despite this long-standing rule, there are websites which ignore it. In doing so, these sites end up with UI which will be difficult, if not unusable, for people using assistive technology such as screen readers. On the flip-side, there are websites that use CSS and/or JS creatively to create UI that *appears* to have nested interactive elements while avoiding nesting them in the markup itself in order to keep the UI accessible. Examples of such websites include [Walmart](https://walmart.ca), [Reddit](https://reddit.com), and [Rotten Tomatoes](https://rottentomatoes.com) among others.
```

nmn: Do we need a more detailed write for the "motivation" for this feature?

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,191 @@
...
+       width="400" 
+       alt="Screenshot of a Reddit topic 'card' from the r/cats subreddit. Shows time of publication, r/cats as a hyperlink, a join and more options button, followed by the title of the post and a graphic of a cat pearing out a window while its owner takes its picture. up/down vote, comment, award and share buttons follow the graphic.">
+
+  <figcaption>
+    <p>With this Reddit example, the whole card is clickable and doing so and would navigate a user to the Reddit post. But there are also other interactive elements (up/down vote, comments, join, the sub-reddit link, etc.) are also clickable. These appear to be "nested" within the clickable card, but due to work arounds, this is not the case.</p>
+  </figcaption>
+</figure>
+
+### Existing ways to achieve clickable containers
+Today, making a container/card clickable can be achieved by a few different approaches, each with its own tradeoffs:
```
nmn, line 39:
```suggestion
Today, making a container/card clickable can be achieved with a few different approaches, each with its own tradeoffs:
```

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,191 @@
...
+1. Using the `a` element to wrap the content that needs to be clickable.  The issue with this is:
+  a. doesn't support nested interactive elements without creating accessibility issues.
+  b. using an `a` element to wrap content is not supported in certain contexts - e.g., such as trying to wrap a row within a table.  The HTML parser will remove such hyperlinks from the rendered table markup.
+  c. even without nested interactive elements, the accessible name of the hyperlink can become overly verbose, or may not be exposed as all depending on the element the hyperlink wraps.
```
nmn, line 43:
```suggestion
  c. even without nested interactive elements, the accessible label of the hyperlink can become overly verbose, or may not be exposed as all depending on the children elements.
```

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,191 @@
...
+2. Using CSS to position / stack interactive elements. For instance, using CSS to put the link, or a link's `::before` or `::after` pseudo element "behind" the "nested" interactive elements, so that the link catches clicks that are not handled by these other interactive targets. The issue here is:
+  a. no standardized way to do this, with some implementations better than others - particularly such implementations that use duplicate links, creating redundancy for keyboard and/or screen reader users (even if the duplicate link is `tabindex=-1`).
```
nmn, line 45: Does expanding the `::before` or `::after` element create an additional tabbable element? Shouldn't it simply expand the hit area of the existing link or button?

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,191 @@
...
+  b. this sort of pattern often makes it difficult if not impossible for some users to select what appears to be static text.
```
nmn, line 46:
```suggestion
  b. this sort of pattern often makes it difficult if not impossible for some users to select what appears to be static text. 
  c. Such a pattern usually relies on all intermediate elements between the container/card and the primary action within to have `display: static`. Depending on the visual design requirements, this may not be feasible.
```

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,191 @@
...
+  a. no standardized way to do this, with some implementations better than others - particularly such implementations that use duplicate links, creating redundancy for keyboard and/or screen reader users (even if the duplicate link is `tabindex=-1`).
+  b. this sort of pattern often makes it difficult if not impossible for some users to select what appears to be static text. 
+3. Using JavaScript to Capture clicks: have the container element capture clicks, and delegate them to the main link only if the original target is not another interactive target.
+   a. requires the use of JS, which in itself is not a bad thing, but may undesired by some developers - leading them to use alternate patterns, or,
+   b. can lead to other UX issues, such as activation on `mousedown` rather than 'up', or people forgoing the inclusion of a nested hyperlink in favor of only using the JS, leading to the element being accessible to pointer events, but often not to keyboard or assistive techology users. 
+
```
nmn, line 50:
```suggestion
  c. can conflict with certain browser behaviours such as context menus and popup blocking.

```

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,191 @@
...
+<section class="card" commandfor="card123-title" commands="click contextmenu">
+  <a href="/post?id=123" id="card123-title">Post Title</a>
+  <img ...>
+  <button>Join</button>
+  <button>Share</button>
+</section>
+```
+
+This a nice idea as it could allow delegation of clicks to more than just "links". For instance, delegating to a `button` element instead. 
+The tradeoff here might be that this would need two attributes to declare, rather than just one. And that the `id` of the desired element to delegate to might be necessary to declare more often than the common case of extending the click-area of the first hyperlink within a container.
```
nmn, line 160: As a point of discussion, I think the need for explicit IDs is a positive.

nmn (approved):
I like the improvements made and I've suggested a couple of my own improvements. All my improvements are an effort to improve language so treat them all like suggestions and not corrections.

noamr (approved):
Great improvements, thanks! I have some follow ups for later.

site/src/pages/components/link-area-delegation-explainer.mdx
scottaohara: Name is the correct term. 
I will revise the other bit as neither of our wording is quite right for that

site/src/pages/components/link-area-delegation-explainer.mdx
scottaohara: No those don’t receive focus.  This item isn’t talking about that though, but rather implementations that use two actual links. Walmart does this and it’s rather sloppy

site/src/pages/components/link-area-delegation-explainer.mdx
scottaohara: Probably should. But I didn’t have time to get into that. That was another one of my comments that I forgot to repost from my review that didn’t get posted since the explainer got merged while I was reviewing

site/src/pages/components/link-area-delegation-explainer.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+## The clikcable container or "card" pattern
+Commonly a clickable container, often referred to as a "card", will present a graphic or other media, text, a primary call to action (link or button) as well as other supplementary interactive elements:
+
+<figure>
+  <img src="https://github.com/user-attachments/assets/27e5076d-bc98-4fb8-a429-e8462ff9024c" 
+       width="400" 
+       alt="Screenshot of a Reddit topic 'card' from the r/cats subreddit. Shows time of publication, r/cats as a hyperlink, a join and more options button, followed by the title of the post and a graphic of a cat pearing out a window while its owner takes its picture. up/down vote, comment, award and share buttons follow the graphic.">
```
scottaohara, line 31:
```suggestion
       alt="Screenshot of a Reddit topic 'card' from the r/cats subreddit. Shows time of publication, r/cats as a hyperlink, a join and more options button, followed by the title of the post and a graphic of a cat peering out a window while its owner takes its picture. up/down vote, comment, award and share buttons follow the graphic." />
```

site/src/pages/components/link-area-delegation-explainer.mdx
scottaohara: thanks.  going to rework this a bit further in an updated commit i'm working on

site/src/pages/components/link-area-delegation-explainer.mdx
scottaohara: re-reading, i realize maybe you meant that comment just in context to this section.  i'm pulling in some of your ideas / adjusting further.

i'm also changing out the rotten tomatoes example for amazon and best buy, since i could easily find this pattern on both those sites, and still wasn't quite sure where on rotten tomatoes to look.

site/src/pages/components/link-area-delegation-explainer.mdx
scottaohara: updating this to clarify, as i can see why you thought i was saying the pseudo elements would receive focus.

site/src/pages/components/link-area-delegation-explainer.mdx
scottaohara: thanks.  rewording slightly but pulling this in

site/src/pages/components/link-area-delegation-explainer.mdx
nmn: Thanks!

nmn (approved):

mfreed7 (approved):
I will rubber-stamp approve this, based on all of the existing reviews.

noamr:
I think this can be submitted? I don't have write access

PR #1125, opened 2024-11-08
Update explainer with button behaviors

mfreed7:
...

### DIFF
--- a/site/src/pages/components/customizableselect.mdx
+++ b/site/src/pages/components/customizableselect.mdx
@@ -187,7 +187,7 @@ select, ::picker(select) {
 
 To replace the select's built-in button, a native `<button>` element is used. That tells the browser that everything inside the `<button>` should be included in the in-page display of the `<select>` element. I.e. it's just a simple placeholder/container. Normal button activities and attributes, e.g. form submission, are routed to the `<select>` to handle.
 
-To ensure the behavior described above, the nested `<button>` is rendered `inert`. It is also set to `display:contents` by the UA stylesheet, so that there aren't two sets of box decorations. The button and it's descendants do not participate in the a11y tree; therefore, any attributes on the button or descendants have no effect in the a11y tree.
+To ensure the behavior described above, the nested `<button>` is rendered `inert`. It is also set to `display:contents` by the UA stylesheet, so that there aren't two sets of box decorations. The button and it's descendants do not participate in the a11y tree; therefore, any attributes on the button or descendants have no effect on a11y.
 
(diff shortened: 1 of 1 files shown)


site/src/pages/components/customizableselect.mdx
```diff
@@ -182,6 +182,13 @@ select, ::picker(select) {
 </div>
 </div>
 
+
+#### Why a `<button>`?
+
+To replace the select's built-in button, a native `<button>` element is used. That tells the browser that everything inside the `<button>` should be included in the in-page display of the `<select>` element. I.e. it's just a simple placeholder/container. Normal button activities and attributes, e.g. form submission, are routed to the `<select>` to handle.
+
+To ensure the behavior described above, the nested `<button>` is rendered `inert`. It is also set to `display:contents` by the UA stylesheet, so that there aren't two sets of box decorations. The button and it's descendants do not participate in the a11y tree; therefore, any attributes on the button or descendants have no effect in the a11y tree.
```
chrishtr, line 190:
```suggestion
To ensure the behavior described above, the nested `<button>` is rendered `inert`. It is also set to `display:contents` by the UA stylesheet, so that there aren't two sets of box decorations. The button and it's descendants do not participate in the a11y tree; therefore, any attributes on the button or descendants have no effect on a11y.
```

aleventhal (approved):
Do we need to say anything about, the button needs to be the first child element of the select and no other buttons in the select participate this way?

site/src/pages/components/customizableselect.mdx
mfreed7: Done.

mfreed7:
> Do we need to say anything about, the button needs to be the first child element of the select and no other buttons in the select participate this way?

Is this a requirement?

josepharhar (approved):

scottaohara:
having the button part being actually inert won't cause any issues in calculating the value of the select, will it?

as we had previously resolved that if someone were to do:
```
<select>
  <button> foo <selectedcontent>bar <span class="hide">baz</span></selectedcontent></button>
  <option selected>bar <span class="hide">baz</span></option>
</select>

<style>
  selectedcontent .hide { display: none }
</style>
```

that the value for the select would be returned as "foo bar" instead of the option's "bar baz" text

aleventhal:
It shouldn't. But we can prototype. I'm sure I can work around any issues.

josepharhar:
I think we should merge this. You could also remove the text about removing the split button section from the commit message since this PR no longer does that.

PR #1139, opened 2025-01-16
Update invokers explainer

lukewarlow:
- Remove left over references to `input[type=button]` as that's not supported.

- Update explanations for type="reset" and type="submit".

- Add explanation for why `input[type=button]` isn't supported.

- Add explicit type="button" to all the example buttons.


lukewarlow:
I haven't called out the `<form><button commandfor="foobar" command="--frobulate">Doesn't submit form nor frobulate foobar</button</form>` restriction, I'm not sure if that's too into the weeds to call out?

keithamus (approved):

PR #1144, opened 2025-01-24
Add the `:has-interest` pseudo class

mfreed7:
See the new text, but this adds a pseudo class for the "has interest" state.


lukewarlow:
Not going to block on this but I think :interest better matches existing pseudo classes like :hover and :focus

lukewarlow (approved):

mfreed7:
> Not going to block on this but I think :interest better matches existing pseudo classes like :hover and :focus

Yep, I totally agree. Due to the 2-rename rule, I'm going to keep this (worse) name for now, so that it can be properly renamed to `:interest` by the group. 😄 

Thanks for the review!

mfreed7:
For future reference, if anyone knows why the two numbered points starting "Simply hovering an element should not be enough to show interest " look odd, I'd love to know. I can't see that locally, and I don't know what it takes to fix the display.

PR #1153, opened 2025-02-07
Upgrade to Astro v5 and preact

lukewarlow:
This happens to improve the prefetch behaviour so that it only happens on link hover instead of on page load.


site/src/components/contributors.astro
```diff
@@ -6,7 +6,8 @@ const CONTRIBUTORS = [];
 const fetchContributors = async (page) => {
   const response = await fetch(`${CONTRIBUTORS_URL}?page=${page}`);
   const contributors = await response.json();
-  CONTRIBUTORS.push(...contributors);
+  if (Array.isArray(contributors))
```
lukewarlow, line 9: The home page crashes locally pretty quickly without this check, as GitHub API has rate limits.

site/src/components/navigation/navigation-container.jsx
```diff
@@ -1,12 +1,31 @@
-import { useStore } from '@nanostores/react'
-import React from 'react'
-import { isMenuOpen } from '../../state/menuState'
+import { useSignal } from '@preact/signals'
```
lukewarlow, line 1: Should probably just swap this for a useState no need to bring in signals just for this.

site/src/components/navigation/navigation-container.jsx
```diff
@@ -1,12 +1,31 @@
...
+    if (navRef.current) {
+      navRef.current.addEventListener('command', handleCommand)
+    }
+
+    return () => {
+      if (navRef.current) {
+        navRef.current.removeEventListener('command', handleCommand)
+      }
+    }
+  })
```
lukewarlow, line 26: This needs an empty array as the second param

mfreed7 (commented):
I'm definitely not comfortable reviewing this one - anyone else that can review and is authoritative on Astro?

site/src/components/navigation/navigation.astro
```diff
@@ -58,7 +58,7 @@ const routesToNotPrefetch = ['/research/component-matrix']
   }
 </style>
 
-<NavigationContainer client:load>
+<NavigationContainer client:only>
   <CommunityLinks className="mobile" />
```
rschristian, line 62: Does this mean the JS for `CommunityLinks` runs client-side too? It inlines a bunch of SVGs into JSX, definitely want to avoid having it in any client bundles.

lukewarlow: Good point will need to double check that.

lukewarlow: So there's 4 inlined SVGs into the JSX but that's the case anyway. Seems fine?

rschristian: Probably better off converting them into a sprite sheet & consuming with `<use>`.

Not that a few SVGs is the end of the world but that's objectively the most expensive way to use them & by a large margin too.

gregwhitworth:
Yeah, let me dig in to how to meaningfully review this in Netlify

gregwhitworth:
@lukewarlow aside from Preact I think we can close or re-purpose this PR?

gregwhitworth:
@lukewarlow friendly ping on this

PR #1155, opened 2025-02-15
Update a11y section and add FAQ about hot-key vs. focus

mfreed7:


mfreed7:
@aleventhal and @scottaohara I'd love a review of this, since it's (mostly) a11y related. I've stolen much of the text from your "Popover=Hint A11y Proposal" document, but also added a lot more about the keyboard stuff.

site/src/pages/components/interest-invokers.explainer.mdx
```diff
@@ -75,6 +82,10 @@ Mouse users will show interest in an element by mouse-hovering the element for a
 
 For keyboard users, the UA will provide a hot-key such as Alt-Down, which will "show interest" in the element. Another hot-key such as the ESC key will provide a way to trigger "loseinterest". These hot-keys must be chosen by the UA to be convenient for the user, while also not conflicting with existing UA-provided and OS-provided hot-keys.
 
+Additionally, the appearance of the focus ring will be altered slightly when focus is on elements with `interesttarget`. This allows the user to discover that something is "different" about this element, and know that the keyboard activation hot-key will allow them to show interest. Since accessibility guidelines state that color alone cannot be enough to differentiate states, this focus ring change will likely need to be something like a slightly-thicker outline, or other shape change.
+
+An alternative idea is to instead change the "location" hint that most browsers include in the lower-left corner when focusing a link element. This area currently shows the destination URL for the link, but it could have additional text added such as "press alt-down for additional information" when an element with `interesttarget` is focused.
```
aleventhal, line 87: Should it be in the bottom right corner instead? This way if it's a link, the user will continue to get the url as a hint.

site/src/pages/components/interest-invokers.explainer.mdx
```diff
@@ -247,16 +259,76 @@ Using those insets, a developer could implement a long-press activated hovercard
...
+#### Hot-key to show interest
+
+Pros:
+- No chance for "hijacking" the tab navigation cycle.
+- Fast/immediate to activate (no delays needed).
+
+Cons:
+- Not very discoverable.
+
+#### Conclusion: hot-key
```
aleventhal, line 323: If I had my druthers the conclusion would be try both and let everyone play with it until we figure out which is best, but I know this is potentially more development cycles. It might be fewer dev cycles in the end though, because it will settle the argument better if we tried both and there was a clear consensus.

site/src/pages/components/interest-invokers.explainer.mdx
```diff
@@ -75,6 +82,10 @@ Mouse users will show interest in an element by mouse-hovering the element for a
 
 For keyboard users, the UA will provide a hot-key such as Alt-Down, which will "show interest" in the element. Another hot-key such as the ESC key will provide a way to trigger "loseinterest". These hot-keys must be chosen by the UA to be convenient for the user, while also not conflicting with existing UA-provided and OS-provided hot-keys.
```
aleventhal, line 83: I know this isn't settling on a key yet, but we sould look at alt+space, ctrl+space as well.
Alt+down might be nice, admittedly. I think Windows already supports it for opening a `<select size=1>` What would happen if the select also had an interest target?

site/src/pages/components/interest-invokers.explainer.mdx
aleventhal: I think in most cases, people who are tabbing are pressing tab pretty rapidly. Admittedly this falls apart if the person is unable to press keys quickly. However, I think finding a good setting for the dwell-before-show time is probably possible, where it doesn't lead to lots of extra popovers showing, doesn't lead to extra tabbing, and is more discoverable than Alt+down.

site/src/pages/components/interest-invokers.explainer.mdx
scottaohara: agree we don't want to overwrite the url hint whenever possible.

but maybe the hints could be stacked on top of each other, or the information is presented on a single line?

it's just easier to look to one corner of the screen instead of both sides.

mfreed7: I agree with both comments - I've added a sentence that it could also go on the right to avoid interfering.

site/src/pages/components/interest-invokers.explainer.mdx
mfreed7: Good questions. I've added alt-space and ctrl-space as other ideas down in the `### Keyboard` section.

We'd have to make sure we resolve any overlapping hotkey problems, such as the `<select>` one you mention. I'd think we shouldn't override existing hotkeys, right? So the `<select>` would win in this case.

site/src/pages/components/interest-invokers.explainer.mdx
mfreed7: I think that's the point of this explainer and the conversation (tomorrow) about this. I'd rather not implement *both*, if we can resolve it ahead of time.

mfreed7:
I'd love to get this landed before we talk about it live tomorrow, if anyone's willing to give it an approval.

site/src/pages/components/interest-invokers.explainer.mdx
aleventhal: Another option for links is to put the url, then a separator, and then the hot key info.

https://github.com/openui/open-ui/pull/1155  —  Press Alt+down for more information

site/src/pages/components/interest-invokers.explainer.mdx
aleventhal: Maybe we implement Alt+down/space/whatever, and then a javascript bookmarklet to open on a timer, that people can tweak/play with as a prototype to what it would be like. Just so that we can show that it actually is or isn't a good idea.

site/src/pages/components/interest-invokers.explainer.mdx
aleventhal: Makes sense to me.

mfreed7 (commented):
Thanks!

PR #1168, opened 2025-02-20
Add dark mode to site

lukewarlow:


lukewarlow:
These colours have currently mostly been picked at random based on my perception of their contrast so its possible they dont meet requirements (particularly some of the links look slightly dark compared to others so need to check that).

But I think it's a big win having dark mode.

mfreed7 (approved):
Seems ok to me, but I'm definitely not an expert in this area. I'll click Approve, but it'd be good to get someone more authoritative to review, maybe?

lukewarlow:
@gregwhitworth do you have thoughts on the use of modern styling such as light-dark() Vs using preference media queries (basically what browser support do we care about?)

Also do you have thoughts on this change in general.

gregwhitworth:
@lukewarlow first off thanks for tackling this. Looking at a few of the pages we have contrast issues that will have to be addressed to land this. The component matrix:

image

Code blocks. Here is dark mode:

image

Versus light mode:

image

The anatomy aren't visible due to leveraging transparency and assuming a white background:

image

The [.note classes](https://deploy-preview-1168--open-ui.netlify.app/components/checkbox.research.findings/) likewise end up with poor contrast:

image

Same with the Slider ["Parts and Concepts" table](https://deploy-preview-1168--open-ui.netlify.app/components/slider.research.parts/)

image


Looking at the various High Contrast solutions and or Color Pages they seem fine. So I think if you tweak the above two we should be in a good spot.

gregwhitworth (changes requested):
See prior comment, a few scenarios end up with poor contrast due to these changes

lukewarlow:
I've changed everything that I could find, I don't understand what you mean about anatomies? They seem to render fine (per your screenshot?)

lukewarlow:
@gregwhitworth I believe I've addressed all the concerns you've raised, other than the bit about anatomies which I don't understand what you mean? They appear to work correctly to me at least.

site/src/components/component-name-matrix.css
```diff
@@ -30,16 +30,16 @@
...
 }
 .component-name-matrix .cell.found {
     background: rgba(0, 255, 0, 0.1);
-    color: rgb(0, 64, 0);
+    color: light-dark(rgb(0, 64, 0), rgb(0, 191, 0));
 }
 .component-name-matrix .cell.not-found {
     background: rgba(255, 0, 0, 0.1);
-    color: rgb(64, 0, 0);
+    color: light-dark(rgb(64, 0, 0), rgb(191, 0, 0));
```
gregwhitworth, line 42: Image

I think this view doesn't look great. My instinct is to just have this have the background blead through. It'll require a media query of course but I'm red green color blind and this hurts my eyes (may just be me though)

site/src/components/component-name-matrix.jsx
```diff
@@ -11,8 +11,9 @@ const SORT_NEXT = {
   MATCH_COUNT: SORT_OPTIONS.COMPONENT_NAME,
 }
 
-const ComponentNameMatrix = (props) => {
-  const [sort, setSort] = useState(SORT_OPTIONS.COMPONENT_NAME)
+function ComponentNameMatrix() {
+  // const [sort, setSort] = useState(SORT_OPTIONS.COMPONENT_NAME)
```
gregwhitworth, line 15: This removal of `useState` breaks the button for sorting. Please put this back.

gregwhitworth (changes requested):

Issue #1165, opened 2025-02-19
[Range] Range Input Styling API Proposal


brechtDR:
## Introduction

This proposal introduces a styling API for HTML range inputs, allowing developers to customize the appearance of both single and dual-range inputs. This enhancement aims to provide greater design flexibility while maintaining the benefits of native form controls.

## Use Cases

* Creating range inputs that match a site's design system
* Implementing custom visualizations for range selection
* Enhancing the usability of range inputs with visual feedback

## Real-world examples

* [tweakers.net](https://tweakers.net/smartphones/vergelijken/)
* [theswex.com](https://theswex.com/nl/whisky-catalogus/flessen)
* [bol.com](https://www.bol.com/nl/nl/l/boeken/8299/)
* [asadventure.com](https://www.asadventure.com/nl/c/wandelen/wandelschoenen/categorie-a-laag.html)
* [toyota.de](https://www.toyota.de/neuwagen?sortOrder=modelIndex)
* [vanhelden.be](https://www.vanhelden.be/nl/schrijfgerief)
* [shein.com](https://euqs.shein.com/risk/challenge?captcha_type=905&redirection=https://euqs.shein.com%2Fpdsearch%2Ftable+lamp%2F%3Fici%3Ds1%26ref%3Deur%26rep%3Ddir%26ret%3Deuqs&risk-id=E2720750169660833792)
* [airbnb.com](https://www.airbnb.com/)
* [devexpress.com](https://docs.devexpress.com/WPF/DevExpress.Xpf.Editors.TrackBarEdit)

<details>
<summary>Examples of styled range sliders</summary>

![Image](https://github.com/user-attachments/assets/e6e23bff-0619-4859-a26a-2db224ea65d0)
![Image](https://github.com/user-attachments/assets/156514f6-fcf6-40d8-b2e0-148e5f4b23cf)
![Image](https://github.com/user-attachments/assets/847f892a-f106-4f80-ab07-dab7ea4729c2)
![Image](https://github.com/user-attachments/assets/ac4d0354-d8f1-401b-b0e1-08fadf8da609)
![Image](https://github.com/user-attachments/assets/abfa65a9-4c4d-4e62-9706-1152ea0c1da8)
![Image](https://github.com/user-attachments/assets/8236c63e-70fc-458b-a991-bc73bc6e0a20)
![Image](https://github.com/user-attachments/assets/e494bfd9-9917-4272-8852-c282d778b0a1)
[...]

Issue #1164, opened 2025-02-19
[Range] Dual Handle Range Input Proposal


brechtDR:
I though, why not start a proposal here. If there is interest by the group, I don't mind writing a first explainer as well.

## Introduction

This proposal introduces a new feature for HTML range inputs: the ability to select two values within a given range using a single input element. This enhancement aims to provide a more intuitive and space-efficient way to select a range of values, commonly used in scenarios such as price filters, age ranges, or time spans.

## Use Cases

* Price range selectors in e-commerce filters
* Age range inputs in demographic forms
* Time range selectors for scheduling applications
* Data range selectors in analytics tools

## Real-world examples:

* [tweakers.net](https://tweakers.net/smartphones/vergelijken/)
* [theswex.com](https://theswex.com/nl/whisky-catalogus/flessen)
* [bol.com](https://www.bol.com/nl/nl/l/boeken/8299/)
* [asadventure.com](https://www.asadventure.com/nl/c/wandelen/wandelschoenen/categorie-a-laag.html)
* [toyota.de](https://www.toyota.de/neuwagen?sortOrder=modelIndex)
* [vanhelden.be](https://www.vanhelden.be/nl/schrijfgerief)
* [airbnb.com](https://www.airbnb.com/)
* [shein.com](https://euqs.shein.com/risk/challenge?captcha_type=905&redirection=https://euqs.shein.com%2Fpdsearch%2Ftable+lamp%2F%3Fici%3Ds1%26ref%3Deur%26rep%3Ddir%26ret%3Deuqs&risk-id=E2720750169660833792)
* Extra codepen example that does allow crossing: [codepen.io](https://codepen.io/utilitybend/pen/wBvMmZp)

<details>
<summary>Screenshots of Real-world examples</summary>

![Image](https://github.com/user-attachments/assets/d9bf24c9-a183-4bcb-9fe1-6fa53bbf6fad)
![Image](https://github.com/user-attachments/assets/e718b637-0484-46c1-b80b-3cca3a5793d5)
![Image](https://github.com/user-attachments/assets/82a7d460-9154-4134-a860-464ebef78db4)
![Image](https://github.com/user-attachments/assets/e647fae7-06ff-4c84-8a61-732c60bd36e4)
![Image](https://github.com/user-attachments/assets/01c9638d-c106-42e0-934c-34c9017a83b9)
![Image](https://github.com/user-attachments/assets/678a724c-0dd0-4870-a898-e92b2acd5db8)
[...]

PR #1173, opened 2025-03-03
[Enhanced Range Input] Start of explainer

brechtDR:
## First draft

This is an explainer that could be used as a starting point for working towards an enhanced range input. The idea of this explainer was based on feedback of the following issues and listening to developer needs.

- https://github.com/openui/open-ui/issues/1165
- https://github.com/openui/open-ui/issues/1164

This explainer should be looked at as a first draft to open up discussions, the questions in the bottom of this explainer can be used as a guideline for further discussion. 

Also, feel free to comment on writing-style. As this is my first explainer from scratch, I tried to mimic the tone from other explainers. (usually I write articles, and I'm not a native speaker)

## Opinionated

The idea is highly opinionated by me and was created from a front-end perspective (how I would love to see it). Of course it did mix in some developer feedback and opinions as stated before. Wether this is feasible from an engineering standpoint is something to be looked at as a group, same goes for accessibility and styling concerns.


site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,475 @@
...
+
+<RangeAnatomy/>
+
+The standardized anatomy includes:
+
+- `::range-track`: The main track along which the thumb(s) move.
+- `::range-fill`: The filled portion of the track, typically between the minimum value and the thumb (or between thumbs for multi-handle ranges).
+- `::range-thumb`: The draggable handle(s) used to select values.
+- `::range-segment`: Sections of the track between handles in multi-handle ranges.
+- `::range-tick`: Optional tick marks along the track for value representation. (when datalist is paired).
```
mfreed7, line 90: nit: the trailing `.` should be inside the `()` and `When` should be capitalized.

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,475 @@
...
+
+We propose extending the range input to support multiple handles, starting with dual-handle support as the most common use case. This will be achieved through new HTML attributes and corresponding JavaScript APIs.
+
+### Accessibility Enhancements
+
+To improve accessibility, especially for multi-handle ranges, we propose making individual thumbs focusable and providing clear audio cues for screen readers.
+
+### Backwards Compatibility
+
+The reason for opting in is to ensure backward compatibility with existing range inputs. Since current implementations are a mess, it feels like a safer bet. Browsers that do not support the new features will fall back to the current behavior of range inputs, and it opens up the window of targeting support via CSS (@supports).
```
mfreed7, line 115: I think I'd just say that backwards compatibility is a required part of any web platform API, and the existing range input is widely used. You can link to this:

https://chromestatus.com/metrics/feature/timeline/popularity/34

And I'd include "Progressive Enhancement" to the title of this section, since that's a nice benefit here.

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,475 @@
...
+### Datalist Integration
+
+We propose standardizing and enhancing the integration of `<datalist>` with range inputs. This will allow for consistent implementation of tick marks and predefined values across browsers, especially when used with the `appearance: base-range` styling.
+
+Key features:
+
+1. Standardized rendering of `<datalist>` options as tick marks on the range track.
+2. Ability to style tick marks using the `::range-tick` pseudo-element.
+3. Ability to style labels of tick marks using the `::range-tick-label` pseudo-element.
+4. Support for `<datalist>` with multi-handle range inputs.
```
mfreed7, line 128: How? Just values in the options like the `value` syntax? E.g. `<option>23.2  34.5</option>`?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,475 @@
...
+    <option value="50" label="50%"></option>
+    <option value="75" label="75%"></option>
+    <option value="100" label="100%"></option>
+</datalist>
+```
+### CSS Properties and Pseudo-elements
+
+To address the current fragmentation and provide a unified styling API, we propose the following pseudo-elements:
+
+1. `::range-track`: Represents the main track of the range input.
```
mfreed7, line 192: This section feels kind of duplicated with the one above. That one sounds like it means to explain the *structure* of the control, while this one is about the pseudos. But perhaps that could be made more clear?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,475 @@
...
+2. `::range-fill`: Represents the filled portion of the track.
+3. `::range-thumb`: Represents the draggable handle(s).
+4. `::range-tick`: Represents individual tick marks on the range input.
+5. `::range-tick-label`: Represents the label associated with each tick mark. This pseudo-element should be able to handle the `content` property for custom content. By default, the content is the value of the tick.
+6. `::range-segment`: Represents sections of the track between handles in multi-handle ranges.
+
+These pseudo-elements can be used in combination with other CSS selectors, such as `:nth-child()`, to provide granular control over individual ticks, labels, and segments.
+
+The following is an example usage, more details on this are explained below:
```
mfreed7, line 201: Maybe put this in a sub-section?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,475 @@
...
+}
+```
+
+### JavaScript API
+
+1. `valueAsNumber`: Returns an array of numbers for multi-handle ranges.
+2. `setRangeValue(index, value)`: Sets the value for a specific handle, without the need for a string.
+3. `handleValues`: Returns an array of handle positions as percentages.
+
+Example usage:
```
mfreed7, line 268: ditto

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,475 @@
...
+```
+
+**HTML:**
+
+```html
+<input type="range" handles="3" min="0" max="100" value="25,50,75">
+```
+## Considerations and Open Questions
+
+1. Should we support more than two handles, or limit to dual-handle ranges? If more, is there a maximum?
```
mfreed7, line 461: A few of these have been discussed in issues, e.g. this one. Perhaps link to those, where they exist?

mfreed7 (approved):
LGTM with some comments. Thanks for putting this together.

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,475 @@
+---
```
lukewarlow, line 1: We have a slider research page already at some point it might be worth deciding what to do with that (potentially link to it from this explainer?)

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,475 @@
...
+
+Example usage:
+
+```css
+input[type="range"][handles="3"] {
+    appearance: base-range;
+}
+
+/* Style the first segment (between the start and the first handle) */
+input[type="range"][handles="3"]::range-segment:nth-child(1) {
```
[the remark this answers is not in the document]
brechtDR: Keith told me something like this might be possible. It's more versatile in that situation for sure. That way you can have a general/global thumb styling and nth-child specific if wanted/needed

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,475 @@
...
+<input type="range" min="0" max="100" list="tickmarks">
+<datalist id="tickmarks">
+    <option value="0" label="0%"></option>
+    <option value="25" label="25%"></option>
+    <option value="50" label="50%"></option>
+    <option value="75" label="75%"></option>
+    <option value="100" label="100%"></option>
+</datalist>
+```
+### CSS Properties and Pseudo-elements
```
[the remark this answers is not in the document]
brechtDR: Not sure if custom properties are the way to go. Was rather thinking about a way with the new attr() capabilities. Certainly worth looking into though. I'll have to do a bit of research for this. Great idea in general though.

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: I will add it to the questions. The idea is to turn those into issues.

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: I think the ticks should actually remain the same in that case no? it should just be able to move 2 handles alongside of them

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: I did add the previous research at the top of the page as Luke suggested. Not sure if that's enough? At the top of the page I did link a few discussions

site/src/pages/components/enhanced-range-input.explainer.mdx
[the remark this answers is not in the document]
lukewarlow: Modern pseudo elements do have the ability to match various pseudo-classes especially if they're element backed pseudo elements which essentially act as a real element for selector rules (https://drafts.csswg.org/css-pseudo-4/#element-like)

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,479 @@
...
+
+- [Background](#background)
+- [Goals](#goals)
+- [Current State of Range Styling](#current-state-of-range-styling)
+- [Proposed Solution](#proposed-solution)
+    - [Standardized Anatomy](#standardized-anatomy)
+    - [Styling API](#styling-api)
+    - [Multi-Handle Support](#multi-handle-support)
+    - [Accessibility Enhancements](#accessibility-enhancements)
+    - [Backwards Compatibility](#backwards-compatibility)
```
myakura, line 19:
```suggestion
    - [Progressive Enhancement](#progressive-enhancement)
```

site/src/pages/components/enhanced-range-input.explainer.mdx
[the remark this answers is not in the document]
lukewarlow: https://drafts.csswg.org/css-forms/#the-control-value-function - this is what I'm refering to.

keithamus (approved):
Great work!

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,479 @@
...
+
+### Accessibility Enhancements
+
+To improve accessibility, especially for multi-handle ranges, we propose making individual thumbs focusable and providing clear audio cues for screen readers.
+
+### Progressive Enhancement
+
+The reason for opting in is to ensure [backward compatibility](https://chromestatus.com/metrics/feature/timeline/popularity/34) with existing range inputs. Since current implementations are a mess, it feels like a safer bet. Browsers that do not support the new features will fall back to the current behavior of range inputs, and it opens up the window of targeting support via CSS (@supports).
+
+For multi-handle ranges, browsers that do not support the feature will treat the input as a standard single-handle range, using the first value provided in the `value` attribute
```
myakura, line 120: > For multi-handle ranges, browsers that do not support the feature will treat the input as a standard single-handle range, using the first value provided in the `value` attribute

does that how current browsers behave? if not, that won't be either backward compatibility nor progressive enhancement. no?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,479 @@
...
+}
+
+input[type="range"][handles="3"]::range-segment:nth-child(3) {
+    background-color: #3357FF;
+}
+```
+
+### JavaScript API
+
+1. `valueAsNumber`: Returns an array of numbers for multi-handle ranges.
```
myakura, line 269: current spec and implementations expects a single Number so i'm not sure reusing this getter for array of numbers

lukewarlow: Yeah properties are cheap enough to add I think we should just use a new one rather than overloading the existing one. This property can just return the first value in the index?

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: Could be discussed wether this is a good idea or not. I'd personally like to tackle that question via an issue

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,479 @@
...
+    - [Considerations and Open Questions](#considerations-and-open-questions)
+
+## Background
+
+The `<input type="range">` element has long been a part of HTML forms, providing a slider control for selecting numeric values. However, its current implementation lacks flexibility in styling and functionality, leading developers to often resort to custom implementations. These custom solutions can result in reduced performance, reliability, and accessibility compared to native form controls.
+
+The web development community has [expressed a need for more customizable and feature-rich range inputs](https://github.com/w3c/csswg-drafts/issues/4410#issuecomment-1087244943), particularly for use cases such as price range selectors, date/time range pickers, and other multi-value selection scenarios.
+
+- [Examples from CSSWG issue](https://github.com/w3c/csswg-drafts/issues/4410#issuecomment-1087244943)
+- [Examples from Open UI issue](https://github.com/openui/open-ui/issues/1165)
```
gregwhitworth, line 36: I recommend linking out to the [parts analysis](https://open-ui.org/components/slider.research.parts/) I did a while back in Open UI

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -0,0 +1,479 @@
...
+
+We propose extending the range input to support multiple handles, starting with dual-handle support as the most common use case. This will be achieved through new HTML attributes and corresponding JavaScript APIs.
+
+### Accessibility Enhancements
+
+To improve accessibility, especially for multi-handle ranges, we propose making individual thumbs focusable and providing clear audio cues for screen readers.
+
+### Progressive Enhancement
+
+The reason for opting in is to ensure [backward compatibility](https://chromestatus.com/metrics/feature/timeline/popularity/34) with existing range inputs. Since current implementations are a mess, it feels like a safer bet. Browsers that do not support the new features will fall back to the current behavior of range inputs, and it opens up the window of targeting support via CSS (@supports).
```
gregwhitworth, line 118: I personally would prefer to see a new element than placing **all** of the parts within a psueudo elements. I think there is value in landing the pseudo elements as well as landing a new element as there will be usecases that won't be achievable by pseudo elements along (this will apply checkbox, radio, password, color, file, and many others).

site/src/components/range-anatomy.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import { AnatomyWrapper, Host, Part, Slot } from './anatomy-components'
+
+const RangeAnatomy = () => {
+  return (
+    <AnatomyWrapper>
+      <Host name="input type=range">
```
gregwhitworth, line 7: Where is the range value(s)?

gregwhitworth (approved):
I don't want to block the PR so that we can begin discussion. But this definately overlaps a LOT with the [CSS Forms work](https://drafts.csswg.org/css-forms-1/) which I do think the Open UI group should review and speak into.

Additionally, while this will unlock improved styling there are functional usecases that may not be feasible. To be clear, we can do this in a layered approach and I like that you're bringing in tick marks.

At any rate, thanks for the work on this and please begin opening issues for your open questions.

site/src/components/range-anatomy.jsx
gregwhitworth: In [CSS forms this function](https://drafts.csswg.org/css-forms-1/#the-control-value-function) looks akin to what I'm referring to but we would follow what we've done with `select` and produce an element for this.

site/src/pages/components/enhanced-range-input.explainer.mdx
lukewarlow: I think the question is, is this statement true or not? I would expect it not to be but I could be wrong.

brechtDR: you're both right, it doesn't. I'll change this

PR #1176, opened 2025-03-06
Change from a new keyboard-focusable property to use interactivity

mfreed7:
It seems better to use a new value for `interactivity` rather than inventing a new property.

### DIFF
--- a/site/src/pages/components/interest-invokers.explainer.mdx
+++ b/site/src/pages/components/interest-invokers.explainer.mdx
@@ -17,7 +17,7 @@ layout: ../../layouts/ComponentLayout.astro
 - [HIDs and Interest](#hids-and-interest)
   - [Mouse](#mouse)
   - [Keyboard](#keyboard)
-    - [The `keyboard-focusable` CSS property](#the-keyboard-focusable-css-property)
+    - [`interactivity: no-keyboard`](#interactivity-no-keyboard)
   - [Touchscreen](#touchscreen)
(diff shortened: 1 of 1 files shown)


lukewarlow (approved):

site/src/pages/components/interest-invokers.explainer.mdx
```diff
@@ -17,7 +17,7 @@ layout: ../../layouts/ComponentLayout.astro
 - [HIDs and Interest](#hids-and-interest)
   - [Mouse](#mouse)
   - [Keyboard](#keyboard)
-    - [The `keyboard-focusable` CSS property](#the-keyboard-focusable-css-property)
+    - [`interactivity: no-keyboard`](#interactivity-no-keyboard)
```
lukewarlow, line 20: Not going to block on bikeshedding but I think it'd be good to name this in terms of focusability rather than keyboard?

For example, no-keyboard could mean no keydown or keyup events fire on it? No escape key etc.

mfreed7: That's fair. Perhaps `not-focusable` ?

mfreed7: ...but the problem is that "keyboard" should be in there somewhere, because it's really only about **keyboard** focusability. So perhaps `not-keyboard-focusable`. Long, but descriptive.

PR #1177, opened 2025-03-10
[Enhanced Range Input] Add author to range input

brechtDR:
Forgot to add myself as author, noticed that this is in other explainers as well.


lukewarlow (approved):
Might be worth adding yourself as a codeowner for the file so if others make changes you're able to approve them (something that we've done for invokers for example)

gregwhitworth (approved):

PR #1215, opened 2025-05-13
Update menu explainer

scottaohara:
This PR is a bunch of suggested changes / clarifications for the content of the explainer.  Fleshing out some ideas, updating some alt text to be more accurate, stuff like that.

There is no associated issue for this work.

### DIFF
--- a/site/src/pages/components/menu.explainer.mdx
+++ b/site/src/pages/components/menu.explainer.mdx
@@ -171,7 +171,7 @@ Additionally - command is a single element, where as this proposal is introducin
 
 The [`<menu>`](https://html.spec.whatwg.org/multipage/grouping-content.html#the-menu-element) element already exists, but it is very barebones. It represents a "toolbar" but is exposed no differently than an unordered list containing list items, each list item then representing a command (represented by a button element) that the user can perform or activate. There is a lot of nuance between the various menu patterns (menubar, navigation menu, toolbar) that this one element cannot differentiate. It has no special keyboard behavior nor unique ARIA roles mapping beyond the implicit roles of the elements it represents. Due to these legacy restrictions, we recommend not modifying it and instead depend on the new elements `<menubar>` and `<menulist>`.
 
-You can read more about what the `<menu>` element is and isn’t, [here](https://www.scottohara.me/blog/2021/10/21/menu.html).
+You can read more about what the `<menu>` element is and isn’t, in Scott O'Hara's blog post [Semantic `<menu>` context](https://www.scottohara.me/blog/2021/10/21/menu.html).
 
 
 ### Popover semantics for `<menulist>`
--- a/site/src/pages/components/menu.explainer.mdx
+++ b/site/src/pages/components/menu.explainer.mdx
@@ -81,22 +81,23 @@ model. See https://github.com/openui/open-ui/issues/1194 for more discussion abo
 
 `<menubar>`
 
-* Menubar should be used to represent a series of `<menuitem>`s that can then open submenus of commands.
-* In a menubar, a menuitem’s most common purpose is to invoke submenus (menulists).
-* An **in-page** menu that can contain a list of `<menuitem>`
-* Menuitems within a menubar are displayed horizontally by default.
-* Supports :disabled, disabled attribute
+* Menubars commonly represent a series of `<menuitem>`s that are used to open their own submenus (menulists) of commands.
+* Unlike menulists, menubars are rendered **in-page** rather than as a popover, so their menuitems are persistently available.
+* Menubars, and thus their menuitems, are commonly rendered horizontally.
+* Supports :disabled, disabled attribute. If a menubar is disabled then all of its descendent menuitems are also disabled.
 
 `<menulist>`
 
-* Menulist should be used to represent submenus of commands.
-* In a menulist, a menu item's most common purpose is to invoke actions (commands) or other submenus (menulists).
-* A **popover** menu that can contain a list of `<menuitem>`
-* Anchored to the button which is declaratively set up to open it
-* Attributes
-    * disabled, open, name?
+* Menulist should be used to represent a listing, or submenu listing of commands.
+* In a menulist, a menuitem's most common purpose is to invoke actions (commands) or other submenus (menulists).
+* Menulists are commonly be rendered as a **popover**. Submenus are also commonly rendered as adjacent popovers.
+* The menulist popover is anchored to the button or menuitem which invoked it
+* Supports :disabled, disabled attribute. If a menulist is disabled then all of its descendent menuitems are also disabled.
 
-**Question: Is `<fieldset>` necessary to group `<menuitem>` together?**
+**Question:** are the `open` or `name` attributes needed for `menubar` or `menulist`?
+???
+
+**Question: Is `<fieldset>` necessary to group `<menuitem>` elements together?**
 
 Yes. While, we can use the <code>["name"](https://html.spec.whatwg.org/C#attr-fe-name)</code>
 attribute to logically group exclusive checkable `<menuitem>`s together, to support captioned groups
@@ -106,7 +107,7 @@ without a `<fieldset>`, they are grouped together under their parent `<menulist>
 but are logically separate and are not checkable.
 
 Note, `<fieldset>` is a more general grouping element than `<optgroup>`, which we briefly
-considered, but decided against since it is used for grouping options.
+considered, but decided against since it is used, and named, specifically for grouping options.
 
 Our proposal includes adding a `checkable` attribute on `<fieldset>`, to group `<menuitem>`s as
 either radios or checkboxes, instead of introducing new `<menuitemcheckbox>` and `<menuitemradio>`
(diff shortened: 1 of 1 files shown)


site/src/pages/components/menu.explainer.mdx
```diff
@@ -140,21 +143,21 @@ Similarly to customizable select, we expect everything to be wrapped inside `<me
...
 
 The `<select>` element is a form-associated element and contains `<option>` elements. Each of these `<option>` elements must have a server-readable value. As a user chooses an option, specific events are triggered (change, input) and a new value is set.
 
-On the other hand, the proposed `<menulist>` element should be used to display a list of commands/actions. Each `<menuitem>` should be thought of as a command that is activable. That action could be to open a new nested menulist element or to trigger a specific command. It must adhere to [ARIA guidelines for menu and menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/).
+On the other hand, the proposed `<menulist>` element should be used to display a list of commands/actions. Each `<menuitem>` should be thought of as a command that is activable. That action could be to open a new nested menulist element or to trigger a specific command. It is not generally expected for the invoking element's value or visible label to be modified when activating a `<menuitem>`.  It must adhere to ARIA guidelines for [menu and menubar patterns](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/).
```
hidde, line 160: not part of your change, but I don't think “activable” is a word? And it's not very clear what “command that is activable” is. Maybe we can fix at the same time

```suggestion
On the other hand, the proposed `<menulist>` element should be used to display a list of commands/actions. Each `<menuitem>` should be thought of as performing a specific action. That action could be to open a new nested menulist element or to trigger a specific command. It is not generally expected for the invoking element's value or visible label to be modified when activating a `<menuitem>`.  It must adhere to ARIA guidelines for [menu and menubar patterns](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/).
```

site/src/pages/components/menu.explainer.mdx
```diff
@@ -168,7 +171,7 @@ Additionally - command is a single element, where as this proposal is introducin
 
 The [`<menu>`](https://html.spec.whatwg.org/multipage/grouping-content.html#the-menu-element) element already exists, but it is very barebones. It represents a "toolbar" but is exposed no differently than an unordered list containing list items, each list item then representing a command (represented by a button element) that the user can perform or activate. There is a lot of nuance between the various menu patterns (menubar, navigation menu, toolbar) that this one element cannot differentiate. It has no special keyboard behavior nor unique ARIA roles mapping beyond the implicit roles of the elements it represents. Due to these legacy restrictions, we recommend not modifying it and instead depend on the new elements `<menubar>` and `<menulist>`.
 
-You can read more about what the menu element is and isn’t, [here](https://www.scottohara.me/blog/2021/10/21/menu.html).
+You can read more about what the `<menu>` element is and isn’t, [here](https://www.scottohara.me/blog/2021/10/21/menu.html).
```
hidde, line 174: maybe we could improve link text?

```suggestion
You can read more about what the `<menu>` element is and isn’t, in Scott O'Hara's blog post [Semantic `<menu>` context](https://www.scottohara.me/blog/2021/10/21/menu.html).
```

site/src/pages/components/menu.explainer.mdx
```diff
@@ -81,22 +81,25 @@ model. See https://github.com/openui/open-ui/issues/1194 for more discussion abo
 
 `<menubar>`
 
-* Menubar should be used to represent a series of `<menuitem>`s that can then open submenus of commands.
-* In a menubar, a menuitem’s most common purpose is to invoke submenus (menulists).
-* An **in-page** menu that can contain a list of `<menuitem>`
-* Menuitems within a menubar are displayed horizontally by default.
-* Supports :disabled, disabled attribute
+* Menubars commonly represent a series of `<menuitem>`s that are used to open their own submenus (menulists) of commands.
+* Unlike menulists, menubars are rendered **in-page** rather than as a popover, so their menuitems are persistently available.
```
hidde, line 85: we could consider saying “part of normal flow” here instead of “in-page”? (like [in CSS](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Introduction#normal_layout_flow))

hidde (commented):
Very helpful additions and explanations. Added a few ideas.

site/src/pages/components/menu.explainer.mdx
scottaohara: that's fine by me.  i just kept that term as it was previously written

site/src/pages/components/menu.explainer.mdx
```diff
@@ -81,22 +81,25 @@ model. See https://github.com/openui/open-ui/issues/1194 for more discussion abo
...
-* In a menulist, a menu item's most common purpose is to invoke actions (commands) or other submenus (menulists).
-* A **popover** menu that can contain a list of `<menuitem>`
-* Anchored to the button which is declaratively set up to open it
-* Attributes
-    * disabled, open, name?
+* Menulist should be used to represent a listing, or submenu listing of commands.
+* In a menulist, a menuitem's most common purpose is to invoke actions (commands) or other submenus (menulists).
+* Menulists are commonly rendered as a **popover**. Submenus are also commonly rendered as adjacent popovers on larger viewports. 
+  * On smaller viewports, submenus may overlay or replace the higher-level menulist from which they were invoked. 
+  * A "back" menuitem may be rendered to allow a user to "navigate up" to the higher-level menulist.
```
dizhang168, line 95: Great idea. I have seen such use cases especially for mobile web app, where there isn't enough space to show multiple nested submenus (or to have better UX).

site/src/pages/components/menu.explainer.mdx
```diff
@@ -81,22 +81,25 @@ model. See https://github.com/openui/open-ui/issues/1194 for more discussion abo
...
+* The menulist popover is anchored to the button or menuitem which invoked it
+* Supports :disabled, disabled attribute. If a menulist is disabled then all of its descendent menuitems are also disabled.
 
-**Question: Is `<fieldset>` necessary to group `<menuitem>` together?**
+**Question:** are the `open` or `name` attributes needed for `menubar` or `menulist`?
+???
```
dizhang168, line 100: The `open` attribute is useful if we want to default the menulist as open. But maybe we should avoid that, and be explicitly to only open with user activation behaviors. It is definitely not necessary on `menubar`.

The `name` attribute is meant to group radio menuitem elements together. It is probably not necessary for `menubar` and `menulist`.

site/src/pages/components/menu.explainer.mdx
dizhang168, line 160: nit: I wrote ARIA guidelines originally, but that is wrong. Should be changed to APG recommendations.

dizhang168 (approved):

site/src/pages/components/menu.explainer.mdx
hidde:
```suggestion
* Unlike menulists, menubars are rendered part of normal flow rather than as a popover, so their menuitems are persistently available.
```

site/src/pages/components/menu.explainer.mdx
domfarolino: Is this included because it's something you envision the platform providing itself, or are you just enumerating common developer-level options for menu rendering?

site/src/pages/components/menu.explainer.mdx
domfarolino: I agree that `name` isnt' needed if we're doing the `<fieldset>` approach, which is our current plan. I don't see a need for `open` attribute unless people want to be able to set a menu's open status with JavaScript besides just activating it. I guess I'd prefer to wait for such asks to come on, and stick with the simpler, smaller proposal of not including this attribute initially.

lukewarlow: One thing to be clear on, we cannot do open as a two way binding. Aka it cannot reflect state like details and dialog. So if we want an initiallyopen we should plan to name it accordingly.

domfarolino: Why not?

lukewarlow:
@domfarolino is this still needed or have the clarifications been made elsewhere?

site/src/pages/components/menu.explainer.mdx
scottaohara: wow... completely dropped the ball on noticing these questions...

this was added because i think it'd be great if UAs provided this sort of styling - or at least made it very easy to provide this sort of styling by default. 

styling for menu popups for small viewports / zoomed in viewports is often overlooked, which can make it rather difficult if not impossible for low vision users to interact with the menuitems when authors haven't been diligent at making sure their menus actually appear in the viewport.

but, i realize this might be more effort than all browsers want to take on.  so I said "may" to provide wiggle room.  maybe mobile browsers do this and desktop browsers leave it up to authors to remember.   shrug

scottaohara:
@lukewarlow, comparing many of my changes to the published explainer it looks like this is still relevant.

site/src/pages/components/menu.explainer.mdx
lukewarlow: > Why not?

Because it's a flustercuck for details and especially dialog, this will have many of the issues dialog has. I believe Mason also shares this opinion.

mfreed7: I share this opinion. The `open` attribute should be considered a bad idea. The issues stem from the fact that modifying DOM (to add or subtract an attribute) is prohibited or just bad in several places, such as removing an element from the DOM. The best pattern is:

1. A pseudo class like `:open` to reflect state
2. (If needed) an attribute that controls the *initial* state, such as `initiallyopen`.

lukewarlow:
Looks like this has a conflict preventing merge. Would be good to get this merged. @mfreed7 / @domfarolino

mfreed7:
Oh shoot, I forgot about this one. @dbaron to see how to best merge it with the existing/updated explainer.

gregwhitworth:
@dbaron friendly ping on the above cc: @mfreed7

PR #1218, opened 2025-05-14
Add button mockups and more survey details to interesttarget explainer

mfreed7:
This adds more mockups of interesttarget activation patterns for `<button>` elements. It also adds another survey image that includes details of the keyboard and touchscreen accessibility of each site.


keithamus (approved):
Nice!

PR #1222, opened 2025-05-16
Add speculation rules section to interesttarget

mfreed7:
This was brought up by the TAG, so I'm adding some details.


lukewarlow (approved):

mfreed7:
Thanks for the quick reviews on these, @lukewarlow !

lukewarlow:
You're welcome! They're small enough to do from the phone app.

PR #1223, opened 2025-05-22
Add source property to openable options bag.

lukewarlow:
- Also refactor explainer slightly.

This property is needed for some of the discussed behaviours such as focus management to work correctly when using the imperative APIs.


mfreed7 (commented):
This one is now old, right?

Issue #1158, opened 2025-02-15
[openable] focus management behaviour


lukewarlow:
We need to further define the focus management behaviour for openables. While they don't take focus by default as described [here](https://open-ui.org/components/openable.explainer/#focus-management). Should they adjust focus order such that the openable is next in the tabindex (like popover)?

mfreed7:
I'm tentatively removing Agenda+ - feel free to put it back if you'd like to discuss this again.

PR #1224, opened 2025-05-22
[Openable] focus management updates

lukewarlow:
Built atop #1223 

Closes https://github.com/openui/open-ui/issues/1158

### DIFF
--- a/site/src/pages/components/openable.explainer.mdx
+++ b/site/src/pages/components/openable.explainer.mdx
@@ -56,6 +56,30 @@ TODO
 
 This section lays out the full details of this proposal. If you'd prefer, you can **[skip to the examples section](#example-use-cases) to see the code**.
 
+### IDL
+
+The proposed IDL for the Openable API is:
+
+```webidl
+[Exposed=Window]
+partial interface HTMLElement {
+  attribute boolean openable;
+
+  undefined openOpenable(optional OpenOpenableOptions options = {});
+  undefined closeOpenable();
+  boolean toggleOpenable(optional ToggleOpenableOptions options = {});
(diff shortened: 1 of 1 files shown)


site/src/pages/components/openable.explainer.mdx
```diff
@@ -56,6 +56,30 @@ TODO
...
+### IDL
+
+The proposed IDL for the Openable API is:
+
+```webidl
+[Exposed=Window]
+partial interface HTMLElement {
+  attribute boolean openable;
+
+  undefined openOpenable(optional OpenOpenableOptions options = {});
```
mfreed7, line 68: `openOpenable` is a mouthful, but I don't have anything better for you.

site/src/pages/components/openable.explainer.mdx
```diff
@@ -56,6 +56,30 @@ TODO
...
+  boolean toggleOpenable(optional ToggleOpenableOptions options = {});
+  boolean? defaultOpen;
+}
+
+dictionary OpenOpenableOptions {
+  HTMLElement source;
+}
+
+dictionary ToggleOpenableOptions : OpenOpenableOptions {
+  boolean force;
```
mfreed7, line 79: Do you need a source in the toggle options?

lukewarlow: It extends OpenOptions so inherits source from there

site/src/pages/components/openable.explainer.mdx
lukewarlow: Fully agree we'll sort it when we renamed the whole API because openable itself is pretty meh.

PR #1231, opened 2025-06-07
Stop explicitly setting font size for website

veyndan:
The default font size for most browsers is 16px already, but some people (including myself) like to toggle the default font size in the browser settings. Explicitly setting the default font size to 16px in CSS overrides that preference.


keithamus:
@veyndan if you're comfortable doing so could you please join the OpenUI CG and sign the IPR? https://labs.w3.org/repo-manager/pr/id/openui/open-ui/1231

veyndan:
@keithamus Done! 😄

PR #1236, opened 2025-06-10
Remove curly braces around note

josepharhar:
I'm not sure if the curly braces were supposed to do something interesting, but they are causing a build failure.


mfreed7 (approved):
Nope, just meant to be curly braces. Thanks!

PR #1239, opened 2025-06-20
Add "option 3.5" to the always-expanding interestfor touchscreen section

mfreed7:
This is another idea, which I consider a "v2" feature, that gives back a bit of the discoverability via an `::interest-hint` pseudo element.

### DIFF
--- a/site/src/pages/components/interest-invokers.explainer.mdx
+++ b/site/src/pages/components/interest-invokers.explainer.mdx
@@ -5,7 +5,7 @@ layout: ../../layouts/ComponentLayout.astro
 ---
 
 - [Mason Freed](https://github.com/mfreed7), [Keith Cirkel](https://github.com/keithamus)
-- Last updated: May 16, 2025
+- Last updated: June 20, 2025
 
 ## Table of Contents
 {/* DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE */}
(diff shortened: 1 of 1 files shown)


site/src/pages/components/interest-invokers.explainer.mdx
```diff
@@ -229,6 +229,32 @@ This is **likely the most straightforward approach that still meets all of the r
...
+}
+```
+
+When the `::interest-hint` pseudo element has a value set for `content` on an element with the `interestfor` attribute, the pseudo-element is generated as a button following the element. Typically, this is rendered as a small `(i)` icon or similar. When the user taps (**not** long-presses) on this icon, interest is shown in the originating element, showing the hovercard.
+
+In this mode, long-pressing the originating element still brings up the context menu, which still has the additional "View more info" item that can also be used to show interest. And tapping the originating element (not the `::interest-hint` pseudo element) still activates the element normally. So everything else continues to work as before, just with an easier way for the user to show interest, without long pressing.
+
+| `<a>` - long-press | `<button>` - long-press |
+|-------|------------|
+| <img src='/images/interesttarget-mockup-pseudo-element.gif' height="600" alt='Mockup of the pseudo element approach on links' /> | Not typical, since long-press on the button already shows interest immediately |
```
dbaron, line 255: I think you need to add this image, no?

site/src/pages/components/interest-invokers.explainer.mdx
```diff
@@ -229,6 +229,32 @@ This is **likely the most straightforward approach that still meets all of the r
...
+  /* On touch devices, show the interest-hint pseudo */
+  .my-link::interest-hint {
+    content: '' / 'profile preview';
+    background-image: url('icon.png');
+    /* styles */
+  }
+}
+```
+
+When the `::interest-hint` pseudo element has a value set for `content` on an element with the `interestfor` attribute, the pseudo-element is generated as a button following the element. Typically, this is rendered as a small `(i)` icon or similar. When the user taps (**not** long-presses) on this icon, interest is shown in the originating element, showing the hovercard.
```
dbaron: Maybe not now, but it might be good to be clearer about this "Typically" statement, in particular, whether it's the author that's responsible for making it happen (which I think is the case) or whether it's part of the platform.

site/src/pages/components/interest-invokers.explainer.mdx
mfreed7: Oops! Fixed.

site/src/pages/components/interest-invokers.explainer.mdx
mfreed7: Good point. I added a bit more, clarifying that it'll be up to the developer.

PR #1250, opened 2025-07-31
Add [Reflect] extended attributes to explainer webidl

lukewarlow:
This is now defined in HTML spec so we should make use of it for extra clarity.


keithamus (approved):

keithamus:
@dbaron, @gregwhitworth, @mfreed7, @scottaohara, and/or @travisleithead this will require a review from one of you in order to be merged.

mfreed7:
> This is now defined in HTML spec so we should make use of it for extra clarity.

You could include a link to https://html.spec.whatwg.org/#using-reflect-via-idl-extended-attributes here

PR #1259, opened 2025-08-12
beforefocus/focusNext() explainer.

mshoho:


site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+## document.focusNext()
+
+`document.focusNext()` method just exposes the default action the browsers have for `Tab` and
+`Shift+Tab` presses.
+
+The method's signature should be like:
+
+```ts
+focusNext(contextElement?: HTMLElement, direction?: FORWARD | BACKWARD);
```
frehner: Does it feel weird to do `focusNext(undefined, "BACKWARD")`?

mshoho: Well, it's just an initial draft. It could either an object for options, or it could be two methods (`focusNext()` and `focusPrev()`). I suppose an object would be more future-proof in case some more options would be useful in the future.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+## Introduction
+
+The browsers give very little options for implementing complex keyboard navigation
+patterns. We are currently stuck with workaround techniques like roving tabindex and/or
+dummy focusable elements to redirect keyboard focus. These techniques require modifying the
+DOM on the fly and have limitations. They cause reflows and style recomputations, and
+could be hard or impossible to apply in some particular scenarios. For example, you might
+be owning a parent list component, but not owning the contents of the list items. Or
+you have to embed a cross-origin iframe you are unable to poke inside.
```
keithamus: I'd suggest rewording this initial paragraph to describe the problem statement more concisely:

```suggestion
Browsers include complex heuristics for determining the next element to focus, but these aren't
exposed to web developers, leaving them to implement these heuristics from scratch, often requiring
DOM modificatons, or writing code that causes reflows and style recomputations. It can be hard or
sometimes impossible to apply in some particular scenarios (for example, you might have control of
a parent list component, but not the contents of the list items).
```

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+While having additional new high level abstractions to handle keyboard navigation (like,
+for example, [focusgroup](./focusgroup.explainer)) sounds plausible, in reality it
+would be hard or impossible to align on the APIs that would cover all necessary cases —
+there is a multitude of different applications and no single standardized approach for them.
+It would also involve a huge body of work in the browser engine.
```
keithamus: This seems to talk down the focusgroup explainer which doesn't feel right to me. I'd suggest:



```suggestion
While having additional new high level abstractions to handle keyboard navigation (such as the
[focusgroup](./focusgroup.explainer)) will be useful, there are a multitude of different
applications for focus, and lower level primitives will give web developers more flexibility.
```

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -0,0 +1,137 @@
...
+   move focus to the first focusable element inside the iframe (if the iframe has focusable
+   elements) or past the iframe (if there is nothing to focus in the iframe) — the same thing
+   `Tab` press default action does.
+* The same logic applies for mirroring `Shift+Tab` behaviour when `direction` is set ot `BACKWARD`.
+* `document.focusNext()` should have no effect if the page is not focused or the focus is in
+   the address bar.
+* When `trapped` is true, the focus should cycle inside the page, skipping the address bar.
+
+Implementing `document.focusNext()` will be fairly straightforward as no new behaviour needs
+to be added, just the one that is already present should be reused.
```
keithamus: I don't think this sentence is necessary, I also worry it might read as presumptuous to make a claim about how straightforward an API would be to implement.

keithamus (commented):
@mshoho thanks so much for working on this! I think the kernel of this really points to a pain that developers have, and I think there's something worth exploring further here. 

I'm not going to go too into discussing the API surface, nor my personal concerns - we can do that as we go further through the process - but I have some initial thoughts just on the prose:

keithamus:
Also @mshoho, before merging, you'll need to [join the OpenUI Community Group](https://www.w3.org/community/open-ui/join) and sign the IPR commitments which allow us to publish your work.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -0,0 +1,137 @@
...
+for example, [focusgroup](./focusgroup.explainer)) sounds plausible, in reality it
+would be hard or impossible to align on the APIs that would cover all necessary cases —
+there is a multitude of different applications and no single standardized approach for them.
+It would also involve a huge body of work in the browser engine.
+
+Instead of high level APIs, this proposal focuses on having two fairly simple lower level
+functions which would arguably allow the application developers to implement whatever patterns
+they prefer without working around the browser limitations and impossibilities.
+
+Namely, one method called `document.focusNext()` and one event called `beforefocus`.
```
smhigley, line 29: This is an interesting idea, and there are definitely quite a few use cases for wanting to move focus not to a specific element, but to the next or previous focusable element from a given starting point.

I also wonder if exposing a specific query selector for focusable elements (e.g. `:focusable`) might also accomplish some of the same use case but in a more flexible way that fits in well with existing focus management techniques.

One drawback is this would not cover sending focus into or out of iframes, which would be very useful to be able to do, providing it doesn't break encapsulation.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -0,0 +1,137 @@
...
+`FOCUS_BACKWARD` means that the element is about to be focused because `Shift+Tab` is
+pressed or `document.focusNext()` is called with `direction` set to `BACKWARD`.
+
+`PROGRAMMATIC` means that the element is about to be focused because the application has
+called `element.focus()`.
+
+`AUTOFOCUS` means that the element is about to be focused because of the `autofocus` DOM
+attribute.
+
+`UI` means that the element is about to be focused because the user has clicked (or tapped)
```
smhigley, line 109: I think this would need careful consideration around whether it increases the ability to detect assistive tech. For example, screen readers often trigger focus based on virtual cursor location. That behavior wouldn't fit any of these categories, and if it were put under e.g. programmatic or UI focus it would be easy to detect based on the lack of preceding `element.focus()` calls or click / tap events.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+Implementing `document.focusNext()` will be fairly straightforward as no new behaviour needs
+to be added, just the one that is already present should be reused.
+
+
+## beforefocus Event
+
+`beforefocus` event is dispatched _before_ the browser wants to focus an element. Most important
+part is that `preventDefault()` of the `beforefocus` event should prevent the element from
+receiving focus.
```
smhigley, line 87: It might help to add a use case for this part of the proposal

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -0,0 +1,137 @@
...
+`contextElement` is the starting point from which the browser should start looking for the next
+element to focus. It is optional and equals to `document.activeElement` by default.
+
+`direction` is the direction of the search. `FORWARD` to get the `Tab` press behaviour and
+`BACKWARD` to get `Shift+Tab` press behaviour. It is also optional and equals to `FORWARD` by
+default.
+
+`trapped` means that instead of going to the address bar after the last element, the search
+should continue from the beginnig of the page (or from the end of it in case of moving
+backwards). False by default.
```
smhigley, line 57: This is more a question for folks who work on browsers -- are there any security concerns around authors being able to programmatically send focus into the browser chrome when `trapped` is false?

site/src/pages/components/beforefocus.explainer.mdx
keithamus: I don't think engines can reasonably implement something like `:focusable` as it depends a lot on the state of, for e.g. the current focus.

site/src/pages/components/beforefocus.explainer.mdx
keithamus: Yes I don't think this part of the proposal is something engines will reasonably want to implement (but I am hesitant to push back on too much of this as a first draft).

site/src/pages/components/beforefocus.explainer.mdx
smhigley: @keithamus do you have more context on that? I'd be interested (though maybe it's tangential to this). I was thinking it'd essentially be a browser-maintained version of this, which is a bit of a pain to maintain from an author's perspective since we need to update it whenever a new attribute or element is added that affects focus, and even this doesn't account for CSS like `display: none`:

```js
export const FOCUSABLE_SELECTOR = [
    "a[href]",
    "button:not([disabled])",
    "input:not([disabled])",
    "select:not([disabled])",
    "textarea:not([disabled])",
    "*[tabindex]",
    "*[contenteditable]",
    "details > summary",
    "audio[controls]",
    "video[controls]",
].join(", ");
```

keithamus: Absolutely! 

[Here's the ~500LOC Firefox uses to determine the next element to tab to](https://searchfox.org/mozilla-central/source/dom/base/nsFocusManager.cpp#4226-4705). It includes things that the selector does not and cannot represent such as:

 - traversing in and out of shadow DOMs
 - fixups for when a popover has been shown using a declarative `<button popovertarget>` button, so that...
   - the button can skip over focusable content between it and the popover, so the next focusable element after the button is the first focusable element in the popover.
   - when you move focus beyond the last element in a popover it should focus on the next focusable element after the invoker 
 - Skipping content without layout (it iterates over the `PrimaryFrame` which can be though of as the rendered pixels). As you rightly say your selector list won't account for a button which is `display:none`, `visibility:hidden`, or any of the other ways it has a collapsed frame. This would be possible to emulate roughly with something like `checkVisibility()` in script.
 - Counts for lots of other small oddities/esotera like adding `tabindex` to a slot, `<area>` elements, native browser elements, etc.
 - It also has to do all of this cross process so it can find the next tabbable element in an iframe, which may or may not be cross origin (therefore in an isolated process).

So for `:focusable` to be implemented, it would depend on the thing currently in focus, plus the visibility of the next element, and every time focus changes or frankly a whole _bunch_ of DOM changes happen, we'd need to reinvalidate all of that, for an unbounded number of elements (remember `:focus`/`:focus-within` is just as complex but only matches ~one element). We'd also have to worry about "cycles", for example `:focusable{display:none}` gets a little recursive because `:focusable` would make something non-focusable, which would invalidate that rule, thereby making it `:focusable` again, and so on...

mshoho: The iframes situation is a huge part of reasoning in my proposal.

Currently, the browser's default behaviour for Tab can cross the iframe boundaries. When nothing is left to focus in the iframe, you hit Tab — the browser will move focus to the next element in the parent application. Or to the address bar. This type of behaviour cannot be achievable with any kind of selector.

But exposing the Tab press default action for programmatic invokability would be sufficient enough (together with beforefocus event) — you don't need the reference to cross origin element, you just need to be able to tell the browser to do what it currently does for Tab press. And it should not be a security concern.

It should not be more complicated than it has to be.

The idea is not to find next focusable element in the page — that is currently doable. The idea is to be able to _move_ to the next focusable element programmatically the same way the browser does with Tab default action — penetrating the boundaries of iframes and being able to go to the address bar.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: The screen reader's focusing should be in FOCUS_FORWARD/FOCUS_BACKWARD category — the same as tabbing (depending on the previous element position), not in UI or PROGRAMMATIC.

And I already do detect scenarios when the focus is moved without clicks/taps to enable keyboard navigation state when screen readers move focus. Nothing new from that side here — you can already listen to `focusin` and detect that it's not called programmatically and without clicks.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: There should be a secure way. Like it could be tied to a trusted Tab press (you could have a couple of ticks after the actual Tab press when you can invoke the going to the address bar behaviour).

I currently already have a hack called moveOutWithDefaultAction — when Tab is pressed in the middle of the page, you can synchronously create an input in the end of the DOM, focus it, and allow the default action to move you to the address bar. Without it, every focus trap in a modal dialog locks you in a page — you basically cannot use the keyboard to go to the address bar and type another URL before you close the modal. None of the libraries that provide focus trap implement it (other than Tabster), but I think it's important to have.

keithamus:
Thanks for signing the IPR @mshoho. I think it would be worth us taking each of the open conversations to their conclusion, before we merge this. I am keen to get this in and start the process of refinement and socialising with browser implementers!

site/src/pages/components/beforefocus.explainer.mdx
keithamus: I think this section of the explainer will need a lot more refinement, discussion and exploration. As a first draft this seems fine to me but I imagine this will change significantly with further iterations. I will leave it up to @mshoho as to how much it changes in this PR, though.

site/src/pages/components/beforefocus.explainer.mdx
keithamus: I think we can conclude this one then. Perhaps as a refinement in a follow up we can discuss the relative trade-offs and include the idea of "what about a `:focusable` pseudo class?" in the explainer, summarising this conversation for others to read.

site/src/pages/components/beforefocus.explainer.mdx
keithamus: @mshoho I'm curious why you need the workaround - focus trap in a native modal dialog always allows the user to tab round to the browser's UI. Are you using custom focus trap code?

Either way, we can leave this in for the initial draft if you'd like - you tell me. But I don't know how successful we'll be at getting this part of the feature request implemented.

mshoho: There are scenarios when native modal is not enough.

For example, native dialog cannot be multicontainer. Like you have a dialog with overflow:hidden on its contents (or — to make it even more tedious — overflow:hidden on some random container up the DOM). Inside that dialog you have a custom popup, or suggest, or something else that wants to visually be able to go outside of the overflow:hidden boundaries. And you need the focus to be circling inside the modal _and_ the popup as if they are a consecutive whole. And sometimes you want it to be able to go to the address bar (if it's a normal dialog) sometimes you don't (like it's a message deep inside a threaded conversation and once you have intentionally spent time to keyboard-navigate inside it — it's more convenient to have you locked inside the message, because it's too easy to accidentally lose context otherwise — you can look at the MS Teams chat message with the reactions popup as example).

If you render the popup inside the dialog, it will be cropped, so you have to render it outside, close to body — you start hitting the native dialog limitations. It can technically be somehow solved if you own the entirety of your application — you can recompose and readjust the containers. But that stops being really possible once you have big enough application with a lot of developers and/or when you have to reuse some third party components you cannot easily adjust. And even if you own it all, it is not always possible — in the chat message example above when you want to modalize the message — you cannot move it out of the conversation context and the popup will be cropped by the conversation container.

All that is actually one of the reasons I am suggesting these low level changes instead of high level abstractions. I have spent quite some time designing and implementing those high level abstractions myself (with a reasonable success), and I would love to eventually see them in the browser, but realistically even the seemingly simple and straightforward things like modal dialog are not that simple when it comes to corner cases, and once they are in the browser it is even harder to adjust them without breaking backwards compatibility.

So, let's better have something that would allow to be flexible and allow to avoid hacks. And higher level abstractions could eventually follow.

site/src/pages/components/beforefocus.explainer.mdx
smhigley: @mshoho the issue around AT detection with FOCUS_FORWARD and _BACKWARD is the same as the others -- if a user triggers it with a screen reader, it would not be preceded by either a tab key event or a `focusNext()` call, making this an additional detection vector.

I agree with Keith though, this can be revisited in further iterations.

site/src/pages/components/beforefocus.explainer.mdx
smhigley: @keithamus thanks for the thorough explanation and code link! That's really helpful, +1 on capturing it somewhere in an explainer.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: I have added a couple of usage examples.

mshoho:
> Great work @mshoho!
> 
> Let's merge this, and perhaps we can put some time in the agenda to discuss it. Would you be comfortable attending one of the virtual meetings and presenting on this topic @mshoho? The next one will be Thursday at 20:00CEST.

@keithamus certainly! Please send the join link my way and how much time I have to present it. Thanks!

PR #1260, opened 2025-08-15
Scoped focusgroup explainer

janewman:
A proposal for a more scoped version of focusgroup. This proposal differs from the [original explainer](https://open-ui.org/components/focusgroup.explainer/) in the following ways:
* Usage of the attribute is strictly limited to select aria roles.
* No CSS support
* Grid support is moved to be a future consideration.

Please see the "Alternative Scoping Mechanism" for another way to scope the scenarios where focus group can be used **without** breaking the norm of aria being a semantic-only set of attributes. If there is consensus that this approach is favored, I will flip the proposal to favor that approach instead.


site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -0,0 +1,1093 @@
...
+  <button type="button" tabindex="-1">Bold</button>
+  <button type="button" tabindex="-1">Italic</button>
+  <button type="button" tabindex="-1">Underline</button>
+</div>
+```
+What to notice:
+* `focusgroup` enables down/right arrow keys to advance the focus to the next focusable element in
+   DOM order (up/left keys will move the focus in reverse) just like in the native radio group.
+* `tabindex` is still required to make the `<span>` elements focusable (the `focusgroup` attribute
+   doesn't make any elements focusable by default, including the `<div>` it's declared on).
```
myakura: > * `tabindex` is still required to make the `<span>` elements focusable (the `focusgroup` attribute
>   doesn't make any elements focusable by default, including the `<div>` it's declared on).

I don't see the span element; is that a leftover of some change?

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -0,0 +1,1093 @@
...
+* edge wrapping (`wrap`) so users can cycle continuously.
+* last-focused memory on re-entry. State and selection changes (e.g., toggling Bold) remain normal button semantics, `focusgroup` never infers or changes selection.
+
+In this next example, the author is using a
+[tab control pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) where the tab activation
+behavior is decoupled from selection ("manual tab activation"):
+
+```html
+<div role="tablist" focusgroup="inline wrap no-memory" aria-label="Common Operating Systems">
+  <button id="tab-1" type="button" role="tab" aria-selected="false" aria-controls="tabpanel-1" tabindex="-1">Mac</button>
```
myakura:
```suggestion
  <button id="tab-1" type="button" role="tab" aria-selected="false" aria-controls="tabpanel-1" tabindex="-1">macOS</button>
```

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: Thanks for catching! this was indeed a holdover from a previous change, removed this line as it doesn't apply in any sense (buttons are focusable be default), and since subsequent examples already call out the need for authors to use tabindex for non-focusable elements.

lukewarlow:
A discussion I would like to have or at least something to acknowledge on this is, we could just build these patterns into HTML in their entirety. We have select which handles this (at least to some extent) already. We'll have listbox select once that bit is finished being specced.

There's separately a menulist and menubar proposal: https://open-ui.org/components/menu.explainer/

We could feasibly take some of menubar and make a toolbar element (if anything toolbar seems a simpler thing to tackle).

So based on 
https://github.com/openui/open-ui/blob/c5b73c8e32b9df818d95ee963228e024c03cff6e/site/src/pages/components/scoped-focusgroup.explainer.mdx#eligible-roles

That leaves tablist and future grid left to do.

From that perspective it could be argued that grid should be the thing that we do tackle upfront and the rest are less needed now?

To be clear I don't wish for this to sound disparaging and I'm absolutely not saying what you've done is bad. I just think it's something worth discussing and acknowledging (in future iterations not blocking this PR).

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -0,0 +1,1187 @@
...
+    <first-thing role="gridcell" focusgroup="grid-cell">…</first-thing>
+    <cell-container role="none">
+      <my-cell role="gridcell" focusgroup="grid-cell">…</my-cell>
+      <my-cell role="gridcell" focusgroup="grid-cell">…</my-cell>
+    </cell-container>
+    <cell-container role="none">
+      <my-cell role="gridcell" focusgroup="grid-cell">…</my-cell>
+      <my-cell role="gridcell" focusgroup="grid-cell">…</my-cell>
+    </cell-container>
+  </my-row focusgroup="grid-row">
```
keithamus:
```suggestion
  </my-row>
```

janewman:
**Update incoming**
After discussing this in today's Aria WG meeting, @smhigley had what I think is a great addition to the "Alternative Scoping Mechanism" that I will be updating this PR with shortly.

### Usage as previously explained in this proposal.
```html
<div id="one" role="radiogroup" focusgroup="wrap" aria-label="Text formatting">
  <div role="radio" aria-checked="false" tabindex="-1">Bold</div>
  <div role="radio" aria-checked="true" tabindex="0">Italic</div>
  <div role="radio" aria-checked="false" tabindex="-1">Underline</div>
</div>
```
In this approach, role is required to be present and match a supported role. **This breaks norms around aria being semantic-only.**
### Alternative scoping approach that keeps aria as semantic-only (preferred)
```html
<div id="two" focusgroup="radiogroup wrap" aria-label="Text formatting">
  <div role="radio" aria-checked="false" tabindex="-1">Bold</div>
  <div role="radio" aria-checked="true" tabindex="0">Italic</div>
  <div role="radio" aria-checked="false" tabindex="-1">Underline</div>
</div>
```
In this example, `focusgroup` will "implicit the specified role into the element", in this case, the element with id of "two" will have the role of `radiogroup`.
### Sarah's Addition to the alternative scoping approach
```html
<div id="two" focusgroup="radiogroup wrap" aria-label="Text formatting">
  <div aria-checked="false" tabindex="-1">Bold</div>
  <div aria-checked="true" tabindex="0">Italic</div>
  <div aria-checked="false" tabindex="-1">Underline</div>
</div>
```
With this addition, children whose focus is managed by `focusgroup` will also have their appropriate role set, ensuring:
1. Authors are guided into following accessible patterns.
2. Less attributes for authors to set explicitly.

## Upcoming changes
Before meeting today, I had intended on publishing an update that makes the second approach (alternative scoping mechanism) the default used in the document, but I think Sarah's suggestion is quite compelling, and will be moving forward with this approach, with the others listed in an "alternatives considered" section.

janewman:
> A discussion I would like to have or at least something to acknowledge on this is, we could just build these patterns into HTML in their entirety. We have select which handles this (at least to some extent) already. We'll have listbox select once that bit is finished being specced.
> 
> There's separately a menulist and menubar proposal: https://open-ui.org/components/menu.explainer/
> 
> We could feasibly take some of menubar and make a toolbar element (if anything toolbar seems a simpler thing to tackle).
> 
> So based on https://github.com/openui/open-ui/blob/c5b73c8e32b9df818d95ee963228e024c03cff6e/site/src/pages/components/scoped-focusgroup.explainer.mdx#eligible-roles
> 
> That leaves tablist and future grid left to do.
> 
> From that perspective it could be argued that grid should be the thing that we do tackle upfront and the rest are less needed now?
> 
> To be clear I don't wish for this to sound disparaging and I'm absolutely not saying what you've done is bad. I just think it's something worth discussing and acknowledging (in future iterations not blocking this PR).

@lukewarlow, I appreciate you raising this this, and added a small section in explainer to directly address it: "Why not just add new native elements to cover these patterns?"

While I do love to see these common patterns getting native support, I still think that focusgroup has a role to play as well. I believe that developers will continue to have a need for building their own controls to add behaviors or styling not possible in the native version provided by the user agent, especially thinking about web components. In addition, being lighter weight, I think focusgroup can also help fill the gaps as we wait for these larger proposals to be standardized and implemented, and when new patterns emerge, focusgroup could be an easy first step in unifying behavior.

Let me know what you think here, what I am getting wrong or missing? I appreciate the feedback!!

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -0,0 +1,1183 @@
...
+
+## Differences from the original explainer
+[Original focusgroup explainer.](/components/focusgroup.explainer)
+### CSS support is now a future consideration
+Applying `focusgroup` (and related behaviors) via CSS is out of scope for this explainer and deferred for a possible, backwards-compatible future addition.
+
+### Grid support is now a future consideration
+Applying `focusgroup` to grid-like structures (2-D navigation) is out of scope for this explainer and deferred for a possible, backwards-compatible future addition.
+
+### Focusgroup is now scoped to specific scenarios
```
sorvell, line 121: This restriction seems confusing, complex, and limiting.

Putting aria information outside the role or aria-* attributes seems novel in a confusing way. Is this done in any other scenario? Normalizing consistent behavior of aria roles is a larger, likely very worthy, project which should be considered separately from this feature.

Encoding all of this information in one attribute that also has impact on how the children are managed is very complex. Do explicit roles override role inference? What are the rules there?

The role restriction is also pretty limiting. Providing this feature separate from just enhancing specific platform elements facilitates its use in custom element scenarios which may fit aria roles in ways that need bespoke handling.

janewman: Thanks for raising these concerns!

**Child role inference**
I added a clarified “Role inference” section, an example, and an Open Question explicitly asking whether we should (a) keep it, (b) gate it behind a future modifier (e.g. no-infer / or add later), or (c) drop it for v1. It’s purely a shorthand: explicit roles always take precedence, and it never replaces richer native semantics. If this still feels like overly confusing or controversial, I would be fine deferring this part of the behavior it to “Future Considerations”, so we ship the core improvement first.

**Why have focusgroup take (or imply) a role at all**
The [original un-scoped proposal](https://open-ui.org/components/focusgroup.explainer/) let authors create arrow-key loops in arbitrary “div soup” which risked confusing keyboard users and screen reader flows, as well as raising additional issues around how to communicate these new "controls" to screen readers. These issues are what I understand to be the biggest blockers to moving the original forwards, and when considering what mechanisms we can use to inform screen readers, role seems like a straightforward choice. This allows us to constrain the new behaviors to well-understood composite patterns anchors behavior in existing APG semantics, meaning that ATs can keep using current role-based heuristics, with no new mode switching or announcements needed to have the expected functionality.

All said, I am not overly committed to the way this explainer currently describes _how _ authors supply these semantics and would be interested in hearing of alternative ways we can ensure that the behaviors focusgroup usage would provide are well defined enough that we can mitigate the accessibility concerns that blocked the original from moving forwards.

sorvell: > Why have focusgroup take (or imply) a role at all

It's currently possible to setup any of this bad behavior so I guess the concern is making it easier or seem blessed? I'd rather address those concerns more directly via limitations instead of a very specific bespoke aria application for this feature. For example, restrictions on nesting focus groups and requiring that a role be set for the group to function. It just seems like specifically baking the behavior direction into the functionality is too tight a coupling and overly restrictive.

janewman: I think there are two questions that this conversation is trying to answer:

#### 1. What is the right amount of scoping or limitation for focusgroup (if any)?
I'll agree that the net-new behavior here is having an HTML attribute take use aria as a required input, and while this certainly limits potential usage significantly from the original, un-scoped explainer, I feel like the restricted version still carries enough benefit for us to consider moving forwards with. 

One thing to consider - while trying to restrict usage in the future if an un-scoped version was implemented would be near-impossible, the reverse isn't. If down the road we can solve the issue of communicating the expected focus mode and navigation controls to screen readers for use-cases not covered by any known pattern or aria-role, I think it would be reasonable to then move to allow arbitrary `focusgroups` in a backwards compatible way, perhaps by making the role an optional parameter.

#### 2. What mechanism should control this scoping (if any)?
> requiring that a role be set for the group to function.

This was my initial approach ([see the first example in this comment](https://github.com/openui/open-ui/pull/1260#issuecomment-3211641378)), but early feedback and Aria WG discussion raised the issue that aria needs to continue to be semantic-only, and this approach broke this behavior by tying behavior to the aria role.
___
The existing, [un-scoped focusgroup explainer ](https://open-ui.org/components/focusgroup.explainer/) currently does a good job of outlining what the feature could look like without any scoping or limitations but ran into roadblocks later around these accessibility issues. This explainer attempts to use restrictions to provide functionality that screen readers could use without needing any new APIs or behavioral changes by scoping the types of scenarios that `focusgroup` can be used in to those that already have good screen reader support.

sorvell:
> This was my initial approach (https://github.com/openui/open-ui/pull/1260#issuecomment-3211641378), but early feedback and Aria WG discussion raised the issue that aria needs to continue to be semantic-only, and this approach broke this behavior by tying behavior to the aria role.

If aria is semantic only then it should not impact behavior. The proposal here tries to side-step that by using aria roles outside of aria attributes and then inferring the attributes based on those settings. Although this is slightly similar to how some native elements have inferred roles, I would still say that it is giving aria non-semantic behavior and therefore still runes afoul of this criticism.

janewman:
> @janewman how would you like to proceed? If you'd like a more thorough review & feedback on this PR, we can definitely do that, but I am in a mind to say we should merge this as-is and can continue to iterate in a more fine-grained fashion. Let me know your preference.

@keithamus I agree, and would like to merge this proposal as-is.  I'll don't see any reason why I can't address further feedback in updates.

keithamus:
Someone like @mfreed7 or @gregwhitworth will need to approve (https://github.com/openui/open-ui/blob/main/.github/CODEOWNERS#L17)

Perhaps while we're looking at this - @janewman would you like to add yourself to that line so future PRs from others can be reviewed by you.

lukewarlow:
I agree we should get this merged and address issues in follow-ups.

mfreed7 (approved):
LGTM!

mfreed7:
> Someone like @mfreed7 or @gregwhitworth will need to approve (https://github.com/openui/open-ui/blob/main/.github/CODEOWNERS#L17)
> Perhaps while we're looking at this - @janewman would you like to add yourself to that line so future PRs from others can be reviewed by you.

Done.

janewman:
@keithamus or anyone with permission, can you merge this when you get the chance?  I think it has enough approvals.

PR #1262, opened 2025-08-19
add workflow to generate telecon agendas

keithamus:
@mfreed7 @gregwhitworth 

This workflow runs every Wednesday at mightnight in order to generate a telecon agenda `.md` file, with the date set to the Thursday following (as in, the next day).

The contents of the markdown file follow the format of prior files - it has a heading with the date of the Thursday that week, followed by a bullet-point list of issues. The issues are listed in numerical order - smallest issue number first - to ensure that older issues get a chance to be handled before others. 

Given this runs on a Wednesday at midnight, the chairs can re-order them at any point before the meeting. The workflow can also be manually re-run to pick up new agenda items, at any point.

### DIFF
--- a/.github/workflows/generate-telecon-agenda.yml
+++ b/.github/workflows/generate-telecon-agenda.yml
@@ -1,0 +1,38 @@
+name: Generate Telecon Agenda
+on:
+  # We can manually dispatch this workflow to re-build the agenda at any time.
+  workflow_dispatch:
+  schedule:
+    # ... but it also automatically runs every Wednesday at midnight UTC
+    - cron: "0 0 * * 3"
+jobs:
+  build:
+    name: Generate Telecon Agenda
+    runs-on: ubuntu-latest
+    permissions:
+      # Allows this GitHub Action to push to the main repo
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - name: Generate Telecon Agenda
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-telecon-agenda.yml
```diff
@@ -0,0 +1,38 @@
+name: Generate Telecon Agenda
+on:
+  # We can manually dispatch this workflow to re-build the agenda at any time.
+  workflow_dispatch:
+  schedule:
+    # ... but it also automatically runs every Wednesday at midnight UTC
+    - cron: "0 0 * * 3"
```
mfreed7: Perhaps run it at 5 minutes after midnight, just to avoid weird race conditions with the `date` call below? It should work as-is, but putting it right at midnight makes me nervous.

mfreed7 (approved):
I assume you tested this, right? Assuming so, it looks great to me! Thanks for implementing this.

.github/workflows/generate-telecon-agenda.yml
keithamus: I picked 8:30am UTC instead which should not be midnight in any timezone

keithamus:
> I assume you tested this, right? Assuming so, it looks great to me! Thanks for implementing this.

Yeah I tested this quite a bit here: https://github.com/keithamus/test/actions/workflows/generate-telecon-agenda.yml. Tested in various conditions, e.g. this won't commit an agenda with no agenda+ issues, tested weird titles, tested re-runs, etc.

PR #1267, opened 2025-08-26
Add notes to link delegate explainer for modifier keys and pseudo-classes

### LINKED ISSUE #1254: [Link delegation] modifier key passthrough
The explainer talks about the container being augmented with the context menu when right clicked, which would be great for links. 

Can this also automatically pass through modifier keys such as Command ⌘ / CTRL, so links could open in new tab?

And is it ok that a modifier key is being pressed but the delegation goes down to a button?

### LINKED ISSUE #1255: [Link delegation] does the target get :hover, :active styles?
In playing with an example of a table with [links](https://codepen.io/afrehner/pen/empvrjM) and [buttons](https://codepen.io/afrehner/pen/raOyvrv) as targets, I noticed that it could be a little difficult to determine what action was actually going to happen when I click on the row. So I manually added `:hover` styles on the link/button to try and further indicate what the user is interacting with when they click. Perhaps the example above isn't the most clear, as there's only one link/button in each example, but it potentially would stand out more if there were multiple links/buttons.

My question is: should this be an expectation of the platform itself, that it automatically applies the `:hover`, `:active` styles on the target? Or leave that up to userland code to decide?

Or, alternatively, maybe a new pseudo class?

frehner:
Closes #1254 
Closes #1255 

Currently having issues `yarn` installing locally, so I can't get a preview of the changes working yet.

### DIFF
--- a/site/src/pages/components/link-area-delegation-explainer.mdx
+++ b/site/src/pages/components/link-area-delegation-explainer.mdx
@@ -6,8 +6,8 @@ name: Link Area Delegation (Explainer)
 layout: ../../layouts/ComponentLayout.astro
 ---
 
-- Authors and contributors: [@noamr](https://github.com/noamr), [@scottaohara](https://github.com/scottaohara), [@nmn](https://github.com/nmn), [@jakearchibald](https://github.com/jakearchibald)
-- Last updated: October 16, 2024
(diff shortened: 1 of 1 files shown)


frehner:
Well, thankfully the Netlify preview still works. Appears the formatting is ok

keithamus (approved):

PR #1269, opened 2025-09-01
Remove stale bot

keithamus:
It doesn't seem like something we're really embracing. We have a lot of open issues, I think we'd be better off doing some semi-regular triage. 

See discussion here, also: https://discord.com/channels/714891843556606052/714891844013785190/1407435550385766571


gregwhitworth:
I agree on the regular triage session; ironically the tagging is meant to help with that since it sets the label. Possibly the label itself is sending the wrong message and what we really want is a "needs-triage" style label and no message added unless we can use placeholders (eg: {issue owner} what are the next steps for this issue?). Then we can leverage that label for top pri triaging? That said, a monthly 1 hour triaging session (not the same as the Thursday session) would be useful. Could be open to everyone but chairs are required?

keithamus:
@gregwhitworth I've added a `needs-triage` workflow instead. It's essentially the same workflow but with softer wording 😉

PR #1284, opened 2025-09-17
Add an initial explainer for `<toolbar>`

lukewarlow:


site/src/pages/components/toolbar.explainer.mdx
```diff
@@ -0,0 +1,141 @@
...
+### How can users customize the UI of this toolbar element with CSS?
+
+We propose this element should not support the CSS appearance property and just have a "base" appearance by default like the `<dialog>` element.
+
+**Questions**
+
+* `<toolbar>` should display children horizontally, but support using CSS writing-mode.
+    * Can have vertical case by:
+        * Set CSS `writing-mode: vertical-lr` on the `<toolbar>`
+        * Should this affect arrow-key directionality?
```
smhigley, line 74: Both arrow key directionality and `aria-orientation` would be my initial thought. Having CSS affect semantics can be a bit tricky, but I think other people might have a better handle on how feasible this specific case is (vs. an html attribute)

site/src/pages/components/toolbar.explainer.mdx
```diff
@@ -0,0 +1,141 @@
...
+* Does `<toolbar>` support separators?
+    * If so, we should support `<hr>` for this.
+* Toolbars generally should be labelled, it is a MUST if there are multiple on a page. Should we support this via HTML,
+  or still rely on aria?
+    * Could use `<legend>` to do this?
+
```
smhigley, line 81: I'd like to throw in another general question on how to handle keyboard behavior when authors throw in weird / complex children (e.g. a select that already has keyboard behavior, or nested toolbars, etc. etc.)

keithamus (approved):

site/src/pages/components/toolbar.explainer.mdx
lukewarlow: Have added that as a seperate bullet to the list.

site/src/pages/components/toolbar.explainer.mdx
lukewarlow: Added that in and linked to the focusgroup relevant issue.

lukewarlow:
@smhigley I'll let you double check the updates and then get this merged

Issue #1281, opened 2025-09-15
[focusgroup] Default behaviour for key conflict elements


gfellerph:
The focusgroup explainer [interactive content inside focusgroups](https://open-ui.org/components/scoped-focusgroup.explainer/#interactive-content-inside-focusgroups) section describes that some elements like `<input type="number">` or `<input type="number">` already have actions assigned to arrow keys that conflict with the focusgroup navigation. There is a list of potential solutions to this issue in the explainer.

> Fortunately, there are several solutions to this problem:
> - Remove tabindex=-1 from the “Increase/Decrease Font” buttons.
> - Move the “Increase/Decrease Font” buttons before the combobox. (Refer to “Avoid including controls whose operation requires the pair of arrow keys used for toolbar navigation” in the [Toolbar control pattern](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/).) Additionally, [opt-out](https://open-ui.org/components/scoped-focusgroup.explainer/#opting-out) the <input> control from focusgroup participation so that arrow keys skip it. Alternatively, [turn off the focusgroup’s memory](https://open-ui.org/components/scoped-focusgroup.explainer/#disabling-focusgroup-memory) so that focus isn’t automatically returned to the combobox.
> - Use script to intercept focusgroup-related keydown events on the <input> and move focus manually. Also consider [limiting the focusgroup](https://open-ui.org/components/scoped-focusgroup.explainer/#limiting-linear-focusgroup-directionality) to one axis and reserving the other axis for operating the `<input>`.

I'd like to propose another approach that would provide a default behaviour that can be enhanced by authors.

### Turning off roving behaviour when a key conflict element gets focused

It's best explained with a demo: https://gfellerph.github.io/focusgroup-polyfill/. The first example "regular focusgroup" in this experimental implementation has an input field amongst other focusable elements inside the focusgroup. The memory functionality is on. However, when the input field is focused by arrow key navigation, the memory functionality on all other focusgroup candidates is turned off. Tab or shift tab leaves the input field and focuses the next focusable element in either direction. If this is a focusgroup candidate and not a key conflict element, memory behaviour is restored. Tabbing again will leave the focusgroup. The key conflicting `<input>` can never be the memory element. To change the behaviour, the following options are (still) available:

- Skipping over the input when navigating with arrow keys can be achieved by setting `<input focusgroup="none">`
- Disabling the arrow keys on input fields and other elements that already use arrow keys for other actions is left to the author
[...]

Issue #1008, opened 2024-02-27
[focusgroup] clarify and handle cases where scrollable regions have focusgroups (accessiblity)


travisleithead:
Mentioned first in issue #990.

In some cases, a focusgroup's capability of scrolling the document to the next focusable item in a focusgroup can be in conflict with a user's need to use the arrow keys to manually scroll a scrollable container in order to read all content.

See demo by @scottaohara: https://codepen.io/scottohara/pen/abMaxNx

In such situations, focusgroup needs to give priority to scrollable-area scrolling.

One proposed solution is to automatically disable focusgroup arrow-key navigation to the next focusable item when that next focusable item is not "in view". In which case, once arrow-key scrolling has moved the item into view, then the focusgroup arrow-key navigation will move the focus to that item.
* Pros: resolves the issue, can be automatic.
* Cons: makes focusgroup arrow key navigation a bit less predictable if the arrow key will "magically" switch between content scrolling and focusgroup navigation. (May need to experiment with this in the case of a listbox-scenario to see how unexpected it is...)

travisleithead:
The `wrap` capability may also be impacted by this...

From @gfellerph 

> In Scotts example was a small scroll area filled with long content accordion items. The titles are focusgroup candidates and while the first title is focused, the content is so long, that the second title is out of view. The issue describes that the focus should not jump to the second title when the arrow down key is pressed, but should rather scroll the container until the second title is in view and then focus it. The first title is now out of view. If wrapping behavior is applied to the focusgroup, an arrow down keypress should jump back to the first title, but it's out of view so the focusgroup does not capture the event and the page/container scrolls further down (or does nothing if the end of the page is reached).
> 
> This could just be handled with an exception to that rule, but I don't really know what's expected from a user perspective vs. what's defined in markup.

travisleithead:
Thinking about @scottaohara's idea of a modifier key, and if we can't make a default solution "just work" (I'm not saying we can't, but just generating some other possibilities...) I wonder if `Shift`+Arrow keys could be a built-in fallback to enable the UA to scroll the scrollable area when focus is on a focusgroup element inside of a scrollable container. "Shift" + arrow keys is typically reserved for patterns in which selection should be extended, so maybe not a good idea if it gets in the way of some actual JS selection logic... Another approach is a built-in on/off toggle mechanism. A bad example: the never-used ScrollLock key on some physical keyboards that could be used to toggle between arrow key navigation and arrow key scrolling behavior...?

[...]

PR #1289, opened 2025-09-25
[focusgroup] Add sections on scrolling and keyboard conflicts

janewman:
Clarifies key conflict element behavior and adds documentation for scrolling interactions in the scoped focusgroup explainer. 
This addresses how focusgroups handle interactive elements that consume arrow keys (like text inputs) and introduces an automatic tab-escape behavior for native elements with keyboard conflicts

Related issues:
* Key conflict element behaviors: https://github.com/openui/open-ui/issues/1281
* Scrolling interactions: https://github.com/whatwg/html/issues/11641#issuecomment-3310120042

### DIFF
--- a/site/src/pages/components/scoped-focusgroup.explainer.mdx
+++ b/site/src/pages/components/scoped-focusgroup.explainer.mdx
@@ -714,6 +714,8 @@ In this example:
 - Focus is automatically scrolled into view when navigating between items that are off-screen
 - Authors should be aware that since focusgroup navigation takes priority over scrolling, they should take care when constructing large focusgroups to avoid a situation where content can be missed when jumping from one item to another.
 
+**Accessibility considerations:** When focusgroup items are separated by large amounts of content, arrow key navigation can skip over intermediate content that users might need to read. See the [open question about scrolling behavior](#open-questions) for potential solutions being considered.
+
 #### 2. Scrollable region within a focusgroup
 
 When a scrollable element is contained within a focusgroup, the behavior depends on whether there's a conflict between the focusgroup's arrow key handling and the scrollable element's needs.
@@ -1232,7 +1234,9 @@ approaches were considered.
   - (C) Split into two attrs: `pattern="tablist" focusgroup="wrap"` (clearer separation, extra verbosity & API surface).
   - (D) Native elements only (e.g., future `<tabs>`, `<toolbar>`, `<menubar>`); attribute becomes redundant—risk: slower coverage, custom element ecosystems still need declarative navigation.
   - Criteria to decide: author error rate, implementation complexity, consistency with existing HTML token patterns, incremental ship path.
-8. **Keep or drop child role inference (or defer as future consideration):** Should v1 exclude automatic child role assignment entirely to reduce complexity
+6. **Scrolling behavior when focus target is not in view:** Should focusgroup navigation automatically prioritize scrolling over focus movement when the next focusable item is not currently visible? This addresses accessibility concerns where arrow key navigation can skip over intermediate content that users need to read (see [GitHub issue #1008](https://github.com/openui/open-ui/issues/1008)). Potential solution:
+  - Temporarily disable focusgroup navigation when the target item is out of view, allowing normal scrolling until the item becomes visible.
+7. **Keep or drop child role inference (or defer as future consideration):** Should v1 exclude automatic child role assignment entirely to reduce complexity
       and perceived overreach (keeping only container pattern + navigation)? Rationale for revisiting: reviewer concern about mixing semantics & behavior;
       authors can still supply explicit roles; deferring would let us ship navigation sooner and gather data on real author pain before standardizing inference.
 


site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -581,48 +611,102 @@ sure users can "escape" these elements. Built-in elements provide this via the t
...
+  <!-- More content that makes this scrollable -->
+</div>
+```
+
+In this example:
+- Left/right arrows navigate between toolbar buttons and wrap around
+- Up/down arrows scroll the container since the focusgroup is limited to `inline`
+- Focus is scrolled into view automatically when navigating between buttons
+
+Note: Focusgroup is primarily intended for grouping related controls, so it is not recommended to use focusgroup on large scrollable lists or grids of items. Instead, use focusgroup on a toolbar or menu that controls the larger scrollable content. The above example is provided to illustrate the interaction between focusgroup and scrolling, and is **not** a recommended use of focusgroup.
```
scottaohara: this note seems to rule out focusgroup eventually being used to create a grid pattern - or some might read it as saying that a scrollable list of options, for a custom combobox, would somehow be wrong?

i understand this may just be current scoping to limit the initial feature.  so this may need nothing more than some minor adjustments to allude to the fact this is the _current_ plan.  Which would help align it to the section "(Future Consideration) Grid focusgroups", that this note presently says it is not intended for.

janewman: Great point! 
I've updated the example to instead show what I would consider a good usage of focusgroup, and have removed/replaced this larger note with a short point about taking care to ensure that jumping around the page doesn't cause users to miss content.

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -608,48 +611,103 @@ sure users can "escape" these elements. Built-in elements provide this via the t
...
+When there is a conflict between the arrow keys consumed by the interactive element and the focusgroup's navigation, focusgroup will **not** interfere with the interactive element's behavior. This means that the normal way to move focus between focusgroup items (arrow keys) will **not** work when focus is within such an element.
+
+Examples:
+- `<input>` elements (most, but not all types use arrow keys)
+- `<textarea>` elements
+- `<select>` elements when the arrow-key axis used by the focusgroup is the same axis used by the select
+- Elements with `contenteditable`
+- Focusable scrollable regions, when the scroll direction is in the same axis as the focusgroup
+- Custom elements with arrow key handlers
+- Elements with `preventDefault()` on arrow keys
```
gfellerph, line 625: I know it's just a list of examples, but maybe it would be good to have a complete list in the future.

```suggestion
- Elements with `preventDefault()` on arrow keys
- Audio and video elements with visible controls
- Iframes and object tags with focusable elements inside
```

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -608,48 +611,103 @@ sure users can "escape" these elements. Built-in elements provide this via the t
...
-   [turn off the `focusgroup`'s memory](#disabling-focusgroup-memory) so that focus isn't
-   automatically returned to the `combobox`.
-- Use script to intercept `focusgroup`-related keydown events on the `<input>` and move focus
-   manually. Also consider [limiting the `focusgroup`](#limiting-linear-focusgroup-directionality) to
-   one axis and reserving the other axis for operating the `<input>`.
+In this example:
+- The input field **is** a focusgroup item reachable via arrow keys.
+- Arrow keys are consumed for text cursor navigation once focus is within the input.
+- To ensure the user can still reach the other elements in the focusgroup, pressing Tab moves focus to the "Save" button, and Shift+Tab moves focus to the "Italic" button.
+
```
gfellerph, line 681: If the "Go" button is not reachable by tab navigation, is it reachable by arrow key navigation in this example? It might be good to list an additional point here to explain what happens to this button in each case or why it might have been taken out of tab order.

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -608,48 +611,103 @@ sure users can "escape" these elements. Built-in elements provide this via the t
...
+    <div tabindex="-1" aria-selected="false">Item 100</div>
+  </div>
+</div>
+```
+
+In this example:
+- Up/down arrows navigate between list items in the focusgroup
+- Left/right arrows scroll the container horizontally (if needed) since the focusgroup is limited to `block`
+- Focus is automatically scrolled into view when navigating between items that are off-screen
+- Authors should be aware that since focusgroup navigation takes priority over scrolling, they should take care when constructing large focusgroups to avoid a situation where content can be missed when jumping from one item to another.
```
gfellerph, line 715: The case demoed in [@scottaohara s codepen](https://codepen.io/scottohara/pen/abMaxNx), where there is so much content between the focusgroup candidates that sibling candidates are out of view and scrollable content in between is skipped when focus jumps to the next candidate, is not handled here. There is still an open issue for the original focusgroup explainer (https://github.com/openui/open-ui/issues/1008). Should this issue be mentioned here?

janewman: Added and linked to a new "Open Question" around if we should have the behavior of scrolling when the next focusgroup item is not yet in view, rather than jumping.

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: I answered this question in the latest update, but answering here as well: "Go" is still in the focusgroup, this example shows that authors have control over the next focused item when exiting the key conflict element via setting tabindex, just like they control focusgroup entry in general.

Let me know what you think here!

gfellerph: I think that would be the expected behaviour.

gfellerph (approved):

sorvell:
[Here's a little example](https://codepen.io/Steve-Orvell/pen/ogbYgbZ?editors=1000) to help evaluate related behavior of native elements.

Chrome and Firefox activate and scroll to the focused/selected radio when arrow keys are used, Safari activates it but does not scroll to it (but it does scroll to the item focused by the tab key, seems sort of inconsistent).

Segmenting the radio group with focusable elements works mostly as proposed in the explainer but the segments without an active radio cannot be focused with tab, effectively changing the focus order when the selected radio changes.

janewman:
@mfreed7 can you PTAL when you get the chance? I need a +1 from another focusgroup owner.

janewman:
@mfreed7 (or really anyone with permission) can you click the "merge" button? I still don't have Github permissions...

@gregwhitworth can you give me permissions to be able to do this myself? This and the ability to add labels to issues would be really helpful!!

PR #1290, opened 2025-09-29
[focusgroup] Clarify tab behavior when focusgroup is bisected by an opted-out item

janewman:
The current explainer didn't have much to say about how opted out items impact sequential focus navigation, this update remedies that by describing the behavior.

This ensures that there will never be tab loops and that focusgroups won't cause unintuitive jumping when moving with tab/shift+tab

### DIFF
--- a/site/src/pages/components/scoped-focusgroup.explainer.mdx
+++ b/site/src/pages/components/scoped-focusgroup.explainer.mdx
@@ -37,6 +37,8 @@ layout: ../../layouts/ComponentLayout.astro
 - [Enabling wrapping behaviors](#enabling-wrapping-behaviors)
 - [Limiting linear focusgroup directionality](#limiting-linear-focusgroup-directionality)
 - [Opting-out](#opting-out)
+  - [Impact on sequential focus navigation](#impact-on-sequential-focus-navigation)
+  - [Nested Focusgroups](#nested-focusgroups)
 - [Disabling focusgroup memory](#disabling-focusgroup-memory)
 - [Adjustments to sequential focus navigation](#adjustments-to-sequential-focus-navigation)
 - [(Future Consideration) Grid focusgroups](#future-consideration-grid-focusgroups)
@@ -788,10 +790,11 @@ To opt-out:
 |---|---|
 | `focusgroup="none"` | Opt-out: this element and its shadow-inclusive descendants are not considered `focusgroup` candidates. |
 
-In the following example of a toolbar, a help section opts-out of `focusgroup` behavior so that any
-interactive content inside it is bypassed when arrowing among the primary formatting controls.
+### Impact on sequential focus navigation
 
-Example:
+When elements opt-out of a `focusgroup` using `focusgroup="none"`, they can effectively divide the focusgroup for sequential focus navigation purposes, creating multiple tab stops where the focusgroup would normally have only one. This happens because opted-out elements remain in the normal sequential focus navigation order while splitting the focusgroup into separate segments.
+
+In the following example, a help section opts-out of `focusgroup` behavior so that any interactive content inside it is bypassed when arrowing among the primary formatting controls, but remains reachable via tabbing.
 
 ```html
 <div focusgroup="toolbar inline wrap" aria-label="Text formatting">
@@ -804,6 +807,24 @@ Example:
   <button type="button" tabindex="-1">Underline</button>
 </div>
 ```
+
+**Sequential focus navigation behavior:**
+
+1. **Entering from before:** Pressing Tab to enter this focusgroup from a preceding element will focus "Bold" (the first `focusgroup` item).
+2. **Tab from "Bold":** Pressing Tab moves focus to "Help" (the first opted-out element in sequential order).
+3. **Tab from "Help":** Pressing Tab moves focus to "Shortcuts" (following normal sequential focus navigation within the opted-out subtree).
+4. **Tab from "Shortcuts":** Pressing Tab will **re-enter the focusgroup**, but will only consider `focusgroup` items that follow "Shortcuts" in DOM order. In this case, focus moves to "Underline".
+5. **Shift+Tab from "Help":** Pressing Shift+Tab will move focus back to the focusgroup, but will only consider `focusgroup` items that precede "Help" in DOM order ("Bold" or "Italic"). The specific item chosen follows the standard [sequential focus navigation rules](#adjustments-to-sequential-focus-navigation), applied only to the focusgroup segment before "Help".
+
+**Key behaviors:**
+
+- **Multiple tab stops:** The focusgroup effectively has two sequential focus navigation entry points: one before the opted-out section and one after.
+- **Memory limitations:** When re-entering a focusgroup segment, focusgroup memory only applies to `focusgroup` items within that segment. Memory of items in other segments is not considered.
+- **Direction-based scope:** The direction of sequential focus navigation (Tab vs Shift+Tab) determines which focusgroup segment is considered for entry.
+
+**Arrow key navigation is unaffected:** Arrow keys will skip over the opted-out elements entirely. Pressing the right arrow on "Italic" will move focus directly to "Underline", treating the opted-out span as if it doesn't exist for arrow navigation purposes.
+
+### Nested Focusgroups
 
 When a `focusgroup` definition is applied to an element, it implicitly opts out of any ancestor's
 `focusgroup`s. This ensures that every element can only belong in one `focusgroup` at a time.
@@ -872,6 +893,10 @@ change:
 To ensure that a `focusgroup` always has at least one tab stop in the sequential focus navigation
 order, and to provide the appropriate "hook" for a `focusgroup`'s "memory" to redirect focus to the
 last-focused element in a `focusgroup`, a change to sequential focus navigation is needed.
+
+Note that when elements [opt-out of focusgroup participation](#impact-on-sequential-focus-navigation),
+they can create multiple entry points for sequential focus navigation within what would otherwise be
+a single focusgroup.
 
 This change is intended to ensure that focus is directed to one of the following `focusgroup`
 candidates whenever focus enters a `focusgroup`. The first matching condition is always taken:


site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -501,9 +505,22 @@ down/left moves focus backwards assuming the `<p>` element has `writing-mode` `h
...
+### Focusgroup segments
+
+A **focusgroup segment** is a contiguous group of focusgroup items that can be navigated without crossing any [opted-out elements](#opting-out). When elements within a focusgroup use `focusgroup="none"` to opt-out, they effectively divide the focusgroup into separate segments for sequential focus navigation purposes.
+
+Each segment operates independently for certain behaviors:
+- **Memory:** Each segment maintains its own last-focused element memory
+- **Guaranteed tab stop:** Each segment ensures at least one tab-accessible element
+- **Sequential navigation:** Tab and Shift+Tab treat each segment as a separate focusgroup entry point
+
+For example, if a focusgroup contains items A, B, C, D, E and item C opts-out, this creates two segments: [A, B] and [D, E]. Arrow key navigation can move freely within each segment and skip over the opted-out element C, but sequential focus navigation (Tab/Shift+Tab) will treat these as two separate tab stops.
```
gfellerph: Is this also true if the opted out element does not contain any focusable elements (which would be weird, but they might not be rendered yet in a conditional form with react or whatever)?

Because if there are no focusable items within the opted out element, there seems to be no reason to segment and add an apparently random tabstop.

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -805,6 +823,31 @@ Example:
...
+
+4. **Tab from "Shortcuts":** Pressing Tab will **re-enter the focusgroup**. The focusgroup segment now includes only items that follow "Shortcuts" without crossing opted-out elements ("Underline"). Since there's no memory for this segment and "Underline" has `tabindex="-1"`, the guaranteed tab stop algorithm focuses "Underline".
+
+5. **Shift+Tab from "Help":** Pressing Shift+Tab will move focus back to the focusgroup segment that precedes "Help" in tree order ("Bold" and "Italic"). The guaranteed tab stop algorithm is applied only to this segment, considering any memory and sequentially focusable items within this segment only.
+
+**Key behaviors:**
+
+- **Focusgroup segments:** As defined above, opted-out elements create boundaries that divide the focusgroup.
+- **Guaranteed tab stop per segment:** Each segment follows the [guaranteed tab stop algorithm](#guaranteed-tab-stop-algorithm).
+- **Memory scope:** Focusgroup memory only applies within the segment being entered.
```
gfellerph: @janewman you showed an example at some point that has a search box in the middle of a toolbar. If that example also had a focusgroup none after the searchbox, segmenting the group. If a user enters a search term, focus might jump to a results list beyond the focusgroup. When the user then shift-tabs back to the toolbar, the memory element or guaranteed tab stop of the second segment will be focused and not the search bar in the first segment. To me, this seems to break with the premise of memory that should allow users to jump back in where they left off. Should memory not apply to the entire focusgroup, thus not respecting segmentation?

I guess this will rarely be an issue though since when tabbing through the page, at least one element of the last segment would receive focus and become the memory element and therefore it also makes sense to re-enter there when shift-tabbing back. It would only be noticeable if focus leaves the focusgroup through some other means than keyboard navigation.

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: I'll clarify this, and I'm going to add an additional requirement that they are in tab order (not tabindex=-1) before the focusgroup would be considered split. I agree that if there isn't a tabstop in the middle, there is no reason to segment.

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: My thinking is that if the author went out of their way to make an item to both not participate in the focusgroup and wanted it to appear in tab order, we should respect their decision and never "jump" past it.

In this example, let's assume that the pressed save and then focus was moved to somewhere below the toolbar, leaving the save button as the last focused item.
```html
<div focusgroup="toolbar inline wrap" aria-label="Text formatting">
  <button type="button">Bold</button>
  <button type="button">Save</button>
  <button type="button">Italic</button>
  <span focusgroup="none">
    <button type="button">Super Important</button>
  </span>
  <button type="button">Underline</button>
  <button type="button">Strikethrough</button>
  <button type="button">link</button>
</div>
```
I see a few different ways this situation can be handled:
## 1. memory item is one-per-focusgroup, memory takes precedence over opted out element in taborder precedence.
Here, shift tabbing would restore focus immediately to the Save button, skipping over the "Super Important" button. Tabbing again moves focus to the Super Important button. Tabbing again moves focus to:
a. The content after the toolbar, as we are "after" the focusgroup. This preserves the single tabstop concept.
b. "Underline" which now replaces the "Save" button as the last focused item. This introduces a new rule about what items in the focusgroup are allowed to be considered when entering from an opted-out element.
c. The "Save" button, as we want to restore memory (this creates a loop and is clearly not an option in my opinion)
## 2. memory item is one-per-focusgroup, memory does not take precedence over opted out items in taborder precedence.
Here, shift tabbing would move focus to the "Super Important" button. Shift tabbing again moves focus to the "Save" button. Tabbing from here moves focus back to the "Super Important" button, and tabbing again moves focus to:
a. The content after the toolbar, as we are "after" the focusgroup. This preserves the single tabstop concept.
b. "Underline" which now replaces the "Save" button box as the last focused item. This introduces a new rule about what items in the focusgroup are allowed to be considered when entering from an opted-out element.
c. The "Save" button, as we want to restore memory. (this creates a loop and is clearly not an option in my opinion)
## 3. memory item is one-per-focusgroup _segement_, memory does not take precedence over opted out element in taborder precedence.
Here, shift tabbing would move focus to the "link" button (assuming there is no memory item). Shift tabbing again moves focus to the "Super Important" button. Shift tabbing again moves focus back to the "Save" button, Tabbing again moves focus to the "link" button. But then what happens to memory if the opt-out is dynamically removed and the two segments are joined? 
a. Memory is cleared.
b. Both are kept until focus re-enters the group, prioritizing the closest to the element focus is moving from.

__________________________

I never like the feeling of "skipping" content that 1-a and 2-a brings, that is my main motivation for #3
Let me know if you have strong feelings towards one of these or have an idea that wasn't mentioned.

Edit: updated example to use a "Save" button to avoid having to also mention the key-conflict behavior here and overcomplicating things.

janewman: I spent some time thinking about this one, and I am coming back around to a single memory item per focusgroup, not per segment. I want to retain the idea of a focusgroup segment to facilitate the entry algorithm. This means that if the memory element is not is the segment being navigated to, memory won't end up being used. 
See this same example, assume that the user pressed save and then focus was moved to somewhere below the toolbar, leaving the save button as the last focused item.
```html
<div focusgroup="toolbar inline wrap" aria-label="Text formatting">
  <button type="button">Bold</button>
  <button type="button">Save</button>
  <button type="button">Italic</button>
  <span focusgroup="none">
    <button type="button">Super Important</button>
  </span>
  <button type="button">Underline</button>
  <button type="button">Strikethrough</button>
  <button type="button">Link</button>
</div>
```
1. Shift+Tab from after the toolbar would move focus to the "link" button. "Link" is now the memory item.
2. Shift+Tab again moves focus to the "Super Important" button. 
3. Shift+Tab again moves focus to the "Italic" button, as memory was set to "link" in the first step. "Italic" is now the memory item.

If an author wanted to make Save be the element that a user would jump to when shift-tabbing from after the toolbar, they have a few options:
* Reconsider making "Super Important" an opted-out item.
* Reconsider making "Super Important" In the middle of the focusgroup, either put it outside or on at one of the ends.
* Rather than having a focusable opt-out element splitting the focusgroup, consider making two focusgroups, each with their own memory. This would have the downside of not being able to arrow between these groups.

As always, let me know your thoughts here; more than happy to pivot if there is another approach that makes more sense.

gfellerph: Thanks for the update and inputs, got me thinking too. What I had in mind initially was: 

1. User navigates to "Save", that's now the memory
2. Save moves focus beyond the toolbar
3. Shift+Tab moves back to the toolbar, but because the three buttons below "Super Important" are part of the focusgroup and not the memory, focus would go to "Super Important" first, following the DOM order
4. Shift + Tab would jump to "Save" since that's the memory element

What I did not consider enough with this approach is what happens when step 4 would be a Tab instead of Shift + Tab. Then I'd expect focus to move to "Underline", handling the `focusgroup=none` similar to a key conflict element. But I realised that this is inconsistent and I also never was quite sure if it would be a good idea to skip the three buttons in step 3. Therefore I think the last described approach is the sensible way to go, especially with the alternative options that bring a lot of flexibility to handle more complex UIs.

janewman:
@mfreed7 can you PTAL when you get the chance? I need a +1 from another focusgroup owner.

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: That is what is in the latest, thanks for helping get this to a better spot!

travisleithead (approved):
Love this approach, which nicely handles the use cases.

janewman:
@keithamus, or @gregwhitworth (or anyone else) can we merge this? I am still waiting on getting permissions to click that button.

mfreed7:
> @keithamus, or @gregwhitworth (or anyone else) can we merge this? I am still waiting on getting permissions to click that button.

Sorry you have to keep asking this. I wish I understood the Github permissions model better.

janewman:
> > @keithamus, or @gregwhitworth (or anyone else) can we merge this? I am still waiting on getting permissions to click that button.
> 
> Sorry you have to keep asking this. I wish I understood the Github permissions model better.

I appreciate you clicking this button for me!

Issue #1295, opened 2025-10-03
[focusgroup] Should we remove the "future considerations" out of the new explainer?


janewman:
Specifically, the grid section takes up a lot of the existing explainer, and is not something we are ready to go after for the first version of the API, should it be removed from [scoped Focusgroup (Explainer) | Open UI](https://open-ui.org/components/scoped-focusgroup.explainer/)?
We could keep it as a mention, similar to how this doc treats CSS support.

lukewarlow:
Maybe move to a separate document that's linked to? No need to be in the sidebar.

Alternatively put it in a details so it's collapsed?

janewman:
grid section was collapsed, closing this

Issue #1277, opened 2025-09-04
Should "Inactive" be more obvious?


bkardell:
If you look at any of the inactive proposals on the site, it isn't super obvious that they are inactive or legacy or that we gave up on them or something in favor of something new (which we'll be doing with the new focus group)... or why.  It might be useful as it is with specs to have some warning or different styling drawing a little bit of attention to that fact.

lukewarlow:
In the case of Focusgroup we should maybe even use the old url for the new one and have the old document completely hidden (except via a link from the new one)

PR #1298, opened 2025-10-07
[focusgroup] General clarifications and edits, move previous explainer out of the sidebar

janewman:
Major Changes in this PR:
* Aligns document around the idea that each focusgroup segment will have a **single, guaranteed tab stop**, even if the element would normally be focusable via sequential tab navigation, e.g. natively focusable or a tabindex >= 0. This means:
  * Authors don't have to manage tabindex outside of ensuring their items are focusable and managing first-entry order.
  * Users can **not** tab between focusgroup items in the same segment, even if they have a positive tabindex and would otherwise expect to be in sequential tab order. 
* Renames this page to "Focusgroup" and removes the other version of the explainer from the sidebar (still linkable from this document), see: https://github.com/openui/open-ui/issues/1277#issuecomment-3255219831.
* Removes grid examples outside the grid section, and makes the contents of the grid section collapsed. https://github.com/openui/open-ui/issues/1295
* Misc. Adds missing or broken links, allows table of contents to be hidden, misc editorial.

### DIFF
--- a/site/src/pages/components/scoped-focusgroup.explainer.mdx
+++ b/site/src/pages/components/scoped-focusgroup.explainer.mdx
@@ -527,7 +527,7 @@ Example:
 ```html
 <div focusgroup="toolbar" aria-label="Text Formatting" aria-controls="…">
   <div>
-    <button type="button" aria-pressed="false" value="bold" ><span>Bold</span></button>
+    <button type="button" aria-pressed="false" value="bold"><span>Bold</span></button>
     <button type="button" aria-pressed="false" value="italic"><span>Italic</span></button>
     <button type="button" aria-pressed="false" value="underline"><span>Underline</span></button>
   </div>


site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -138,11 +150,11 @@ Opting an element and its subtree out of an ancestor focusgroup, see: [Opting-ou
...
-| `inline` / `block` | Optional | Restrict arrow navigation to that logical axis. Omit to allow both axes. |
-| `wrap` | Optional | Loops end→start and start→end. |
-| `no-memory` | Optional | Disable restoring last-focused item on re-entry (memory is on by default). |
-| `none` | Standalone | Opt-out: exclude element and its subtree from ancestor focusgroup. Cannot be combined with any other token. |
+| `<behavior>` | Yes (unless using [`none`](#opting-out)) | Token describing the intended behavior. See: [Supported Behaviors](#supported-behaviors). |
+| [`inline`](#limiting-linear-focusgroup-directionality) / [`block`](#limiting-linear-focusgroup-directionality) | Optional | Restrict arrow navigation to that logical axis. Omit to allow both axes. |
+| [`wrap`](#enabling-wrapping-behaviors) | Optional | Loops end→start and start→end. |
+| [`no-memory`](#disabling-focusgroup-memory) | Optional | Disable restoring last-focused item on re-entry (memory is on by default). |
+| [`none`](#opting-out) | Standalone | Opt-out: exclude element and its subtree from ancestor focusgroup. Cannot be combined with any other token. |
```
gfellerph, line 157: Should there be a token for the case someone wants the arrow key navigation but not the "only one tabstop for the whole group" behaviour?

Something like `no-roving` (naming TBD) could allow users to navigate by arrow keys within the focusgroup but not touch the original tab order. Every focusable element within the focusgroup would still be their own tabstop. This could also be something to add later to keep scope creep at bay, but I'd be interested in feedback.

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -541,25 +513,23 @@ The `focusgroup`'s memory is cleared if the last-focused item becomes non-focusa
...
 
 Example:
 
 ```html
 <div focusgroup="toolbar" aria-label="Text Formatting" aria-controls="…">
   <div>
-    <button type="button" aria-pressed="false" value="bold" tabindex="-1"><span>Bold</span></button>
-    <button type="button" aria-pressed="false" value="italic" tabindex="-1"><span>Italic</span></button>
-    <button type="button" aria-pressed="false" value="underline" tabindex="-1"><span>Underline</span></button>
+    <button type="button" aria-pressed="false" value="bold" ><span>Bold</span></button>
```
gfellerph:
```suggestion
    <button type="button" aria-pressed="false" value="bold"><span>Bold</span></button>
```

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: I created https://github.com/openui/open-ui/issues/1302 to track this request.

>This could also be something to add later to keep scope creep at bay, but I'd be interested in feedback.

Personally, I agree, this seems like something that would be easy to add as an addition, but as always, I can be flexible. If people feel like this is what would stop them from being able to use the feature, this is a fairly simple addition.

I said this in the issue (feel free to add your own interpretation if it doesn't match) but my vision would be that this token would simply skip any adjustments focusgroup would normally make to sequential focus navigation, e.g. no memory, no key conflict escape, etc. If a user only used tab, it would be as if focusgroup was not applied, but they could still move focus between items with the arrow keys.

janewman:
@mfreed7 or @gregwhitworth (or any other owner) PTAL and let me know what changes are needed, or please hit "submit" if everything looks good.

Issue #1273, opened 2025-09-03
Some thoughts and questions about combobox


josepharhar:
I have some thoughts about how to move forward with combobox and some questions about the explainer: https://open-ui.org/components/combobox.explainer/

### New element

Due to feedback from Apple, I don't think that we should try to make a new `<combobox>` element:
- https://github.com/whatwg/html/issues/9799#issuecomment-1770254871
- https://github.com/whatwg/html/issues/11288#issuecomment-2857464795

Instead, I think that we could try to just reuse `<input type=text list=datalist>` and put base appearance on it (and its datalist), like we did with customizable select.

### filter attribute

Without the proposed filter attribute, are all the options in the datalist supposed to be rendered all the time? Why not just always let filtering happen without the need for another attribute?

### button and selectedcontent elements

The explainer has an example that has a button and selectedcontent element:

```html
<combobox>
  <input type="text"/>
  <button><selectedoption></selectedoption></button>
  <datalist>
    ...
```
I'm guessing that this is supposed to clone the DOM contents of the currently selected option into something inside the text input, in addition to the text that the user is actually typing? I think that the select element with some filtering upgrades could work better for this.

### filtering for select elements

[This HTML issue](https://github.com/whatwg/html/issues/11288) made me think more about how we could try to add a text filter to the select element in some way, which I think would accomplish a lot of use cases related to the combobox proposal. I posted more about that here: https://github.com/openui/open-ui/issues/847#issuecomment-2876984892

I think there could be some overlap with the proposed `search` attribute and how we could make filtering work for select elements like @lukewarlow mentioned here: https://github.com/openui/open-ui/issues/847#issuecomment-2877017007

josepharhar:
Proposed resolutions:
1. Use input and datalist elements instead of creating a new combobox element.
2. Replace the filter attribute with a "manual" value for the search attribute which can disable automatic filtering.

I'd also like to better understand the button and selectedcontent elements usage in the combobox explainer, perhaps that can be in a separate issue.

josepharhar:
I started a PR for a separate combobox explainer: https://github.com/openui/open-ui/pull/1313

PR #1313, opened 2025-10-29
Create alternate combobox explainer

josepharhar:
This is for the resolution here:
https://github.com/openui/open-ui/issues/1273#issuecomment-3412722822

### DIFF
--- a/site/src/pages/components/combobox-old.explainer.mdx
+++ b/site/src/pages/components/combobox-old.explainer.mdx
@@ -8,6 +8,8 @@ layout: ../../layouts/ComponentLayout.astro
 import ComboboxAnatomy from '../../components/combobox-anatomy'
 import Image from '../../components/image.astro'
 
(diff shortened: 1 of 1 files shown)


lukewarlow:
I think this should be in the main file and the old one should be re-urld?

josepharhar:
> I think this should be in the main file and the old one should be re-urld?

I agree, but I want @gregwhitworth to be OK with it first

josepharhar:
I merged the filterable select explainer into this and filled out the filtering APIs section

josepharhar:
I overwrote the existing combobox explainer with the latest commit. @gregwhitworth want to take a look?

josepharhar:
> I think this should be in the main file and the old one should be re-urld?

I re-added the original explainer with and put "old" in the title. Can anyone merge this? @gregwhitworth @keithamus @lukewarlow @mfreed7

josepharhar:
Thanks @lukewarlow!

Issue #847, opened 2023-09-26
How to make an accessible searchable "select" list?


josepharhar:
I was reading this article on gregs website: https://www.gwhitworth.com/posts/2019/can-we-please-style-select/
This article points out why we need `<selectlist>`, but also says that the top frustration with `<select>` is "Not being able to create a good user experience for searching within the list." There is an example [in the demos](https://microsoftedge.github.io/Demos/selectlist/index.html) which adds a text input in the listbox to filter which addresses this top frustration, but based on what I've been hearing from accessibility experts, there should not be any interactive content inside the listbox element except for `<option>`s.

So my question for accessibility people is: what is the right pattern to create a select-like thing where you can filter the options? Is that just called a combobox, where instead of a button to open the listbox there is a text field?

I see in [this combobox in aria practices](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/) that the user can still sort of commit any text they want even if its not a valid option... what if you want to limit the possible choices to whats in the listbox like the one in the demos?

@scottaohara

YummyBacon5:
> there should not be any interactive content inside the listbox element except for \<option\>s

I think that this is only with the `listbox` role, but with the `<listbox>` element, the accessibility API mappings and content model could be different, to allow for this.

---

> where instead of a button to open the listbox there is a text field?

I feel that we shouldn't allow for the search input to replace the selectlist's button element.

Since then authors could use this element as more of a text autocomplete, than a `<select>` element.

But this then raises the question of how dynamic data would work. Since usually text autocompletes have loads of options which users can choose from, for example, a search engine's search bar. Where it'd be unpractical to have all these `<option>` elements.

The ability to search through the options would best work where data is static. For example, searching through countries to choose a location.

yisibl:
We desperately need a Selectlist with a search, for example in very common scenarios such as language or country selection, where the search helps the user to quickly find the desired item.

There is an experimental flag (`dom.forms.selectSearch`) in Firefox.

![image](https://github.com/openui/open-ui/assets/2784308/c0c4432b-1436-4ecd-ae6d-6f98d4090aac)

https://www.deepl.com/en/translator
![image](https://github.com/openui/open-ui/assets/2784308/ae0884c6-625a-41f6-8348-1c6ccb80112b)

kbrilla:
Also, serach value need to be accesible to js for server side filtering for example

una:
[...]

PR #1315, opened 2025-10-30
Create filterable select explainer

josepharhar:
This is for the resolution here: https://github.com/openui/open-ui/issues/847#issuecomment-3412348342

### DIFF
--- a/site/src/pages/components/filterableselect.mdx
+++ b/site/src/pages/components/filterableselect.mdx
@@ -1,0 +1,77 @@
+---
+menu: Active Proposals
+name: Filterable Select Element (Explainer)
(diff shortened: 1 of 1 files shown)


josepharhar:
The content of this explainer has so much overlap with https://github.com/openui/open-ui/pull/1313 that I'm thinking maybe I should just merge the two explainers into one. What do yall think?

josepharhar:
I merged this into https://github.com/openui/open-ui/pull/1313

josepharhar:
> Perhaps we should also consider what key behaviour does; would up/down arrows change the :focused-option?

I added text for this in the other PR

Issue #1316, opened 2025-10-31
[focusgroup] Should the entrance algorithm not consider direction?


janewman:
Created out of https://github.com/openui/open-ui/issues/1314#issuecomment-3469766888

I don't personally feel strongly here, and thought the "Existing" behavior was reasonable, but opening this up of discussion

### Existing behavior
When navigating using sequential focus navigation (tab/shift+tab) the current explainer considers the direction that the user is entering from when determining which focus group item should be focused (when memory is not a consideration).
This comes from: https://github.com/openui/open-ui/issues/537#issuecomment-1934930124
```html
<button>Before</button>
<div focusgroup="toolbar no-memory">
  <button>Bold</button>
  <button>Italic</button>
  <button>Underlined</button>
</div>
<button>After</button>
```
When focus is on "Before" and tab is pressed, "Bold" should be the initially focused item, hitting tab again brings us to "After".
If instead focus was on "After" and shift+tab was pressed, "Underlined" should instead be the initially focused item.

### Potential Change
Instead, we could not take the direction the user is navigating in into consideration, ensuring that both tab and shift+tab will always pick the same item.
```html
<button>Before</button>
<div focusgroup="toolbar no-memory">
  <button>Bold</button>
  <button>Italic</button>
  <button>Underlined</button>
</div>
<button>After</button>
```
When focus is on "Before" and tab is pressed, "Bold" should be the initially focused item, hitting tab again brings us to "After".
If instead focus was on "After" and shift+tab was pressed, "Bold" is still the first item to be focused.

sb3nder:
I think that not considering directionality should be the default behavior, and we should avoid adding an option to support directionality unless there are compelling real-world examples.

The only example I’m aware of that considers directionality is Safari native HTML radio groups:\
If none of the radio buttons are selected, moving focus into the radio group using Shift+Tab places focus on the last radio button instead of the first.\
(Firefox and Chrome, by contrast, always places focus on the first radio button.)

Additionally, @mcking65 made some strong arguments in this issue:
https://github.com/w3c/aria-practices/issues/2258
https://www.w3.org/WAI/ARIA/apg/patterns/radio/
> Based on today's discussion in the APG task force teleconference, the task force is unwilling to commit to introducing an inconsistency between Tab key behavior in the radio group pattern and Tab key behavior in other patterns that support the reversability of Tab navigation. Many members of the task force view Reversability (or "undo") as a valuable principle for Tab key navigation.
>
[...]

Issue #1314, opened 2025-10-30
[focusgroup] Should a new attribute be used to manage entry priority rather than tabindex?


janewman:
Today, an element with tabindex=-1 is **never** considered to be sequentially focusable, but focusgroup is attempting to change this in order to guarantee a single tab stop. Consider the following cases:
```html
<div focusgroup="toolbar">
  <button tabindex=-1>Bold</button>
  <button>Italic (recommended)</button>
  <button tabindex=-1>Underlined</button>
</div>
```
In the case above, the author wants the first focus to be on "Second", but since they didn't disable memory, they expect the user's last focused item to be restored when focus returns to the focusgroup, and this memory could be one of the other options. This is something I think we absolutely should support.
```html
<div focusgroup="toolbar">
  <button tabindex=-1>Bold</button>
  <button tabindex=-1>Italic</button>
  <button tabindex=-1>Underlined</button>
</div>
```
In the case above, we still need to ensure a tabstop in the focusgroup. While this would have been the equivalent to setting them all to "0" (or simply relying on native focusability), because the first example "just works" I think it is reasonable to assume authors would expect this to work as well.
_______
@mfreed7 [suggested ](https://chromium-review.googlesource.com/c/chromium/src/+/7042347/comment/eb13497c_3d9d1269/)focusgroup could avoid the need to use tabindex for managing entry priority with the use of a new attribute that would be set on a focusgroup item: `focusgroup-initial-focus`. I'm typically adverse to scope-creep but I agree with the sentiment that adding an element with tabindex=-1 to the tab order seems unnatural.

With this, the I think following changes would need to be made to the current proposal:
* The "Changes to sequential focus navigation" section would now no longer consider the value of tabindex to factor into the priority calculation.
  * Instead, `focusgroup-initial-focus` could be set on any elements that should be prioritized.
* An element with a tabindex=-1 would **not** be considered a focusgroup item and would not be managed by focusgroup at all.

Here are some examples of usage:
```html
<div focusgroup="toolbar">
  <button>Bold</button>
  <button focusgroup-initial-focus>Italic (recommended)</button>
  <button>Underlined</button>
</div>
```
With this, authors will actually have cleaner looking html, as they only need to set the attribute on a single element, rather than setting tabindex=-1 on all of the elements that shouldn't be prioritized.

```html
<button>before</button>
<div focusgroup="toolbar no-memory">
  <button>Bold</button>
  <button focusgroup-initial-focus>Italic (recommended)</button>
[...]

PR #1322, opened 2025-11-03
[focusgroup] Add entry priority attribute; don't consider direction

janewman:
Main changes in the PR:
## Entry Priority Control: New focusgroup-entry-priority attribute provides explicit control over initial focus target
Addresses #1314
With this, the tabindex value will no longer be used to determine priority, and items that are not keyboard focusable (tabindex="-1") cannot be focusgroup items. This helps focusgroup avoid needing to break the expectation that tab can never send focus to an element with tabindex=-1.

## Direction Independence: Same element receives focus regardless of entry direction (Tab vs Shift+Tab)
Addresses #1316

### DIFF
--- a/site/src/pages/components/scoped-focusgroup.explainer.mdx
+++ b/site/src/pages/components/scoped-focusgroup.explainer.mdx
@@ -787,9 +787,8 @@ The `focusgroup-entry-priority` attribute can be set on individual focusgroup it
 When multiple elements within the same [focusgroup segment](#focusgroup-segments) have the `focusgroup-entry-priority` attribute, the first element with the attribute in DOM order will be prioritized, or if the items are also being managed by reading-flow, the first such item in reading order. The direction of navigation (Tab vs Shift+Tab) does not affect which element is chosen - the same element will always receive focus when entering the segment, promoting consistent user experience.
 
 **Key behaviors:**
-- Elements with `focusgroup-entry-priority` are prioritized over elements that are merely sequentially focusable
-- If no element has `focusgroup-entry-priority`, the first sequentially focusable element in the segment is chosen (in DOM order, or if the items are also being managed by reading-flow, in reading order)
-- If no sequentially focusable elements exist, the first focusable element in the segment is chosen (excluding elements with negative `tabindex`) (in DOM order, or if the items are also being managed by reading-flow, in reading order)
+- Focusgroup items with `focusgroup-entry-priority` are prioritized.
+- If no focusgroup item has `focusgroup-entry-priority`, the first focusgroup item in the segment is chosen (in DOM order, or if the items are also being managed by reading-flow, in reading order).
 - Memory (when enabled) takes precedence over entry priority - a previously focused element will be restored even if other elements have `focusgroup-entry-priority`
 
 #### Memory vs Entry Priority Example
@@ -842,6 +841,7 @@ When focusgroups are nested, each focusgroup manages its own entry priority inde
   <button>Close</button>
   <button focusgroup-entry-priority>Exit</button>
 </div>
+<button>After</button>
 ```
 
 In this example, there are two distinct focusgroups, and three focusgroup segments:
@@ -856,7 +856,7 @@ In this example, there are two distinct focusgroups, and three focusgroup segmen
 1. **Tab into main toolbar:** Focus goes to "Print" (entry priority)
 2. **Tab from "Print":** Focus moves to "Italic" (entry priority in nested toolbar)
 3. **Tab from nested toolbar:** Focus moves to "Close" (continuing in main toolbar)
-4. **Shift+Tab into main toolbar from after:** Focus goes to "Print" (entry priority)
+4. **Shift+Tab into main toolbar from "After":** Focus goes to "Exit" (entry priority)
 
 **Directional navigation behavior:**
 - Arrow key navigation within the main toolbar moves between "Save", "Print", "Close", and "Exit", but skips over the nested toolbar


site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -265,24 +272,24 @@ menu has an additional submenu under "Tuition".
 ```html
 <ul focusgroup="menubar inline wrap" aria-label="Mythical University">
   <li role="none">
-    <a popovertarget="aboutpop" href="…" tabindex="-1" title="">About</a>
+    <a popovertarget="aboutpop" href="…" tabindex="0" title="">About</a>
```
gfellerph: Is `tabindex=0` and an empty title used intentionally here? These are not mentioned in the "what to notice" section below. Maybe the example could be simplified.

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -763,6 +770,141 @@ Focusgroups have the following additional features:
...
+  <button>Underlined</button>
+</div>
+```
+
+When multiple elements within the same [focusgroup segment](#focusgroup-segments) have the `focusgroup-entry-priority` attribute, the first element with the attribute in DOM order will be prioritized, or if the items are also being managed by reading-flow, the first such item in reading order. The direction of navigation (Tab vs Shift+Tab) does not affect which element is chosen - the same element will always receive focus when entering the segment, promoting consistent user experience.
+
+**Key behaviors:**
+- Elements with `focusgroup-entry-priority` are prioritized over elements that are merely sequentially focusable
+- If no element has `focusgroup-entry-priority`, the first sequentially focusable element in the segment is chosen (in DOM order, or if the items are also being managed by reading-flow, in reading order)
+- If no sequentially focusable elements exist, the first focusable element in the segment is chosen (excluding elements with negative `tabindex`) (in DOM order, or if the items are also being managed by reading-flow, in reading order)
```
gfellerph: I'm not sure if I'm interpreting these three points correctly. How do sequentially focusable and merely focusable elements differ? The first two points suggest that _sequentially focusable elements_ also describes the Bold and Underlined buttons even though they are not reachable by tab initially. That leaves me wondering how the third state can be achieved when `tabindex=-1` is being ignored by focusgroup navigation.


```suggestion
- Elements with `focusgroup-entry-priority` are prioritized over other focusgroup items
- If no element has `focusgroup-entry-priority`, the first focusgroup item in the segment is chosen (in DOM order, or if the items are also being managed by reading-flow, in reading order)
```

Would it be correct to write "focusgroup item" instead of "sequentially focusable element" and drop the third point?

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -763,6 +770,141 @@ Focusgroups have the following additional features:
...
+  <!-- Nested focusgroup with its own entry priority -->
+  <div focusgroup="toolbar" aria-label="Text formatting">
+    <button>Bold</button>
+    <button focusgroup-entry-priority>Italic</button>
+    <button>Underline</button>
+  </div>
+
+  <button>Close</button>
+  <button focusgroup-entry-priority>Exit</button>
+</div>
```
gfellerph, line 843: Because the after button is mentioned below.

```suggestion
<button>Before</button>
<div focusgroup="toolbar" aria-label="Main toolbar">
  <button>Save</button>
  <button focusgroup-entry-priority>Print</button>

  <!-- Nested focusgroup with its own entry priority -->
  <div focusgroup="toolbar" aria-label="Text formatting">
    <button>Bold</button>
    <button focusgroup-entry-priority>Italic</button>
    <button>Underline</button>
  </div>

  <button>Close</button>
  <button focusgroup-entry-priority>Exit</button>
</div>
<button>After</button>
```

site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -763,6 +770,141 @@ Focusgroups have the following additional features:
...
+2. **Tab from "Print":** Focus moves to "Italic" (entry priority in nested toolbar)
+3. **Tab from nested toolbar:** Focus moves to "Close" (continuing in main toolbar)
+4. **Shift+Tab into main toolbar from after:** Focus goes to "Print" (entry priority)
+
+**Directional navigation behavior:**
+- Arrow key navigation within the main toolbar moves between "Save", "Print", "Close", and "Exit", but skips over the nested toolbar
+- Arrow key navigation within the nested toolbar moves between "Bold", "Italic", and "Underline"
+- Each focusgroup maintains independent arrow navigation
+
+This demonstrates how entry priority works independently for each nested focusgroup and how exiting and entering nested focusgroups respects each segment's entry priority.
```
gfellerph, line 866: If each segment can have it's own entry priority, the above 4. step should focus on "Exit" instead of "Print".

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: This was a rather ugly example, thanks for calling this out.

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: Thanks for catching!

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: Absolutely, thanks for catching!

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: This wasn't right, thanks for calling out.
I put all of these in terms of "focusgroup items" and dropped the incorrect third point, with this I think this is much clearer.

janewman:
> In some areas the wording can be a bit confusing. Please understand these comments as suggestions. If they don't make sense, it's absolutely fine to reject them.

Thanks for the second set of eyes, all of your comments made sense to me, I really appreciate your help in ensuring clarity here.

Issue #1188, opened 2025-04-10
[menu] improve distinction between menubar and toolbar


josepharhar:
Based on the feedback here: https://github.com/openui/open-ui/issues/1179#:~:text=the%20top%20bar%20is%20a%20menu%20bar%2C%20and%20below%20that%20is%20a%20toolbar

It sounds like we should explicitly say whether the new element we are adding is a menubar or a toolbar, and perhaps link to some aria concept(s).

I think that we are adding a menubar, not a toolbar. I also think that the design we are going for does not preclude us from adding a toolbar element in the future.

Does this sound right @scottaohara ?

josepharhar:
I also just noticed this from the minutes:

> scott: toolbar pattern is important because ? dont allow for quick navigation. toolbars do allow this. if someone wants to jump to a thing they can do that. if theyre all menu items then they cant jump

How do you perform the quick navigation? Is it like a typeahead thing where you type the first letter of the item you want to jump to? Should we implement some keyboard behavior for this?

scottaohara:
>How do you perform the quick navigation?

this comment was made in context of using a screen reader.

with a screen reader and i come to a toolbar of items - i can use my screen reader's special hot keys to navigate to the type of element I want.  e.g., buttons vs other form controls.  I can bring up a dialog / rotor of available elements on the page and jump to the one i want.

menuitems do not have these commands.  there is no way to distinguish a menuitem from a menuitemradio from a menuitemcheckbox without explicitly navigating to it.  Again, where as if i have a toolbar consisting of buttons, checkboxes and a combobox - i can use my screen reader to jump to those individual items, if i know what i'm looking for / dont' want to interact with the "tools" that are of no interest to me at that point.

[...]

PR #1324, opened 2025-11-04
Rewrite the menu elements explainer (1/2)

### LINKED ISSUE #1191: [menu] Explicitly mention support for HR as a separator
It's common to visually separate menu items and select supports this so it seems ~~good~~ consistent for menu to do this too?

### LINKED ISSUE #1193: [menu] Navigation vs menu items use case
During the meeting where we presented the menu elements proposal for incubation ([minutes](https://www.w3.org/2025/03/20-openui-minutes.html)) ([explainer](https://open-ui.org/components/menu.explainer/)), @scottaohara raised the topic of navigation menu bars (https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/) being particularly thorny, and being a different beast than what I'll call the "normal menubar" use case, which is described by https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/.

In particular, see these lines from the minutes:

```
sorvell: navigation is a deep can of worms, people have strong opinions about navigations and menus

...

scott: im going to talk about navigation use case

scott: that is one thats on the apg and is the most contentious

scott: one of the reasons that navigation use case is not preferred is that menuitems are not navigable using screenreaders. theyre meant to be treated like navigation to menus in macos windows etc

scott: as far how theyre used on the web and used for navigation purporses, they can make finding what youre looking for harder for screenreader users because you cant just bring up a list of hyperlink

scott: if the navigation menubar is to go forward, we would need to do work thats never been done before to make this a thing that is not just another way of saying menu

...

scott: i am not opposed to it moving forward, but theres a lot of work moving forward

scott: if its just going to be menu but it has type=nav but no actual difference then we havent helped anyone
```

Somewhere in the minutes I say:

```
dom: thanks, i understand why its a big deviation
```

...but either I was wrong at the time, or I've forgotten. Some of us at Google have been trying to understand how the navigation use case is so different form the "normal (editor example) menubar" use case. Even if we decide navigation is too scary to tackle right away, I still think we need to have this discussion because I fear if we don't design something for this use case, developers will just use the "normal menubar" primitives to achieve this, by stuffing navigating actions/commands in each `<menuitem>`, thus ossifying a poor accessibility experience.

So I think we have a few questions, much of which we intend to discuss at the next meeting:

 - Why is the navigation use case different?
    - The real question here I think is: what would need to be semantically distinct in a `<navbar>` from `<normalmenubar>` in order to make `<navbar>` suitable for the navigation use case? What's the real delta between these
    - @scottaohara mentions "they can make finding what youre looking for harder for screenreader users because you cant just bring up a list of hyperlink", but I don't fully understand this. Why can you just "bring up a list of hyperlinks"? In the non-navigation menubar case, we're certainly going to "just bring up a list of menuitems"—does that have the same accessibility concerns?
 - Is it actually bad if developers manually navigate the page in response to `<menuitem>` clicks? This could be done
 - Multi-page app vs single-page app: I can imagine developers implementing SPA navigations with a normal menubar that responds to `<menuitem>` activations to swap in new content, since they don't need traditional anchor semantics. Is the "SPA use case" considered the same as the "navigation use case", and would it be a bad outcome if developers this? MPAs and SPAs have the same navigation intent: show new content to the user when they click something, but they do this very differently. Is the "navigation intent" what's scary about the navigation use case? Or is it just the raw-anchor-tags-inside-of-the-menuitem MPA case that's scary about the navigation use case?

/cc @dizhang168 @mfreed7 @josepharhar

domfarolino:
This PR rewrites most of the menu elements explainer. The changes are too vast to describe in detail here, but at high level it:

 - Leads with punchier visuals and markup showing the menu elements in action
 - Refactors and rewords almost all of the descriptions of the new elements, adding more details where missing
 - Goes into more detail on the content model for each element, and points to OpenUI issues where more discussion is needed
 - Moves the accessibility section up before "General questions", as it is more part of the "proposal"
 - Describes our resolution of the navigation menubar pattern discussion, including prohibiting the `<a>` from the content model in menu lists and menu bars, and references default styles of `display: none;` for them in this case; also describes the new navigation menulist proposal that's coming soon
 - Tries to elaborate more on the distinction between toolbar and how it relates to this proposal—but not much
 - Refactors lots of the General Questions

The title says (1/2) because next we should:
 - Rearrange or refactor some of the "General questions"
 - Do something with the "Examples & code snippets" section, since it kind of competes with the examples I added to the top of this, but also kind of supplements it. I'm not sure the best fate for it yet.
 - Reorganize lots of the random scattered links towards the bottom of the explainer
 - Keep a running list of every menu elements issue that's been resolved, somewhere towards the bottom

Closes #1193. Closes #1191. Makes progress on #1188.


site/src/pages/components/menu.explainer.mdx
```diff
@@ -136,70 +146,292 @@ elements. We envision the following markup:
...
+still be keyboard reachable and focusable, just not activatable. See [Issue
+#1274](https://github.com/openui/open-ui/issues/1274) for more discussion.
+
+### Keyboard behavior
+
+See [§ What should the keyboard behavior be?](#what-should-the-key-board-behavior-be) for details.
+
+### Navigation menubars
+
+In short, there are longstanding accessibility issues with navigation menus are our proposal does
```
mfreed7: s/are/and

site/src/pages/components/menu.explainer.mdx
```diff
@@ -26,101 +26,111 @@ layout: ../../layouts/ComponentLayout.astro
...
 
-* Add an in-page horizontal menubar at the top of a content window; example: Google Doc has a menubar with the items File, Edit, View, Insert, Format, etc.
-* Add a filter button that opens a menulist to a shopping catalogue where user can check between "Sort by relevant", "Sort by newest", "Sort by price".
+## Examples & code snippets
 
+The goal of our proposal is to make it easy for web developers to author accessible menus with very
+little markup and JavaScript.
 
-## New `<menuitem>` elements
+<img src="/images/menu-elements-example-1.png" alt="menu elements expanded showing commands">
```
mfreed7: I think the code should be in a code block here, so it's copy/pastable. The screenshot is great, but should just be the screenshot, not including the code.

site/src/pages/components/menu.explainer.mdx
```diff
@@ -26,101 +26,111 @@ layout: ../../layouts/ComponentLayout.astro
...
+Commands](https://open-ui.org/components/invokers.explainer/), and [CSS anchor
+positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning), much of the
+behaviors needed by the menu use case are provided natively, and do not need to be re-invented by
+web developers or library authors.
 
-As stated, our proposal involves introducing a single `<menuitem>` element to represent menu
-buttons, some of which can be checkable. Instead of representing both non-checkable & checkable
-state-bearing menu items with the single `<menuitem>` element, we could also consider:
+Further, these elements will have the correct implicit ARIA roles, states, and properties, as well
+as a keyboard interaction & focus model that both sighted and accessibility technology users expect.
```
mfreed7: I'd just say "that users expect". The rest of the phrasing is troublesome, since e.g. there are sighted AT users.

site/src/pages/components/menu.explainer.mdx
```diff
@@ -26,101 +26,111 @@ layout: ../../layouts/ComponentLayout.astro
...
+This proposal introduces the `<menubar>`, `<menulist>`, and `<menuitem>` elements to support
+application menus on the web. These menus will:
 
-Use cases:
+ - Represent commands as `<menuitem>`s, grouped together in a `<menulist>` popover or in-page
+   `<menubar>`.
+ - Represent state in checkable and radio-style menu items.
+ - Provide the correct implicit ARIA roles and keyboard interaction model out of the box.
+ - Give web developers default styles that create functional menus without having to re-invent
+   common popover and anchor positioning patterns, or re-implement ARIA patterns.
```
mfreed7, line 43: I'd add a note, right here, that this explainer is for "application menus" and not "navigation menus" and point to the (perhaps placeholder) explainer for the other feature. Just to get that right up front.

site/src/pages/components/menu.explainer.mdx
```diff
@@ -26,101 +26,111 @@ layout: ../../layouts/ComponentLayout.astro
...
+* Provides the implicit [`menubar`](https://w3c.github.io/aria/#menubar) ARIA role.
+* Provides the correct keyboard and focus behavior across child menu items and
+  sub menulists.
 
-For example, a text editor might need to provide a `<menulist>` to provide basic font style adjustments. A group of checkable `<menuitem>` elements to mark content as "Bold", "Italic" or "Underlined", could be provided.  The checked state representing whether the text is currently "Bold" (checked) or not.
+**Content model**: the `<menubar>` element allows the following elements as descendants:
+* `<menuitem>`
+* Anchor `<a>` tags are disallowed in menu bars, due to accessibility concerns around exposing
+  links in ARIA menus, and concerns around the navigation menubar pattern. We enforce this with the
+  content model, and propose default `display: none` styles for anchors that are direct descendants
```
mfreed7: direct descendants or any descendants? I'd think perhaps the latter?

site/src/pages/components/menu.explainer.mdx
```diff
@@ -26,101 +26,111 @@ layout: ../../layouts/ComponentLayout.astro
...
-`<menulist>`
+* Provides the implicit [`menu`](https://w3c.github.io/aria/#menu) ARIA role.
+* Has default anchor positioning styles that anchor it to its
+  [implicit anchor](https://drafts.csswg.org/css-anchor-position/#implicit-anchor-element)—the
+  button or menuitem that invoked it.
+* Supports the `:popover-open` pseudo-class, as it is a native **popover**.
+* Supports the
+  [`toggle` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/toggle_event), as it
+  is a native **popover**.
+* Can be invoked with the existing `show-popover` command, and the new `show-menu` command. It is
```
mfreed7: I'd remove `show-popover` here. We should be pointing to `show-menu` for menu commands. They do differ (see my latest CLs) in some behavior, such as popover stacking.

site/src/pages/components/menu.explainer.mdx
```diff
@@ -26,101 +26,111 @@ layout: ../../layouts/ComponentLayout.astro
...
+* A common ancestor for checkable `<menuitem>`s. This supports both exclusively checkable and
+  multi-checkable menu items.
 
-Our proposal includes adding a `checkable` attribute on `<fieldset>`, to group `<menuitem>`s as
-either radios or checkboxes, instead of introducing new `<menuitemcheckbox>` and `<menuitemradio>`
-elements. We envision the following markup:
+**Checkable menu items**
+
+To support checkable menu items, we propose the new `checkable` **content** attribute on
+`<fieldset>` (no corresponding **IDL** attribute), to group items as either radios or checkboxes;
```
mfreed7: nit: HTML will require you to add an IDL reflection. Any reason not to provide one?

site/src/pages/components/menu.explainer.mdx
```diff
@@ -136,70 +146,292 @@ elements. We envision the following markup:
...
+describes the concrete implications on our proposal.
+
+### Native ARIA role mapping
+
+ - The `<menubar>` element exposes the implicit ARIA
+   [`menubar`](https://w3c.github.io/aria/#menubar) role.
+ - The `<menulist>` element exposes the implicit ARIA
+   [`menu`](https://w3c.github.io/aria/#menu) role.
+ - The `<menuitem>` element exposes the implicit ARIA
+   ['menuitem'](https://w3c.github.io/aria/#menuitem) role by default.
```
mfreed7: Nit: use backticks for `menuitem`

site/src/pages/components/menu.explainer.mdx
```diff
@@ -370,96 +583,51 @@ We understand it would be helpful to have this feature be natively supported. If
...
-2. I tried Voiceover on google docs and tab navigation for the menus doesn't seem to work very well but when I am able to tab to Help => Privacy Policy (opens a new tab), it's not announced as a link.
+ - The `<nav>` element, which implicitly exposes the ARIA landmark
+   [`navigation`](https://w3c.github.io/aria/#navigation) role.
+    - Since hyperlinks and buttons are commonly children of this sort of navigation menu, authors
+      generally do not need to declare specific ARIA roles to be accessible, and users can rely on
+      standard <kbd>Tab</kbd> keey navigation to reach each interactive navigation element.
+ - The abused [Navigation Menubar Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/)
+    - These generally have poor accessibility characteristics, and are not recommended.
+ - The [Disclosure Navigation Menu Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/)
```
mfreed7: This link is identical to the one just above, which is confusing since one says "this is bad" and the other says "this is better".

site/src/pages/components/menu.explainer.mdx
```diff
@@ -370,96 +583,51 @@ We understand it would be helpful to have this feature be natively supported. If
...
-While we think it could be possible to address them to the satisfaction of the accessibility
-community, and users of accessibility technology, this usage of menus is still very much under
-discussion. See [§ How we can use menu elements to support navigation
-menus](#how-can-we-use-menu-elements-to-support-navigation-menus), and
-https://github.com/openui/open-ui/issues/1193.
 
 ## Examples in code
 
-
 ### Google Docs: nested menulist
```
mfreed7, line 702: Not modified here, but as I'm reading this explainer, the nested menulist example code is weird, since it doesn't use `command` and `commandfor` to link the main menubar to the submenu menulists. And it uses the outdated `checkable=multiple` syntax. Might need a quick cleanup?

mfreed7 (approved):
LGTM with some comments. Ok to land with fixes here or in a followup.

site/src/pages/components/menu.explainer.mdx
domfarolino: Good idea. Making this change now.

site/src/pages/components/menu.explainer.mdx
domfarolino: I added a note but since we don't have an explainer yet and I don't want to block this on landing one first, I just mentioned that the new proposal for that use case is not yet published.

site/src/pages/components/menu.explainer.mdx
domfarolino: I think I agree. Done.

site/src/pages/components/menu.explainer.mdx
domfarolino: Hmm, I'm not sure why I had it this way originally, but yeah I see no special reason to avoid reflection.

site/src/pages/components/menu.explainer.mdx
domfarolino: Oops! Nice.

site/src/pages/components/menu.explainer.mdx
domfarolino: Yep, this needs to be https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/ instead. Fixed.

site/src/pages/components/menu.explainer.mdx
domfarolino: Good call. Done.

PR #1345, opened 2025-12-23
sources: update W3 component links

nektro:


keithamus (approved):
Excellent work. Thanks for doing this.

PR #1347, opened 2025-12-30
sources: Lion: fix links, add missing components, alphabetize list

nektro:


site/src/sources/lion.json
```diff
@@ -6,15 +6,36 @@
...
+    {
+      "name": "Collapsible",
+      "openUIName": "Collapse",
+      "url": "https://lion-web.netlify.app/components/collapsible/"
+    },
     {
       "name": "Combobox",
       "openUIName": "Select",
-      "url": "https://lion-web.netlify.app/components/inputs/combobox/overview/"
+      "url": "https://lion-web.netlify.app/components/inputs/combobox/"
```
lukewarlow: This URL seems to be wrong, at least some components notably combobox does seem to require the /overview/ on the end.

Also I think the domain is https://lion.js.org (at least that's what shows on Google?)

lukewarlow (changes requested):
You also need to fix the IPR issues. I believe this can be done by signing up as a member of the community group via the w3c website.

lukewarlow:
Cc @keithamus probably knows the IPR side better.

keithamus:
Yeah I think for small changes which aren't "ideas" the IPR can be marked 'non substantiative' but this might toe the line into "substantiative" so it's probably best @nektro if you can sign the IPR, if you're comfortable doing so. The steps for this involve creating (or signing into your existing) W3 account https://auth.w3.org/login then linking it with your GitHub profile, then finally joining the [Open UI Community Group](https://www.w3.org/community/open-ui/join)

nektro:
Signed and Joined.

nektro:
bump

PR #1350, opened 2026-01-05
Rename :focused-option to :active-option

josepharhar:
This rename was recommended at TPAC and was not previously discussed anywhere: https://www.w3.org/2025/11/11-openui-minutes.html


keithamus (approved):
👍

PR #1361, opened 2026-01-16
Clarify step-up and step-down are valid on range inputs too.

lukewarlow:


lukewarlow:
Not sure if this is only applicable to range and number, what about say date? Is it useful there? For now apply to just these two but we should think about this.

lukewarlow:
Review ping @keithamus @mfreed7

PR #1374, opened 2026-02-15
Updating the beforefocus proposal.

mshoho:
Updates based on currently ongoing feature-complete implementation for Chromium.

### DIFF
--- a/site/src/pages/components/beforefocus.explainer.mdx
+++ b/site/src/pages/components/beforefocus.explainer.mdx
@@ -73,7 +73,7 @@ dictionary AdvanceFocusOptions {
   Element? startElement;
   Element? stayWithinElement;
   boolean backward = false;
-  boolean wrap = true;
+  boolean wrap = false;
   boolean includeStartElement = false;
 };
 
@@ -136,12 +136,12 @@ menus, focus traps in custom modal dialogs and similar UI patterns.
 Controls whether focus wraps around when reaching the boundaries.
 
 When `stayWithinElement` is specified:
-- `wrap: true` (default): Focus wraps from last element back to first (or vice versa for backward)
-- `wrap: false`: Focus stops at the boundary and `advanceFocus()` returns `false`
+- `wrap: true`: Focus wraps from last element back to first (or vice versa for backward)
+- `wrap: false` (default): Focus stops at the boundary and `advanceFocus()` returns `false`
 
 When `stayWithinElement` is not specified:
-- `wrap: true` (default): Focus wraps from the end of the page to the beginning
-- `wrap: false`: Focus can move to the browser's address bar (normal Tab behavior)
+- `wrap: true`: Focus wraps from the end of the page to the beginning
+- `wrap: false` (default): Focus can move to the browser's address bar (normal Tab behavior)
 
 ```js
 // Stop at boundaries instead of wrapping


site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -1,246 +1,462 @@
 ---
 menu: Active Proposals
-name: beforefocus/focusNext (Explainer)
+name: advanceFocus/beforefocus (Explainer)
 layout: ../../layouts/ComponentLayout.astro
 ---
 
 Authors: [Marat Abdullin](https://github.com/mshoho)
 
-{/* START doctoc generated TOC please keep comment here to allow auto update */}
```
lukewarlow, line 9: This looks wrong to me.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -1,246 +1,462 @@
...
+interface BeforefocusEvent extends UIEvent {
+  target: Element;        // The element about to receive focus
+  detail: FocusType;      // Numeric code indicating the reason for focus
+  bubbles: true;
+  cancelable: true;
+  composed: true;         // Crosses shadow DOM boundaries
+}
+
+enum FocusType {
+  kScript = 1,    // element.focus() was called
```
lukewarlow: Don't use chromium naming here. These shouldn't have the k prefix.

Also is this better off being like a webidl enum than a typescript enum? Aka string values of "script" etc.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -1,246 +1,462 @@
...
+
+## Opting In
+
+Both `document.advanceFocus()` and the `beforefocus` event require explicit opt-in via a meta tag. 
+This design choice ensures backward compatibility and gives developers control over when to enable
+these new behaviors:
+
+```html
+<meta name="focus-navigation" content="advanceFocus beforefocus">
+```
```
myakura, line 43: > Both `document.advanceFocus()` and the `beforefocus` event require explicit opt-in via a meta tag. 
> This design choice ensures backward compatibility and gives developers control over when to enable
> these new behaviors:

Both methods looks new to me. So I wonder why this meta keyword need for them. And by "ensures backward compatibility", does the meta keywords affect other focus-related APIs?

site/src/pages/components/beforefocus.explainer.mdx
mshoho: Thanks, lost it during the merge of two documents. It's back.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: I have removed this interface from the description altogether, it does not matter much at this stage. I have also reordered the sections to put the Note about the future of this property and its current source on top of the property description. At this point using `detail` was just free for my implementation, and I think it is ok for a prototype. If there is a consensus to actually make it a standard, then of course it will have to undergo all due consideration.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: I have rephrased it. I was mostly meaning that the event won't be happening unless there is a clear intent to have it.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -14,233 +14,446 @@ Authors: [Marat Abdullin](https://github.com/mshoho)
...
-`Shift+Tab` presses.
+## Opting In
 
-The method's signature should be like:
+Both `document.advanceFocus()` and the `beforefocus` event require explicit opt-in via a meta tag. 
+This design choice ensures explicit developer intent and prevents the Universe from overheating
+by unnecessarily dispatching new events on every page:
+
+```html
+<meta name="focus-navigation" content="advanceFocus beforefocus">
```
keithamus, line 42: I don't think this meta tag will be necessary.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -14,233 +14,446 @@ Authors: [Marat Abdullin](https://github.com/mshoho)
...
-`direction` is the direction of the search. `FORWARD` to get the `Tab` press behaviour and
-`BACKWARD` to get `Shift+Tab` press behaviour. It is also optional and equals to `FORWARD` by
-default.
+When `stayWithinElement` is specified:
+- `wrap: true` (default): Focus wraps from last element back to first (or vice versa for backward)
+- `wrap: false`: Focus stops at the boundary and `advanceFocus()` returns `false`
+
+When `stayWithinElement` is not specified:
+- `wrap: true` (default): Focus wraps from the end of the page to the beginning
+- `wrap: false`: Focus can move to the browser's address bar (normal Tab behavior)
```
keithamus: We discussed this at length with Dialog and it was determined that the user should always be able to traverse to the browser's UI. So I don't think this is something we can reasonably enable here.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -14,233 +14,446 @@ Authors: [Marat Abdullin](https://github.com/mshoho)
...
+### Event Properties
 
 ```ts
-interface BeforeFocusEvent extends FocusEvent {
-  reason: FOCUS_FORWARD | FOCUS_BACKWARD | PROGRAMMATIC | AUTOFOCUS | UI
+interface BeforefocusEvent extends UIEvent {
+  target: Element;        // The element about to receive focus
+  detail: FocusType;      // Numeric code indicating the reason for focus
+  bubbles: true;
+  cancelable: true;
```
keithamus: I am quite sure we will not be able to make focus cancelable. I don't think it is possible in certain cases, but in cases where it is it would be quite a poor user experience to cancel these.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -14,233 +14,446 @@ Authors: [Marat Abdullin](https://github.com/mshoho)
...
+The `beforefocus` event is dispatched *before* the browser wants to focus an element. Calling
+`preventDefault()` on the event prevents the element from receiving focus.
 
-The event should have an important property to distinguish the reason for the focusing:
+### Event Properties
 
 ```ts
-interface BeforeFocusEvent extends FocusEvent {
-  reason: FOCUS_FORWARD | FOCUS_BACKWARD | PROGRAMMATIC | AUTOFOCUS | UI
+interface BeforefocusEvent extends UIEvent {
```
keithamus: Providing a lot more detail about why this is needed when `focus`, `blur`, `focusin` and `focusout` events all exist is needed, I think.

site/src/pages/components/beforefocus.explainer.mdx
```diff
@@ -14,233 +14,446 @@ Authors: [Marat Abdullin](https://github.com/mshoho)
...
+
+| Value | Name | Triggered By |
+|-------|------|--------------|
+| `0` | None | Catch-all (includes autofocus and accessibility-based focusing) |
+| `1` | Script | `element.focus()` |
+| `2` | Forward | `Tab` key, `advanceFocus()` |
+| `3` | Backward | `Shift+Tab`, `advanceFocus({ backward: true })` |
+| `4` | SpatialNavigation | Spatial navigation (TV/game controller interfaces) |
+| `5` | Mouse | Mouse click on a focusable element |
+| `6` | AccessKey | Access key activation (e.g., `Alt+Key` on Windows) |
```
keithamus: It might be worth checking other implementations and what they do here. I imagine not all of these will be possible.

In addition to this, exposing some of these are _certainly_ a vector for fingerprinting which will be unlikely to get traction in WHATWG.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: Well, there was a concern during the previous discussion that sending a bunch of new events could be a problem. So, I made it opt-in. No events sent by default, only when you know what you are doing.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: I generally agree. I am even the one who came up with the hack allowing to do it long before any support in the native modal (by synchronously focusing last/first element in the app during the Tab press handler and allowing the default action to move focus out), and I was defending the necessity of this behavior a lot over the years. And I still believe it is correct. But from nearly a decade of my experience being massively dedicated to solving these particular problems, from time to time there still are now or never cases when this is needed. Exposing this prop or not, it is _already doable_ using existing API. Mind that a web page is not the only place where these things are applicable. Electron and WebView based applications do need to have more control over what is going on. Trapping focus is already doable by intercepting Tab and preventing the default behavior for it.

But the key point is that _the dialogs are not target of this particular prop_ at all. The primary targets are circular arrow key navigable lists/tables. When Tab still works moving you there and out. But arrow keys allow to conveniently move around the particular component. Without this prop it will still be possible, but it will just require more work.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: Could you please list the cases? I am not sure I understand the point about poor user experience. It already works in my Chromium prototype and it has been fantastic user experience for me so far, allowing to solve problems that are currently literally not solvable. The user experience is created by the people who create the application. It is poor or good because of that. But it also tends to lean towards the poor side when there are obstacles not allowing it to be good. Having canvas allows people to create very poor and inaccessible experiences manually drawing the buttons on the canvas, but that's on the people who use it that way and it is not a reason to not have canvas support.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: I will add a few examples of cases that are hard or impossible to solve.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: I have it in the note, that this is not what it is going to look like. This is just what Chromium gave me for free right now and I have exposed it for prototyping/testing. But I agree, this is confusing, because this proposal overall is not about the particular prototype, but about the thing in general. I will adjust it. In reality I see only 5 needed source options: Browser (like autofocus), Script (programmatic element focus), Forward and Backward, and UI interaction (it shouldn't be named Mouse, because it could be touch or any other mean of interacting with the UI).

mshoho: Adjusted the event description to reduce confusion. Added `reason` property with the options above, converted to WebIDL format. Also adjusted advance focus options interface and method definition to WebIDL for consistency.

lukewarlow:
Thanks for working on this by the way! It sets us in a good position to discuss the API details further.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: I have changed the default initial example to:
```ts
const list = document.getElementById('my-list');
document.advanceFocus({ stayWithinElement: list });
```
And the order of use cases in the description.

mshoho: And flipped the default `wrap` value to be going to the address by default.

site/src/pages/components/beforefocus.explainer.mdx
mshoho: Just a heads up, that I am creating several examples with the descriptions (planning to finish it by the end of this week and update the PR). But in the meantime I have created a fully featured implementation for Firefox too (just to make sure it is doable there as well). So, I have both Chromium and Firefox implementations now (including WPT tests). Once this PR is merged, I'll start the intent to prototype posts (in both Chromium and Firefox) and publish both PRs. I want to use this document as a reference description.

lukewarlow:
@keithamus do you have further changes you need here?

PR #1379, opened 2026-02-23
[focusgroup] Extend child role inference to button, add default modifiers

janewman:
This PR contains three substantive changes:
1. Child role inference now additionally applies to `<button>` elements, since buttons are the dominant building block for composite widget items (tabs, menuitems, etc.). 
2. Behavior tokens gain default modifiers (e.g., tablist → inline wrap, menu → block wrap), reducing boilerplate in common patterns. A new no-wrap token lets authors explicitly disable a behavior token's default wrapping.
3. Added new modifier `nowrap`, `no-memory` was renamed to `nomemory` to follow convention.
**Example for what changes with this PR**
Today (as written without this change):
```html
<div focusgroup="tablist inline wrap">
  <button role=“tab”>Tab 1</button>
  <button role=“tab”>Tab 2</button>
  <button role=“tab”>Tab 3</button>
</div>
```
OR
```html
<div focusgroup="tablist inline wrap">
  <div tabindex=“0”>Tab 1</div>
  <div tabindex=“0”>Tab 2</div>
  <div tabindex=“0”>Tab 3</div>
</div>
```
Proposed change, additionally allowing role inference on button elements and reasonable defaults:
```html
<div focusgroup="tablist">
  <button>Tab 1</button>
  <button>Tab 2</button>
  <button>Tab 3</button>
</div>
```

In each example, the tree exposed to screen readers is identical, a tablist container with three tab children.
Rather than a developer needing to use a div/span to get the role inference, but needing to use tabindex, or using buttons but then having to manually specify the role, with this, developers can just follow the path of least resistance and find themselves with accessible content. In addition, developers won't need to specify any modifiers as the behavior token sets up the expected defaults.

| Behavior | APG Pattern | Minimum container role (when applied) | Minimum child role(s) (when applied) | Default modifiers | APG Pattern Link |
|---|---|---|---|---|---|
| `toolbar` | Toolbar | [`toolbar`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/toolbar_role) | (none) | inline | [APG Toolbar](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/) |
| `tablist` | Tabs | [`tablist`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role) | [`tab`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role) | `inline` `wrap` | [APG Tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) |
| `radiogroup` | Radio Group | [`radiogroup`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radiogroup_role) | [`radio`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radio_role) | (none) | [APG Radio Group](https://www.w3.org/WAI/ARIA/apg/patterns/radio/) |
| `listbox` | Listbox | [`listbox`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role) | [`option`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/option_role) | (none) | [APG Listbox](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/) |
| `menu` | Menu | [`menu`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role) | [`menuitem`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role) | `block` `wrap` | [APG Menu / Menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menu/) |
| `menubar` | Menubar | [`menubar`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menubar_role) | [`menuitem`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role) | `inline` `wrap` | [APG Menu / Menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menu/) |
| `grid` (future) | Grid (2-D) | [`grid`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/grid_role) | (TBD) | (TBD) | [APG Grid](https://www.w3.org/WAI/ARIA/apg/patterns/grid/) |

If authors have something like contextual actions on these tabs, this plays nicely as in most cases these contextual buttons wouldn't want to be included in arrow-key navigation, see:

```html
<div focusgroup="tablist">
  <button>Tab 1</button>
  <button id="contextual-actions-for-tab-1" focusgroup="none">...</button>
  <button>Tab 2</button>
  <button id="contextual-actions-for-tab-2" focusgroup="none">...</button>
  <button>Tab 3</button>
  <button id="contextual-actions-for-tab-3" focusgroup="none">...</button>
</div>
```

### DIFF
--- a/site/src/pages/components/scoped-focusgroup.explainer.mdx
+++ b/site/src/pages/components/scoped-focusgroup.explainer.mdx
@@ -80,7 +80,7 @@ Authors may use the proposed `focusgroup` HTML attribute to declare that
 a subtree of focusable elements will get:
 1. Focus navigation ([not selection](#non-goals)) using directional navigation (arrow key, D-pads, etc.).
 2. A guaranteed tab stop (when at least one focusable element is present) (see [Guaranteed tab stop](#guaranteed-tab-stop)).
-3. Automatic return to the last focused focusable element (unless [`no-memory`](#disabling-focusgroup-memory) is set).
+3. Automatic return to the last focused focusable element (unless [`nomemory`](#disabling-focusgroup-memory) is set).
 4. Optional limited-axis arrow key navigation and optional wrap-around semantics.
 
 By standardizing `focusgroup`, authors can leverage these behaviors in
@@ -141,7 +141,7 @@ What changed:
 Simple usage (space-separated tokens):
 
 ```html
-<div focusgroup="<behavior> [inline|block] [wrap|no-wrap] [no-memory]">
+<div focusgroup="<behavior> [inline|block] [wrap|nowrap] [nomemory]">
   <button>One</button>
   <button>Two</button>
   <button>Three</button>
@@ -155,7 +155,7 @@ Opting an element and its subtree out of an ancestor focusgroup, see: [Opting-ou
 
 Specifying a specific item as the [start element](#the-focusgroupstart-attribute) for focus when entering a focusgroup:
 ```html
-<div focusgroup="<behavior> [inline|block] [wrap|no-wrap] [no-memory]">
+<div focusgroup="<behavior> [inline|block] [wrap|nowrap] [nomemory]">
   <button>One</button>
   <button focusgroupstart>Two</button>
   <button>Three</button>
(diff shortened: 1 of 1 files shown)


smhigley:
I love the change, as you already know :D

One nit -- I think toolbar should probably have a default of `inline` as well.

janewman:
As discussed during the call, I'll update all modifiers to **not** use a dash.

(fixed in latest patchset)

janewman:
> I love the change, as you already know :D
> 
> One nit -- I think toolbar should probably have a default of `inline` as well.

Fixed in latest patchset.

janewman:
Reviewers, PTAL and if there are no new comments or changes needed, please feel free to merge.

lukewarlow:
This looks good to me I've given it a medium level review. Any issues we can fix in follow ups.

lukewarlow:
Ah turns out I'm not a code owner...

janewman:
@mfreed7, thanks for reviewing, while I got some permissions to be able to label issues, I still don't have the ability to see the "merge pull request" button, please go ahead and merge if you feel this is ready to go.

PR #1392, opened 2026-03-06
some of these badges are tags

nektro:


keithamus (approved):
Thanks for this, LGTM!

Issue #1412, opened 2026-03-09
[focusgroup] Should listbox imply block modifier?


lukewarlow:
After https://github.com/openui/open-ui/pull/1379 we now have some behaviours implying default modifiers. Namely toolbar, tablist, menu and menubar.

I'm wondering if listbox should imply block? The listbox aria role implies aria-orientation vertical and most listboxes are rendered vertically. Also `<select size=4>` navigates via vertical arrows but *not* horizontal arrows.

janewman:
I have a slight preference for not adding any axis restrictions by default, but more than happy to discuss if folks don't think this fits.

lukewarlow:
I'm curious do you have a specific thinking on why? Just so I can think it over before the meeting.

janewman:
Just a slight preference "in general". If there is any amount of consensus in the other direction, I'm happy to make the change.

mfreed7:
Actual meeting notes [are here](https://docs.google.com/document/d/1EdbZ2S7m60Lmoa2k0S1QOTD0A8jRsly6ktORXu5FSYg/edit?tab=t.2wxxk63mz70f). (I cleaned them up manually, but if I messed something up please ping me privately and I'll correct it.)

PR #1418, opened 2026-03-12
[focusgroup] update default modifiers for radiogroup and listbox

### LINKED ISSUE #1398: [focusgroup] Should toolbar get the wrap modifier by default?
Currently menu, menubar, and tablist get wrap behaviour, should this apply to toolbar too?

lukewarlow:
Fixes #1398 and #1412


janewman (commented):
exactly what we discussed, lgtm!

janewman:
> exactly what we discussed, lgtm!

Actually, are there examples or explanations that need to be updated with this?

lukewarlow:
> Actually, are there examples or explanations that need to be updated with this?

Good point, I'll have a look.

lukewarlow:
Turns out there's a few other examples to update. Good catch.

Updated

site/src/pages/components/scoped-focusgroup.explainer.mdx
janewman: Can you update this to be just "listbox" now that block is the default?

janewman: Can you make this just "radiogroup" since wrap is the default?

janewman (approved):
lgtm % removing the now-default modifier in examples.

lukewarlow:
@keithamus rubber stamp would be appreciated.

keithamus (approved):

Issue #1420, opened 2026-03-13
[toolbar] Should toolbar have a disabled attribute and how should it behave?


lukewarlow:
Should we include a disabled attribute on toolbars? If so what should it do? Similar to the question in https://github.com/openui/open-ui/issues/1419 possibly it should affect children by setting aria-disabled and disabling activation but not affect focusability?

scottaohara:
i think it'd be nice if `<toolbar disabled>` worked.  similar to `<fieldset disabled>`

see the other issue for commentary on focusability questions

out of scope a bit.... but tbh, i could just see `toolbar` being exactly the same as `fieldset` but with an implicit focusgroup=toolbar.  as it'd be nice if `legend` worked similarly with `toolbar` (as i can think of a few instances where labeling a toolbar would be nice / and it'd be nice to be able to not have to use aria to hookup the label / accname for the toolbar)

lukewarlow:
Well I guess that answers  https://github.com/openui/open-ui/issues/1399 then :p I had thought legend for the label was probably the best approach too.

PR #1421, opened 2026-03-13
[toolbar] Add disabled attribute and legend support

### LINKED ISSUE #1399: [toolbar] Do we need a HTML way to label them?
For toolbars, it is a MUST requirement to label them if there are multiple on a page, I think it's generally good practice to label them always? Should we support this via the HTML somehow as a first class feature or should we leave this to aria-label? How commonly do toolbars have visible labels that might help the discussion.

lukewarlow:
Adds a disabled attribute and child legend support to toolbar

- first child legend will behave like in a fieldset, for naming the toolbar.
- disabled attribute to disable all children (Except those within the legend)
- UA styles to put toolbar legend on its own line above items.

Fixes #1399 and #1420

### DIFF
--- a/site/src/pages/components/toolbar.explainer.mdx
+++ b/site/src/pages/components/toolbar.explainer.mdx
@@ -43,6 +43,28 @@ Use cases:
 - Toolbars are displayed horizontally by default.
 - Provides correct keyboard navigation behaviour.
 
+```webidl
+[Exposed=Window]
+interface HTMLToolbarElement : HTMLElement {
+  [HTMLConstructor] constructor();
+
+  [CEReactions, Reflect] attribute boolean disabled;
+};
+```
+
+Toolbars will support a disabled attribute, when specified it will cause all descendents that support a disabled state to be disabled;
+with the exception of any descendents within its first `<legend>` element (this matches fieldset).
+
+### Naming
+
+Toolbars situationally need an accessible name. Naming can become particularly important when there are multiple
+toolbars on a page, or in situations where authors provide a visible label for a toolbar, with the intent that it would
+serve as the identifier (name) of the toolbar, programmatically.
+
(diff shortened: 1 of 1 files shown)


lukewarlow:
~~Need to update the styling that will maybe be needed to make this work nicely.~~

For now I've added styles that puts the first child legend as flex basis 100% so it's on its own line above the rest of the items.

site/src/pages/components/toolbar.explainer.mdx
```diff
@@ -43,6 +43,25 @@ Use cases:
...
+  [CEReactions, Reflect] attribute boolean disabled;
+};
+```
+
+Toolbars will support a disabled attribute, when specified it will cause all descendents that support a disabled state to be disabled;
+with the exception of any descendents within its first `<legend>` element (this matches fieldset).
+
+### Naming
+
+Toolbars commonly need an accessible name, this is especially important when there are multiple toolbars on a page.
```
scottaohara: how this sound to you?

```suggestion
Toolbars situationally need an accessible name. Naming can become particularly important when there are multiple toolbars on a page, or in situations where authors provide a visible label for a toolbar, with the intent that it would serve as the identifier (name) of the toolbar, programmatically.
```

site/src/pages/components/toolbar.explainer.mdx
```diff
@@ -43,6 +43,25 @@ Use cases:
...
+
+To facilitate this toolbars will support the `<legend>` element similarly to how optgroup and fieldset do.
```
scottaohara:
```suggestion
To facilitate labeling and naming, toolbars will support the `<legend>` element similarly to its use in `fieldset`, and more recently its allowance within the `optgroup` element.
```

PR #1440, opened 2026-04-29
[focusgroup] Add sections on spatial navigation and modifiers

janewman:
Addresses the follow-up questions in the [W3C TAG design review](https://github.com/w3ctag/design-reviews/issues/1152#issuecomment-4346218804) on focusgroup.

 ### Modifier keys
Adds a note that user agents are expected to suppress focusgroup directional navigation when a modifier key (Ctrl, Alt, Meta, or Shift) is held, preserving platform-specific modifier+arrow behaviors. For other input modalities, user agents are urged to consider if the concept of a modifier exists and if so, use discretion for if this should supress the focusgroup navigation behavior.

### CSS spatial navigation
Clarifies that focusgroup (governing arrow-key movement *within* author-defined composite widgets) and [CSS spatial navigation](https://www.w3.org/TR/css-nav-1/) (governing arrow-key traversal across the document) operate at different scopes and are expected to compose.

I spent some additional time today testing the two features looking for edge cases, and found that depending on how content is organized, focusgroup wrap can stop focus exit via spatial navigation, since both use the arrow keys. This would be the same for existing js implementations of roving tabindex, but thought this interesting enough to call out. I don't think focusgroup itself should change its behavior here. Except for this specific case, I thought the two features worked quite nicely together.


lukewarlow (approved):

lukewarlow:
Given the state of CSS spatial navigation (I *think* it's not actually shipped meaningfully other than in some WebKit deployments) I think that's fine, probably something that can be resolved if it ever comes back. 

Tbh (personal opinion) CSS spatial navigation could maybe be fully subsumed by this feature with the right 2d additions.

janewman:
> Given the state of CSS spatial navigation (I _think_ it's not actually shipped meaningfully other than in some WebKit deployments) I think that's fine, probably something that can be resolved if it ever comes back.

+1 on this point.

> Tbh (personal opinion) CSS spatial navigation could maybe be fully subsumed by this feature with the right 2d additions.

I'm of a different (personal) opinion, and think that these two features should remain separate, with CSS spatial navigation being focused on navigating around the document and between controls in a more user agent + user-controlled way (with author-supplied customization where needed) and focusgroup being used for navigation within controls defined by the 
author. 
As always, I'm open to being convinced 😆

PR #1449, opened 2026-05-21
[Range] Add * to slider-thumb and tick 

brechtDR:
I following this idea for an API this would be logical.


lukewarlow (commented):
Perhaps we should allow the non functional version to target all of them and the functional to target specific ones?

lukewarlow:
There were discussions in the csswg about multiple slider-fills so it might be worth looking at those conversations?

gregwhitworth:
> There were discussions in the csswg about multiple slider-fills so it might be worth looking at those conversations?

@lukewarlow mind linking to these conversations?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -426,14 +457,27 @@ rangegroup::slider-segment(2) {
...
 ```
 
+#### Exposing Thumb and Segment Geometry to CSS
+
+To support layout customization beyond a straight horizontal track, we propose exposing each thumb's position along the track as a CSS custom property on the `::slider-thumb(*)` part (e.g. `--slider-thumb-position`, a percentage or `<number>` along the track), with segments similarly exposing their start and size. This is purely additional information for styling — it does not change hit-testing, keyboard behavior, or value computation, which remain based on the linear `min`/`max` model described above.
+
+Authors can combine this with ordinary CSS (such as `offset-path` and CSS anchor positioning) to lay thumbs out along a custom visual path — for example a curved or circular track — without any new HTML attributes or UA-required geometry. See the [wavy-track](https://brechtdr.github.io/enhanced-range-slider-poc/#wavy-track) and [circular-track](https://brechtdr.github.io/enhanced-range-slider-poc/#circular-track) examples in the proof-of-concept for an illustration of this pattern:
+
+```css
+rangegroup::(slider-thumb(*)) {
```
keithamus:
```suggestion
rangegroup::slider-thumb(*) {
```

?

lukewarlow:
Sorry for the delay: https://github.com/w3c/csswg-drafts/issues/12419

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: Woops. Thank you for spotting this.

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -426,14 +457,27 @@ rangegroup::slider-segment(2) {
...
     width: 24px;
     height: 24px;
     background-color: #2196f3;
     border-radius: 50%;
 }
 ```
 
+#### Exposing Thumb and Segment Geometry to CSS
+
+To support layout customization beyond a straight horizontal track, we propose exposing each thumb's position along the track as a CSS custom property on the `::slider-thumb(*)` part (e.g. `--slider-thumb-position`, a percentage or `<number>` along the track), with segments similarly exposing their start and size. This is purely additional information for styling — it does not change hit-testing, keyboard behavior, or value computation, which remain based on the linear `min`/`max` model described above.
```
lukewarlow: Exposing custom properties by the browser is going to be a non-starter I suspect. I think instead a new function that returns the value is a better starting point. (Alternatively, perhaps we can get it from the CSS forms control-value() function or something)

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -426,14 +457,27 @@ rangegroup::slider-segment(2) {
...
+
+Authors can combine this with ordinary CSS (such as `offset-path` and CSS anchor positioning) to lay thumbs out along a custom visual path — for example a curved or circular track — without any new HTML attributes or UA-required geometry. See the [wavy-track](https://brechtdr.github.io/enhanced-range-slider-poc/#wavy-track) and [circular-track](https://brechtdr.github.io/enhanced-range-slider-poc/#circular-track) examples in the proof-of-concept for an illustration of this pattern:
+
+```css
+rangegroup::slider-thumb(*) {
```
lukewarlow, line 480: Is slider thumb exposed on the container or the range inputs?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -502,7 +546,7 @@ rangegroup::slider-segment(2) {
 rangegroup::slider-segment(3) {
     background-color: #ddd;
 }
-rangegroup input[type="range"]::slider-thumb {
+rangegroup input[type="range"]::slider-thumb(*) {
```
lukewarlow: Same question as above because this is exposing it on the range instead?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -78,12 +97,12 @@ There currently is a running draft solution for this found at: [https://www.w3.o
 
 - `::slider-track`: The main track along which the thumb(s) move.
 - `::slider-fill`: The filled portion of the track, typically between the minimum value and the thumb (or between thumbs for multi-handle ranges).
-- `::slider-thumb`: The draggable handle(s) used to select values.
+- `::slider-thumb(*)`: The draggable handle(s) used to select values.
```
lukewarlow: I think we should keep ::slider-thumb as well and just make it select the first thumb. At least on the range inputs. I think we should keep the thumb singular. On the range group it makes sense to just do the functional version.

brechtDR:
@lukewarlow I updated some of the explainer according to your feedback.

> I think we should keep ::slider-thumb as well and just make it select the first thumb

Here I don't really agree; to be honest, it might be worth opening a discussion. I'd expect it to target all thumbs as well, which would make the `(*)` not needed anymore. For consistency's sake, the `(*)` would still be nice to have... It could even be used for differentiating a single range with a multi-range in a reset:

```
::slider-thumb { background: green; } /* single and multi-sliders are all green */
::slider-thumb(*) { background: orange;} /* now all single sliders are green, but the rangegroup ones are orange */
::slider-thumb(2) { background: red;} /* now all single sliders are green, but the rangegroup ones are orange, the second thumb is red */
```

Not taken up in this explainer, but something that came to mind is if these thumbs are actually considered siblings? This would allow you to do:

```
::slider-thumb(*) {
  --start-hue: 180; /* Teal */
  --hue-range: 120; /* Range from Teal to Magenta */
  
  /* distribute a range of hues across all items using sibling-count and index */
  background-color: oklch(
    65% 0.35
      calc(
        var(--start-hue) + (var(--hue-range) / (sibling-count() - 1)) *
          (sibling-index() - 1)
      )
  );
}
```
(pretty much this idea: https://codepen.io/utilitybend/pen/PwPxePK)

But that might be stretching it, and I'll make an issue of this later on. It should certainly be picked up, but maybe out of scope for this particular PR.

Tagging @gregwhitworth and @keithamus as well to inform. I was a bit delayed.

lukewarlow (approved):
I think some of this probably needs further thought and maybe some details will be changed but approving to keep things moving.

brechtDR:
> I think some of this probably needs further thought and maybe some details will be changed but approving to keep things moving.

@lukewarlow  Absolutely. The way I see it, this is still getting the starting point up and running for the first experimental implementations.

PR #1450, opened 2026-05-26
[focusgroup] Switch to token list, allow behavior tokens to not be the first token

janewman:
Updates the following to match the latest iteration of https://github.com/whatwg/html/pull/11723
* The focusgroup attribute is now a token list instead of a string.
* Focusgroup tokens are allowed to be put in any order, with the first valid behavior applied.


lukewarlow (approved):

janewman:
@lukewarlow thanks for the quick review!

PR #1376, opened 2026-02-20
[switch] Update explainer to reflect resolution in #1282

gfellerph:
With the resolution to support the switch attribute proposal, several changes have been introduced to this explainer:
- Moved to non-active proposals in the navigation
- Updated introductory paragraph with reasoning for the new resolution
- Cleaned up Features, Styleability and Accessibility sections and removed switch element references as well as obsolete content
- Removed open questions as explainer is inactive
- Added a table of contents, author and last updated field
- Multiple small fixes like typos and clarifications

Preview: https://deploy-preview-1376--open-ui.netlify.app/components/switch.explainer/

Issue #313, opened 2021-04-06
[CHECKBOX] "unchecked" behaviour


nuxodin:
I would really like it that when a checkbox is unchecked its value is an empty string.

I don't know about others, but i have to do this hack all the time:
```js
let value = el.value; // other input
if (el.type==='checkbox') { // checkboxes
    value = el.checked ? el.value : '';
}
```

nuxodin:
I also use this "hack" to send an "unchecked" checkbox in a form:

```html
<input type='hidden' value='' name='accepted'>
<input type='checkbox' value='1' name='accepted'>
```

This hack has a rating of 592 at stackoverflow 
https://stackoverflow.com/a/1992745/4865307

And it is mentioned at MDN:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox

I think if we have the possibilities to redesign the checkbox, we should improve it to this behavior.

```html
<checkbox value="1" name=accepted checked> <!-- checkbox.value === '1' -->
<checkbox value="1" name=accepted> <!-- checkbox.value === ''  -->
```

Conceivable would also be a future expansion with an attribute like "falsy".
```html
<checkbox value="on" falsy="off" name=accepted>
```

gregwhitworth:
@nuxodin thank you so much for filing this. This is the exact same issue that we just discussed in #311 - @domenic actually hits on the confusion of the current checkbox in this [comment](https://github.com/openui/open-ui/issues/311#issuecomment-816248592).

Likewise this issue you raised was brought up to the TAG [here](https://github.com/w3ctag/design-principles/issues/279). Ultimately the group seems to be in agreement that there need to be better cohesion in how the initial states relate to that of the IDL values themselves.

As far as your proposals are concerned they need to follow the proposal for `open` in `<popup>` in that we should be clear in what is controlling the initial state vs reflection in the IDL. Effectively the groups seem to be leaning towards making it clear in the name that you're setting the initial state & value rather than that content attribute reflecting the value. @domenic do you want to add any additional thoughts here?

nuxodin:
My final solution would be:

```html
<checkbox on="yes" off="no" name=accepted>
```
```js
checkbox.value // no
checkbox.checked = true;
checkbox.value // yes
```

Advantages: 
- Also unchecked checkboxes sends something (as default an empty string)
- the property "value" no longer incorrectly reflects the attribute "value" (the attribute no longer exists)
- "on" and "off" are short and clear attributes

[...]

PR #1470, opened 2026-07-02
Add Switch Attribute Explainer

andresrega-msft:
Add Switch Attribute Explainer


site/src/pages/components/switch.attribute.explainer.mdx
```diff
@@ -0,0 +1,641 @@
...
+
+The attribute approach (`<input type="checkbox" switch>`) provides the same benefit while remaining backwards compatible with older browsers (renders as a checkbox).
+
+---
+
+## Stakeholder Feedback / Opposition
+
+| Vendor | Position | Notes |
+|---|---|---|
+| WebKit/Safari | **Positive** (shipped) | Available behind flag since Safari 17.4 |
```
annevk: The flag is only for the pseudo-elements. The feature itself has no guard.

annevk:
Also how does this relate to #1376?

lukewarlow (changes requested):
As Anne has pointed out there's an existing explainer. This should redirect from the old explainer and the old one should be moved, deleted or otherwise handled so we don't have two show up. Feel free to link back to the old one from the new one

site/src/pages/components/switch.attribute.explainer.mdx
```diff
@@ -0,0 +1,641 @@
...
+</label>
+```
+
+![Default switch appearance](/images/switch-attribute/default-switch.png)
+
+### Alternative proposal
+
+There is a [switch element proposal at Open UI](https://open-ui.org/components/switch.explainer/) that would introduce a standalone `<switch>` element. This explainer focuses on the attribute approach (`<input type="checkbox" switch>`). See [Alternatives Considered](#alternatives-considered) for a comparison.
+
+An implementation of the attribute approach was already introduced by Safari Technology Preview and Safari Beta (https://github.com/whatwg/html/pull/9546#issuecomment-1865357407).
```
lukewarlow: It's shipped in stable Safari.

site/src/pages/components/switch.attribute.explainer.mdx
```diff
@@ -0,0 +1,641 @@
...
+### Proposed Approach
+
+```html
+<input type="checkbox" switch>
+```
+
+Adding the `switch` attribute to a checkbox changes its appearance to a switch control. The control exposes two pseudo-elements for styling:
+
+- **`::track`** — the background track of the switch
+- **`::thumb`** — the sliding knob/indicator
```
lukewarlow: These are out of sync with what is proposed in css-forms I think we should defer to that spec for what's actually proposed and just write the explainer on the basis of what is proposed there. If we need to make changes we can as the pseudos never shipped.

gregwhitworth:
> As Anne has pointed out there's an existing explainer. This should redirect from the old explainer and the old one should be moved, deleted or otherwise handled so we don't have two show up. Feel free to link back to the old one from the new one

Let's keep and link to it, don't delete it but keep the same URL so that anyone that linked to the prior one it still functions as expected. There's valid research in the prior one that we should retain

site/src/pages/components/switch.attribute.explainer.mdx
andresrega-msft: Resolved, thanks!

site/src/pages/components/switch.attribute.explainer.mdx
andresrega-msft: Updated to use ::slider-track/::slider-thumb from css-forms-1. Also added a note that naming is subject to change as the spec evolves. Thanks!

site/src/pages/components/switch.attribute.explainer.mdx
andresrega-msft: Fixed, thanks!

andresrega-msft:
@annevk @gregwhitworth @lukewarlow The new explainer links to the old one in the "Alternative proposal" section and in "Alternatives Considered." Please let me know if this is insufficient in any way. Also, the old explainer will stay at its current URL, no changes to it in this PR. Thanks!

site/src/pages/components/switch.attribute.explainer.mdx
```diff
@@ -0,0 +1,643 @@
+---
+menu: Active Proposals
```
lukewarlow, line 2: Can you also update the existing explainer so the menu value is changed to "Non-active Proposals". Or alternatively remove it's manu value entirely, so it's only linked to from the new explainer.

Either way it would also be good to link to the new explainer from the old one.

site/src/pages/components/switch.attribute.explainer.mdx
```diff
@@ -0,0 +1,643 @@
...
+
+## Design
+
+### Proposed Approach
+
+```html
+<input type="checkbox" switch>
+```
+
+Adding the `switch` attribute to a checkbox changes its appearance to a switch control. The control exposes pseudo-elements for styling, aligned with the [CSS Forms spec](https://drafts.csswg.org/css-forms-1/):
```
lukewarlow, line 170: I think its worth pointing out that whether these pseudos are always exposed or only when appearance:base is set, is up for debate.

site/src/pages/components/switch.attribute.explainer.mdx
```diff
@@ -0,0 +1,643 @@
...
+</label>
+
+<!-- Disabled -->
+<label>
+  <input type="checkbox" switch disabled>
+  Feature unavailable
+</label>
+```
+
+![Default switch appearance](/images/switch-attribute/default-switch.png)
```
lukewarlow: The word default here perhaps could be better as something else. The base switch is unlikely to look like that exactly, and the auto style is UA dependent.

Fwiw the base styles per the CSS forms spec currently look something like below:

Image

site/src/pages/components/switch.attribute.explainer.mdx
```diff
@@ -0,0 +1,643 @@
...
+  background: linear-gradient(135deg, #f6d365, #fda085);
+  box-shadow: 0 2px 6px rgba(0,0,0,0.3);
+}
+```
+
+![Styled switch with gradient](/images/switch-attribute/styled-switch-gradient.png)
+
+### Form Participation
+
+The switch behaves identically to a checkbox for form submission:
```
lukewarlow, line 224: Just a comment that this is perhaps something where a separate enhancement could be made to the input element to be able to specify a checked value but also an *unchecked value* that's submitted as part of form control. I can't remember if the form behaviour was mentioned in the other explainer but this is something where it's not a great behaviour (imo anyway).

site/src/pages/components/switch.attribute.explainer.mdx
andresrega-msft: Resolved. Changed the old explainer to "Non-active Proposals" and added a note at the top linking to the new attribute explainer. Thanks!

site/src/pages/components/switch.attribute.explainer.mdx
andresrega-msft: Resolved.  Added a note mentioning that whether the pseudos require appearance: base is still under discussion. Thanks!

site/src/pages/components/switch.attribute.explainer.mdx
andresrega-msft: Resolved. Updated the image caption to clarify it's showing Chromium's rendering. Thanks!

site/src/pages/components/switch.attribute.explainer.mdx
andresrega-msft: I agree that could be useful. To me it sounds like a good idea for a separate proposal that would benefit checkboxes generally, not just switches.

gregwhitworth (approved):
Approving but one nit, you have dual counting occurring. Thanks for the work on this

Image

gregwhitworth:
@lukewarlow friendly ping on the review here

site/src/pages/components/switch.attribute.explainer.mdx
lukewarlow: Yeah totally agree. Just wanted to note it down while I remembered.

lukewarlow (approved):

lukewarlow:
LGTM happy to iterate on this.

site/src/pages/components/switch.attribute.explainer.mdx
lukewarlow: https://github.com/openui/open-ui/issues/313 - exists already and already mentions this so we'll continue there.

PR #1475, opened 2026-07-13
Replace navigation menubar focusgroup example

### LINKED ISSUE #1474: [focusgroup] Remove navigation menu example from explainer
We should remove the navigation menu [example](https://open-ui.org/components/scoped-focusgroup.explainer/#:~:text=is%20creating%20a-,navigation%20menubar.,-Both) from the focusgroup explainer. This pattern is contentious and not relevant to the key part trying to be explained. Instead we should replace it with a more conventional menu pattern.

This point came up in a [blog post](https://adrianroselli.com/2026/07/focusgroup-tests.html#Opinions:~:text=I%E2%80%99m%20frustrated%20the%20menu%20example%20refers%20to%20web%20site%20navigation) about the feature and I agree that this example should be changed.

lukewarlow:
Fixes #1474 

This also fixes the fact that the example used to have `<a ... popovertarget>` which isn't correct.

### DIFF
--- a/site/src/pages/components/scoped-focusgroup.explainer.mdx
+++ b/site/src/pages/components/scoped-focusgroup.explainer.mdx
@@ -235,32 +235,35 @@ What to notice:
    change would update `aria-selected` values, the `hidden` state of the controlled `role="tabpanel"`,
    and move the `focusgroupstart` attribute to the newly selected tab.
 
-In the following example, the author is creating a
-[navigation menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/).
-Both `menuitem`s in the `menubar` ("About" and "Admissions") have popover `menu`s. The "Admissions"
-menu has an additional submenu under "Tuition".
+In the following example, the author is creating an
+[editor menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/).
+Both `menuitem`s in the `menubar` ("Font" and "Size") have popover `menu`s.
 
 ```html
-<ul focusgroup="menubar" aria-label="Mythical University">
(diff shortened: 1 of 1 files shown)


site/src/pages/components/scoped-focusgroup.explainer.mdx
```diff
@@ -235,32 +235,35 @@ What to notice:
...
-Both `menuitem`s in the `menubar` ("About" and "Admissions") have popover `menu`s. The "Admissions"
-menu has an additional submenu under "Tuition".
+In the following example, the author is creating an
+[editor menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/).
+Both `menuitem`s in the `menubar` ("Font" and "Size") have popover `menu`s.
 
 ```html
-<ul focusgroup="menubar" aria-label="Mythical University">
+<ul focusgroup="menubar" aria-label="Text Formatting">
   <li role="none">
```
lukewarlow, line 244: Does it even make sense to use `<ul>` and `<li>` all through here. Doesn't that negate loads of the automatic role inheritance?

For example the ul in this demo at the top level actually sticks with a role of list *not* menubar.

lukewarlow: Follow up [issue](https://github.com/openui/open-ui/issues/1476) to address that part: I think it's fine for this PR to keep things as they are (or if the other issue is fixed first then this can be updated to match)

janewman: +1 that this is not getting the automatic role inheritance, thanks for filing the issue to track this.

janewman (approved):
lgtm! Thanks for jumping in and addressing these issues!!

keithamus (approved):

PR #1481, opened 2026-07-20
Add "Slotted Options for customizable <select>" explainer

ffiori:
Add "Slotted Options for customizable `<select>`" explainer.

There is a previous first version of this explainer in [MSEdgeExplainers](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/SlottedOption/explainer.md) but moving it here for broader audience and because [customizable select is here](https://open-ui.org/components/customizable-select.explainer/). I'll set the MSEdge explainer as inactive and add a reference to this one when this is merged here.

### DIFF
--- a/site/src/pages/components/select.slotted.option.mdx
+++ b/site/src/pages/components/select.slotted.option.mdx
@@ -1,0 +1,500 @@
+---
+menu: Active Proposals
+name: Slotted Options in Customizable `<select>` (Explainer)
+layout: ../../layouts/ComponentLayout.astro
+authors:
+  - name: Fernando Fiori
+    url: https://github.com/ffiori
+---
+
+## Participate
+
+- WHATWG HTML issue:
+  [whatwg/html#11535](https://github.com/whatwg/html/issues/11535)
+- File feedback on this explainer:
+  [OpenUI issues](https://github.com/openui/open-ui/issues)
+
+## Status of this Document
+
+This document is a starting point for engaging the community and standards
+bodies in developing collaborative solutions fit for standardization. As the
+solutions to the problems described in this document progress along the
+standards track, we will retain this document as an archive and use this section
+to keep the community up to date.
+
+- This document status: **Active**
+- Expected venue: [WHATWG HTML](https://html.spec.whatwg.org/)
+- Current version: this document
+- Previous version: [MSEdgeExplainers/SlottedOption](https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/SlottedOption/explainer.md)
+
+## Table of Contents
+
+- [Introduction](#introduction)
+- [User-Facing Problem](#user-facing-problem)
+  - [Goals](#goals)
+  - [Non-goals](#non-goals)
+- [Motivation](#motivation)
+- [Proposed Approach](#proposed-approach)
+- [Open questions](#open-questions)
+- [Alternatives Considered](#alternatives-considered)
+- [Accessibility, Internationalization, Privacy, and Security Considerations](#accessibility-internationalization-privacy-and-security-considerations)
+- [Stakeholder Feedback / Opposition](#stakeholder-feedback--opposition)
+- [References](#references)
+
+## Introduction
+
+The `<select>` element is one of the most common form controls on the web. With
+the new customizable `<select>` (`appearance: base-select`), developers can
+finally style and structure a real `<select>` to match their design.
+
+Today, you can't build a reusable `<select>`-based component that lets the
+people using your component supply the `<option>`s. If you wrap a `<select>`
+inside a
+[web component](https://developer.mozilla.org/en-US/docs/Web/API/Web_components)
+and expose a
+[`<slot>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) for
+options, the `<select>` ignores the slotted options and its dropdown comes up
+empty.
+
+This explainer proposes that a customizable `<select>` recognize `<option>`,
+`<optgroup>`, and its trigger `<button>` when they are slotted in from outside
+the component, so they behave as if they were written directly inside the
+`<select>`. This lets developers wrap the native `<select>` in a component
+instead of rebuilding it from scratch.
+
+## User-Facing Problem
+
+Say you're building a design system and want to ship a styled select that
+consumers use like the native one, writing `<option>`s as children. Design
+systems are built as web components, so you reach for the obvious approach: put
+a real `<select>` in the component's shadow root and expose a `<slot>` for the
+options:
+
+```html
+<my-custom-select>
+  <template shadowrootmode="open">
+    <select>
+      <slot></slot>
+    </select>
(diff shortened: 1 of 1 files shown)


site/src/pages/components/select.slotted.option.mdx
```diff
@@ -0,0 +1,500 @@
+---
+menu: Active Proposals
+name: Slotted Options in Customizable `<select>` (Explainer)
```
lukewarlow: I would drop the customizable part here. This might not be specific to appearance: base.

site/src/pages/components/select.slotted.option.mdx
```diff
@@ -0,0 +1,500 @@
...
+Today, you can't build a reusable `<select>`-based component that lets the
+people using your component supply the `<option>`s. If you wrap a `<select>`
+inside a
+[web component](https://developer.mozilla.org/en-US/docs/Web/API/Web_components)
+and expose a
+[`<slot>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) for
+options, the `<select>` ignores the slotted options and its dropdown comes up
+empty.
+
+This explainer proposes that a customizable `<select>` recognize `<option>`,
```
lukewarlow: Again I don't think it's tenable to limit this to just appearance base. So I would drop the customizable.

Edit: I've seen the comment below that it's an open question so feel free to disregard this but leaving it anyway.

site/src/pages/components/select.slotted.option.mdx
```diff
@@ -0,0 +1,500 @@
...
+entirely: no `MutationObserver` to sync options, no second copy to keep in step,
+and no rebuilding the control from scratch. Developers keep a real `<select>`
+and let consumers pass `<option>`s the way they already do, so selection,
+accessibility, and form behavior come from the platform instead of being
+reimplemented and maintained by hand. Easier, correct authoring helps users too:
+developers reach for the accessible native control instead of a fragile
+substitute.
+
+> **Note on scope:** This initially targets `<select>` in customizable mode
+> (`appearance: base-select`). Whether to extend it to the default native
```
lukewarlow: Okay I take this and maybe disregard my comments above but I strongly suspect this will not fly. Changing slotting behaviour based on the appearance value is something that has come up and is not desirable from what I understand.

site/src/pages/components/select.slotted.option.mdx
```diff
@@ -0,0 +1,500 @@
...
+  with the changes observable immediately.
+
+### Non-goals
+
+- Extending this behavior to other elements in this proposal (for example
+  `<datalist>`, `<table>`, `<fieldset>`, or `<form>`). Those may be worth doing
+  later, but are out of scope here.
+- Letting arbitrary unrelated elements become a select's options. Only the
+  elements a `<select>` already accepts (`<option>`, `<optgroup>`, and the
+  trigger `<button>`) are in scope.
```
lukewarlow: What about `<hr>`?

lukewarlow (approved):

site/src/pages/components/select.slotted.option.mdx
```diff
@@ -0,0 +1,500 @@
...
+  them; copying them leaves the select using clones that can drift from the
+  originals (attributes, state, and event listeners set in JavaScript don't
+  carry over). The sync also runs asynchronously, so options can show up late or
+  flicker, with potential performance issues.
+- **Rebuild the control by hand.** Drop the native `<select>` and re-create the
+  trigger, popup, option list, keyboard support, focus handling, and
+  screen-reader semantics with custom elements and JavaScript.
+
+Design systems already ship these workarounds; see Shopify's
+[Polaris web components](https://shopify.dev/docs/api/app-home/polaris-web-components/forms/select)
```
frehner, line 115: Can confirm that we (polaris) hate shipping this. 

Note that we also use a custom element `<s-option>` around the native `<option>` element -- which in our case is a necessary wrapper, but at first glance doesn't do anything (and the host element is display:contents). 

[edit] glad to see you actually address that below. Thanks!

site/src/pages/components/select.slotted.option.mdx
```diff
@@ -0,0 +1,500 @@
...
+  before reaching the `<select>`.
+- Recognize options that reach the `<select>` through its flat tree, whether via
+  a `<slot>` or a descendant's shadow tree.
+- Keep the option list in sync as options are added, removed, or re-slotted,
+  with the changes observable immediately.
+
+### Non-goals
+
+- Extending this behavior to other elements in this proposal (for example
+  `<datalist>`, `<table>`, `<fieldset>`, or `<form>`). Those may be worth doing
```
frehner, line 158: a LOT of interest on our part for these future expansions.

especially form and table.

lukewarlow: Do these materially add much extra work? I think it would be nice to have a coherent answer for html elements in general, else we'll have a continuous game of whackamole.

site/src/pages/components/select.slotted.option.mdx
ffiori: Oops, good catch, adding it.

site/src/pages/components/select.slotted.option.mdx
ffiori: I agree that we need a coherent direction. But these elements also add distinct work with fieldset propagation, form ownership/submission, and table parsing and collections, which I would say would be out of scope for this proposal. I added a section framing `<select>` as the first concrete case rather than a permanent special case. Let me know your thoughts.

site/src/pages/components/select.slotted.option.mdx
ffiori: Thanks for confirming this! The `<s-option>` pattern is exactly the nested shadow tree case in the explainer. I also called out Polaris in the stakeholder feedback section now.

site/src/pages/components/select.slotted.option.mdx
ffiori: Agreed, I dropped ‘Customizable’ from the title so the proposal isn’t inherently tied to `appearance: base-select`.

site/src/pages/components/select.slotted.option.mdx
ffiori: That's reasonable, I made the proposal appearance-independent now. But before committing to it for default `<select>` I think we need evidence that this is web compatible, probably with use counters. This also came up in an earlier review thread https://github.com/MicrosoftEdge/MSEdgeExplainers/pull/1357#discussion_r3537925396

I clarified that appearance-independent behavior is preferred while keeping default-appearance support open depending on the evidence we can get.

Open to more discussion on this front.

site/src/pages/components/select.slotted.option.mdx
ffiori: Addressing this in the other comment.

lukewarlow (approved):

PR #1485, opened 2026-07-28
[menu.explainer] Add submenu to menulist content model

davidsgrogan:


davidsgrogan:
@dbaron I think this was a simple omission, but please correct me if it was intentional.

dbaron (approved):

PR #1493, opened 2026-08-06
Mark richer text fields proposal as non-active

### LINKED ISSUE #1492: Mark Richer Text Fields proposal as inactive
The active Open UI proposal Richer Text Fields overlaps with a new feature that has landed in Chrome Beta called Opaque Range.

Webkit have declared a [positive position](https://github.com/WebKit/standards-positions/issues/541) on the opaque range API, as have [Mozilla](https://github.com/mozilla/standards-positions/issues/1289).

AFAIK all the functionality of Richer Text Fields is covered by opaque range, so I would suggest that the Richer Text Fields explainer be moved to the non-active section. Alternatively, if there are elements of the Richer Text Fields proposal that are not met by the new API, I would suggest that the explainer be rewritten.

lukewarlow:
This is mostly replaced by Microsoft's OpaqueRange API.

Fixes https://github.com/openui/open-ui/issues/1492

### DIFF
--- a/site/src/pages/components/richer-text-fields.explainer.mdx
+++ b/site/src/pages/components/richer-text-fields.explainer.mdx
@@ -1,11 +1,13 @@
 ---
-menu: Active Proposals
+menu: Non-active Proposals
 name: Richer Text Fields (Explainer)
 layout: ../../layouts/ComponentLayout.astro
 authors:
   - name: "@keithamus"
     url: https://github.com/keithamus
 ---
+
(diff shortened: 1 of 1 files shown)


lukewarlow:
cc @keithamus feel free to follow-up with changes if there's some part of this proposal you think we still need (I think some towards the end isn't addressed by OpaqueRange)

keithamus:
Yeah I don't think this is non-active, it is just that a portion of it has been implemented (or "graduated").

lukewarlow:
I guess it's non-active in that it's not being worked on right now? Unless you are? Feel free to make a different PR to update the explainer and leave it as active if you'd prefer.

Issue #1489, opened 2026-07-29
[focusgroupV2]  Bikeshed names for the focusgroup itemcontrols modifiers


janewman:
Focusgroup V2 proposes `itemcontrols`, with `noitemcontrols` as its negative counterpart. [Review feedback on PR #1488](https://github.com/openui/open-ui/pull/1488#discussion_r3671214555) noted that `itemcontrols` does not clearly communicate its purpose, on which I agree.

With `itemcontrols`, focusable controls with `focusgroup="none"` that descend from any non-active focusgroup item are not considered for sequential focus navigation.

```html
<div focusgroup="toolbar itemcontrols" aria-label="Documents">
  <span tabindex="0">
    Quarterly report
    <button focusgroup="none">Rename</button>
    <button focusgroup="none">Delete</button>
  </span>

  <span tabindex="0">
    Project plan
    <button focusgroup="none">Rename</button>
    <button focusgroup="none">Delete</button>
  </span>
</div>

<button>Upload document</button>
```

* If focus is on Quarterly report and the user presses Tab, focus moves to its Rename button.
* Pressing Tab again moves focus to its Delete button.
* Pressing Tab again skips the inactive Project plan item and its controls, moving focus to Upload document.
* Directional navigation moves between Quarterly report and Project plan (no change from a focusgroup without this modifier).
* If focus moves to Project plan, its controls become the available through sequential focus navigation.
* Pressing Shift+Tab from "Upload document" restores the previously focused **focusgroup item**, skipping over any nested controls, as they are only considered "active" when focus is on or within them.

sb3nder:
In the following case, if the focus is on Help and you press Tab, where does focus go?
```html
<div focusgroup="toolbar itemcontrols" aria-label="Documents">
  <span tabindex="0">
    Quarterly report
    <div focusgroup="none">
      <button>Rename</button>
      <button>Delete</button>
    </div>
  </span>

  <button focusgroup="none">Help</button>

  <span tabindex="0">
    Project plan
    <div focusgroup="none">
      <button>Rename</button>
      <button>Delete</button>
    </div>
  </span>
</div>

<button>Upload document</button>
```

PR #1488, opened 2026-07-28
[focusgroupV2] Add the V2 proposal and move future considerations out of the original proposal.

janewman:
### This PR separates the current focusgroup proposal from a new effort to add 2D grid and related functionality:
- Keeps the existing explainer focused on the current linear focusgroup design.
- Adds a new Focusgroup V2 explainer covering `itemcontrols`, feed behavior, and two-dimensional grid navigation.
- Moves V2 tokens, algorithms, examples, guidance, and open questions into the new proposal.
- Cross-links both explainers while keeping shared focusgroup behavior defined only in the existing explainer.

### tldr on V2:
* Modifier `itemcontrols` allows authors to "hide" nested focusable elements in inactive focusgroup items, ensuring tab doesn't enter nested content except for the active focusgroup item.
* New _linear_(**1D**) behavior feed which implies itemcontrols allowing for "feeds" of content with nested controls. (Discord chat threads are a very close equivalent of something that does this today, where arrow keys allow you to go from message to message, and tab lets you enter the nested controls. Many social media or communication sites implement a similar pattern.
* New **2D** behavior grid which also implies itemcontrols and allows for 2D navigation of the nested focusables.
  * New **2D-only** modifier manual  switches the grid to be in manual mode, requiring the author to specify rows via the new attribute focusgrouprow. (without this, grid requires use of `<table>` to infer the intended structure.
  * New **2D-only** modifier `flow` makes it so moving past the end of a row "flows" focus to the first item in the next column, and vice-versa.
  * New **2D-only** modifiers `rowflow` and `colflow` which apply flow only in the specified "direction"
  * ⁠New **2D-only** modifiers `rowwrap` and `colwrap` which perform wrapping only in the specified "direction"

### Related issues:
* #1390
* itemcontrols should address some of the issues raised in #1435
* https://github.com/openui/open-ui/issues/1489

Issue #1471, opened 2026-07-03
[focusgroup] Up/down arrow key navigation with inline wrapping content


AshleyScirra:
It would be nice if `focusgroup` could handle inline wrapping content with up/down arrow key navigation. In this case the focusable elements are using inline layout and may wrap across lines. This means they are logically (according to layout) one-dimensional, but visually two-dimensional. Note I don't mean the wrapping mode where moving off the end moves back to the start; I mean wrapping with the content.

To illustrate the point, here is a screenshot of an "icon view" control from [Construct 3](https://editor.construct.net/):

<img width="974" height="654" alt="Image" src="https://github.com/user-attachments/assets/0fa56c43-5833-4c92-8725-f1f6e192ea87" />

Within each group (e.g. "Data & Storage"), items use inline layout, flowing from left to right and wrapping down lines. For example the "Dictionary" item follows "CSV" in DOM order, but the browser layout has wrapped it to the next line.

Note this is *not* a grid layout: the wrapping varies depending on the size of the container. Using a grid would mean having to resort to JavaScript to perform layout or mark the grid cells by inspecting the layout the browser produced (and then know to update that whenever the container changes).

Using `focusgroup` on this kind of content works fine with left/right arrow keys. However the user will expect the up/down arrow keys to work too. For example with the 'Clipboard' item focused, pressing down should move the focus to the 'JSON' item, because *visually* it is below, even though logically it is after it ("to the right") in inline layout. So at the moment we still have to use JavaScript to handle keyboard navigation with up/down arrow keys, which means to some extent reimplementing what the browser already does with the left/right keys.

This might need to be a new mode for backwards compatibility reasons, but I'd argue it would be useful to handle up/down arrow keys by default with the "inline" modifier: content that does not wrap is not affected, and content that does wrap gets additional useful shortcuts.

Issue #1434, opened 2026-04-17
[focusgroup] why `tabindex=-1` elements behave differently from "focusgroup items"?


sb3nder:
For [example](https://codepen.io/sb3nder/pen/WbGOpgJ):
```html
<button>fries</button>
<div focusgroup="tablist">
	<button>cake</button>
	<button tabindex="-1">ice cream</button>
	<button focusgroupstart>bread</button>
</div>
<button>rice</button>
```
For normal FG items, if you focus "cake", then press Tab, focus goes to "rice".

For `tabindex=-1` elements, if you focus "ice cream", then press Tab, focus goes to "bread",
and if you focus "ice cream", then press Shift+Tab, focus goes to "fries".

I understand why it behaves like this from reading the HTML PR, as it says that `tabindex="-1"` elements are not *"focus group items"* and if I follow the algorithm, yada yada... 

What I don't know is ***why*** they behave like this, and consequently, then I guess why they aren't "focus group items."
Do you have any examples? Or do you know of an ARIA APG where this is clear?

~~(For context: I'm asking because `tabindex="-1"` is the only way to work around segments; I wouldn't have asked otherwise.)~~

cc @janewman

sb3nder:
One strange thing is that `tabindex="-1"` elements aren’t "focusgroup items," yet to disable directional navigation, you use `focusgroup="none"`.
`<button tabindex="-1" focusgroup="none">cake</button>`

EDIT: Strange in terms of naming.

janewman:
Sorry for my late reply, this [chromium behavior](https://chromium-review.googlesource.com/c/chromium/src/+/7486357) came out of the other issue you raised: https://github.com/openui/open-ui/issues/1358 and subsequent discussions.

Focusgroup doesn't consider elements with a negative tabindex as participating, and this keeps the internal logic of "only elements with a positive tabindex are keyboard focusable" paradigm, unless the element already has focus (via mouse or programatic). When focused, we temporarily treat the element as if it is a focusgroup, allowing directional input to move focus to its neighbors within the focusgroup and sequantial navigation (tab/shift+tab) to move focus following focusgroup item rules.

That all said, one part of your example seems like a bug to me: Hitting tab when the focus is on "ice cream" be equivalent to the behavior you would get if it was a focusgroup item, and should move focus to "rice" instead. (or fries when moving backwards, regardless of the location of focusgroupstart or if focusgroupstart is present.)

sb3nder:
[...]

PR #1494, opened 2026-08-10
[focusgroupV2] Refine the focusgroupv2 explainer

janewman:
- Defines `itemcontrols` as a filter over sequential focus navigation: controls associated with the active item remain available, while controls associated with inactive items are removed.
- Defines how opted-out controls and independent nested focusgroups associate with containing items without flattening or combining focusgroup scopes.
- Makes entry, active-item state, memory, mutation handling, and nested traversal deterministic.
- Defines strict grid topology and validation, including one cell target per coordinate, rectangular and spanless structure, invalid-state fallback, and author diagnostics.
- Clarifies multi-control cells, negative-`tabindex` sources, native key conflicts, wrapping and flow, role inference, and manual row enrollment.
- Adds practical Feed and Grid examples, authoring guidance, accessibility considerations, alternatives, and open questions.
- Reorganizes and shortens the explainer so author-facing behavior appears before implementation detail.

This is follow-up to improve readability, clarity, and correctness. 
It does not add new tokens or concepts beyond the V2 proposal in #1488.

### Key behaviors:

* HTML and V1 determine sequential focus navigation order. `itemcontrols` only removes nested controls associated with inactive items; it does not add, promote, or reorder focusable areas.
* Independent nested focusgroups retain their own directional navigation, entry, memory, and role inference.
* A valid grid has rectangular, spanless topology and exactly one eligible cell target per participating cell.
* Tab and Shift+Tab continue to use HTML sequential focus navigation; grid coordinates affect directional navigation only.

### Related issues:

* #1489
* #1434
* #1390
* #1471


gregwhitworth:
@janewman forced a redeploy so the preview would show up correctly; I personally prefer to read the doc there and then provide comments in Github :)

Thanks for the update

site/src/pages/components/focusgroup-v2.explainer.mdx
```diff
@@ -15,57 +15,40 @@ Discussion happens on the [Open UI focusgroup issues](https://github.com/openui/
...
 | [`noitemcontrols`](#itemcontrols) | Modifier | Disables explicit or default `itemcontrols` behavior. |
 | [`feed`](#feed-focusgroups) | Behavior | Block-axis linear navigation for ARIA-feed content streams; defaults to `itemcontrols`. |
 | [`grid`](#grid-focusgroups) | Behavior | Two-dimensional directional navigation over a native HTML table; defaults to `itemcontrols`. Pair with [`manual`](#manual-row-enrollment) for generic markup. |
-| [`manual`](#manual-row-enrollment) | Grid modifier | Used as `focusgroup="grid manual"`. Enrolls direct child rows with `focusgrouprow` instead of discovering rows from a native table. |
-| [`flow`, `rowwrap`, `colwrap`, `rowflow`, `colflow`](#grid-wrapping-and-flow) | Grid modifiers | Configure per-axis grid edges: wrap within a row or column, or flow into the adjacent row or column. |
+| [`manual`](#manual-row-enrollment) | `grid` modifier | Used as `focusgroup="grid manual"`. Enrolls direct child rows with `focusgrouprow` instead of discovering rows from a native table. |
```
gregwhitworth, line 48: Can't this just be inferred or are there substantial perf benefits to us having them declare this so we don't have to walk the tree at all to get this information?

site/src/pages/components/focusgroup-v2.explainer.mdx
```diff
@@ -15,57 +15,40 @@ Discussion happens on the [Open UI focusgroup issues](https://github.com/openui/
...
+| [`flow`, `rowwrap`, `colwrap`, `rowflow`, `colflow`](#grid-wrapping-and-flow) | `grid` modifiers | Configure per-axis grid edges: wrap within a row or column, or flow into the adjacent row or column. |
```
gregwhitworth, line 49: Possibly not really relevant but I'm curious, how will this work with ATs? For example if I'm in NVDA and I press 
`Control + Alt + > (arrow key)` to move to the next cell and then I press just the `>` key what will happen?

site/src/pages/components/focusgroup-v2.explainer.mdx
gregwhitworth: I think this answers my question in that it does: https://deploy-preview-1494--open-ui.netlify.app/components/focusgroup-v2.explainer/#automatic-table-topology

site/src/pages/components/focusgroup-v2.explainer.mdx
```diff
@@ -15,57 +15,40 @@ Discussion happens on the [Open UI focusgroup issues](https://github.com/openui/
...
+
+The `grid` behavior also specializes V1's existing `wrap` and `nowrap` modifiers: `wrap` expands to both grid axes, while `nowrap` forces hard edges on both.
```
gregwhitworth, line 51: I should have possibly caught this in V1; but I don't see a definition of "edges". Do you have that somewhere that you can link to?

site/src/pages/components/focusgroup-v2.explainer.mdx
```diff
@@ -483,90 +420,106 @@ When feed items change, the focusgroup updates its item set before the next dire
...
 
-<p class="question">Should a future target-selection mode support multiple eligible descendants in a nonfocusable cell, and if so, how can it avoid ambiguous or unstable target choice?</p>
+<p class="question">**Resolved in part:** Native key-conflict cell targets use HTML sequential focus navigation for Tab escape; `grid` coordinates do not redefine that order. Does focus inside nested controls require additional grid-specific escape behavior?</p>
 
-<p class="question">How should `treegrid` extend grid topology, expansion state, and directional key handling without conflicting with the initial rectangular grid model?</p>
+### grid
 
-<p class="question">Does `itemcontrols` need grid-specific entry, memory, or escape behavior beyond the focusable-cell-boundary model?</p>
+<p class="question">How should a future grid extension represent ragged rows, row and column spans, and indexed topology? Should a future mode skip zero-target coordinates, and if so, how should it preserve rectangularity, block-axis columns, boundary operations, and memory?</p>
```
gregwhitworth: Candidly, this seems like the only approach that would make sense. Fill in coordinate space that doesn't exist to make it so that the V2 grid makes sense so you're still traversing a 2D grid that has gaps. This feels akin to how CSS grid creates implicit grid items if none are provided in order to do that layout math of what "is" available. Worth discussing in an issue though.

site/src/pages/components/focusgroup-v2.explainer.mdx
```diff
@@ -483,90 +420,106 @@ When feed items change, the focusgroup updates its item set before the next dire
...
+
+### Require a focusable boundary for every grid cell
+
+Requiring a focusable cell boundary would simplify target selection and remains the recommended pattern for complex cells. V2 also permits one eligible descendant when a single primary control can represent the coordinate without an extra focus stop.
+
+### Derive focus topology from authored ARIA
+
+Authored `grid`, `row`, `gridcell`, and index properties could describe topology, including virtualized content. V2 does not use them as behavior configuration because ARIA describes accessibility semantics and should not silently change focus behavior. The initial model derives topology from HTML structure and leaves indexed or split-grid topology open.
 
 ## Open questions
```
gregwhitworth, line 464: Do you have open issues for these? If so - let's link them accordingly.

gregwhitworth (approved):
A few questions but no major issues on the changes. Approved.

site/src/pages/components/focusgroup-v2.explainer.mdx
janewman: Screen reader focus will depend on what mode they are in, just like "regular" focusgroups or really any author-written controls/widgets. If the screen reader is in a mode where they drive the focus, then focusgroup logic won't actually be called, as the screen reader is capturing the arrow keys and never sending them to the browser. If the screen reader is in the mode where they delegate focus movement to the browser, (and listen for focus events to follow focus movement) then this would work just like it would for any keyboard user, and the focus events generated will ensure the screen reader's focus stays up to date.

For reference, narrator calls this "scan mode" https://support.microsoft.com/en-us/accessibility/windows/narrator/chapter-3-using-scan-mode
NVDA calls this browse mode, while JAWS calls the opposite (where it follows system focus) forms mode.

At the end of the day, Screen readers work reasonably well with these patterns today, and use the informatinon provided to them to determine what role is appropriate for the situation (and consider user patterns). The screen reader doesn't know or care if the focus movement logic is implemented in JS or natively in the browser, but what focusgroup does hope to do is help provide a more consistent experience, and nudge authors towards best proctices.

site/src/pages/components/focusgroup-v2.explainer.mdx
janewman: Not for perf, but to allow authors to use grid focusgroup for situations where the markup doesn't quite line up 1:1 to a table/grid. If they can specify the rows, and have the right number of cells, focusgroup can still work.

That said, if you have a table, we can infer the rest (automatic, not manual), but if all you have is a bunch of divs, by specifying the rows manually, focusgroup can figure out the cells, ensuring that more complex markup can be migrated.

I'm definitely open to suggestions to how this should/shouldn't work.

site/src/pages/components/focusgroup-v2.explainer.mdx
janewman: definition added!
These are just the "sides" of the grid.

site/src/pages/components/focusgroup-v2.explainer.mdx
janewman: cleaned up the list to what I think is actually worth discussing and added links.

site/src/pages/components/focusgroup-v2.explainer.mdx
janewman: #1496 tracks what we can/can't infer.

site/src/pages/components/focusgroup-v2.explainer.mdx
janewman: filed #1495 to track this discussion.

PR #1503, opened 2026-08-13
Simplify and update overscroll explainer

vmpstr:


site/src/pages/components/overscroll-actions.explainer.mdx
```diff
@@ -144,6 +155,16 @@ logic or haptics), we expose the following events on the host container:
 | `overscrollend` | Fired when the gesture completes and the state has changed. |
 | `overscrollcancel` | Fired when the gesture ends but snaps back to the original state. |
 
+### Backdrop
+
+When the overscroll area is visible, the overscroll area has an
+`:overscroll-backdrop` pseudo element that is styled by default with a
+semi-transparent cover, similar to a dialog `:backdrop`. This includes a
```
mfreed7: nit: `::backdrop`

site/src/pages/components/overscroll-actions.explainer.mdx
```diff
@@ -176,96 +197,6 @@ with a side menu that can be opened with a swipe" />
...
-    Sender: Hello world!
-  </div>
-  <menubar id="message-actions">
-    <menuitem>Archive</menuitem>
-    <menuitem>Delete</menuitem>
-  </menubar>
-</div>
-```
-
-### Swipe to dismiss
```
mfreed7, line 209: Can this one not stay, perhaps with the addition of an "X" in the corner to be the button?

site/src/pages/components/overscroll-actions.explainer.mdx
```diff
@@ -285,82 +216,15 @@ hit-testing and painting order:
...
-For the popover case, the container is more generic, and the end boundaries
-are unclear. Closing them the moment focus shifts outside the popover
-would cause frustrating, accidental dismissals for keyboard users just trying
-to tab through their contents. For overscroll areas, however, the role is usually
-better defined, and the end boundaries more clear. When
-a user intentionally moves their focus out of these areas and back to the
-main page, it is a clear signal they are done, making `blur` a safe and helpful
-trigger to automatically close them.
-
-## Modalness
```
mfreed7, line 332: So it seems like actually ALL of the remaining demos (the menu) are modal, is that right? Perhaps some part of the modality conversation should be left in?

Also this removed the conversation about focus loss, which I see is mostly moot if the menu is modal. But perhaps good to be explicit that focus loss isn't a light dismiss signal? (It's still possible, by focusing out of the web content into the browser UI.)

mfreed7 (approved):
LGTM! Just some small stuff.

site/src/pages/components/overscroll-actions.explainer.mdx
vmpstr: So it's still not trivial in that you have to set up a blank overscroll area in order to get the dismiss behavior. I'm hoping that this can be an emergent pattern that we can then support with something like a `dismissable` attribute directly. So I would prefer to leave this out.

site/src/pages/components/overscroll-actions.explainer.mdx
vmpstr: I've added more to the inertness section, including a couple of issues.

sb3nder: If all case for now are modal, would it make sense to start support only for the `<dialog>` element?

PR #1454, opened 2026-06-04
Normalise all explainer frontmatter

keithamus:
Many of our explainers were a little inconsistent in their formatting

- Some didn't end in `.explainer.mdx` - one has `-explainer.mdx`, another just `.mdx`.
- Some had manually written tables of content, others had generated via a magic string.
- Some were missing the author
- Some were missing the last update date.
- Some, while being more recently updated, forgot to manually update the last-updated date.
- None had a creation date.

Graduated proposals were even more spottier, kind of ad-hoc adding stuff like WHATWG issues/PRs, mdn docs and such.

This PR does a sweep of all explainers, pushing all of this into front matter. The build script then adds some intelligence via the git log: reading the git log for last updated, created, authors, and adding them.

This also creates a Table of Contents component which all explainers have added to them.

PR #1506, opened 2026-08-18
Tidy up frontmatter requirements for graduated explainers

keithamus:
In https://github.com/openui/open-ui/pull/1454 we normalised frontmatter, requiring all proposals to include an author, and graduated proposals to include `whatwg_issue`, `whatwg_pr`, `specification` and `mdn` keys.

After the last telecon agenda we realised a few things:

1. Interest Invokers are effectively graduated, despite the spec PR not landing (all new work is being done in WHATWG, so there's little need for us to incubate the proposal any more).
2. Fully graduated proposals linking to their initial spec PRs might be confusing, and should probably only link to the live spec. An example of this is popover=hint which had significant spec rework, and the original PR is now out of date, and might give people wrong information about the semantics of the feature. 
3. Slightly theoretical at this point, but graduated proposals don't necessarily end up in whatwg, but could end up in CSSWG.

This PR encodes these principles in the validation routine such that:

1. Graduated proposals require one of `whatwg_issue`/`whatwg_pr`/`csswg_issue`/`specification`.
2. Graduated proposals with a `specification` field should not include the  `whatwg_issue`/`whatwg_pr`/`csswg_issue` fields.
3. Graduated proposals always require an `mdn` field.

This also cleans up all the existing graduated proposals, so they pass validation.

### DIFF
--- a/site/src/pages/components/accordion.explainer.mdx
+++ b/site/src/pages/components/accordion.explainer.mdx
@@ -7,7 +7,6 @@ layout: ../../layouts/ComponentLayout.astro
 authors:
   - name: "@dbaron"
     url: https://github.com/dbaron
-whatwg_pr: https://github.com/whatwg/html/pull/9400
 specification: https://html.spec.whatwg.org/multipage/interactive-elements.html#attr-details-name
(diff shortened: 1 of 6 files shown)


lukewarlow (approved):

mfreed7 (commented):
LGTM!

Issue #1460, opened 2026-06-15
Segment dragging on range inputs


brechtDR:
For `<rangegroup>` the following questions was asked:

What could be done to enable segment dragging? At the moment, according to the explainer, a click on the track will move the nearest thumb.

The example given was the [iOS alarm clock](https://support.apple.com/en-mz/guide/iphone/iphf2a780f81/ios).

The behavior is that you can select two times and drag the whole segment, moving the two thumbs together.
It might be a thing we could do with a JS api. I'm not sure it should be the default. I also wonder about accessibility implications for this. I'd like us to at least think about the possibility

This feedback was gathered at CSS Cafe 13/06/2026

brechtDR:
No resolution. Two clusters of use cases:

1. **Endpoints matter** (price min/max): click/drag on the track should move the nearest thumb; people do **not** want the whole range to jump.
2. **The interval matters** (calendar block, video in/out, traces): users expect to **translate** the selected range. Clicking outside may move the block rather than only one end.

Concerns: click-to-move-nearest-thumb vs drag-the-segment fight if both are default; 3+ thumbs make a single “the segment” unclear; shaky-pointer / accidental drags.

**Direction to investigate (not specced):** an author-selectable **mode** (name TBD — not `draggable`) so the control is either thumb/endpoint-oriented or range-oriented. Default should follow **research**, not the room’s 50/50 split. If one pattern is ~90% of libraries **and** apps, we can punt the mode.

Also: if range-oriented, click **between** thumbs should not silently pick a thumb (`preventDefault`-ish); click **outside** the interval could still move the closest thumb — needs to stay consistent with drag. Related: #1278.

Keyboard/focus for a “segment” control will be a **separate** issue. If the segment is draggable, it should be focussable

Next for me: survey common libraries vs in-app pickers and report back here. Removing the `agenda+` until that lands.

brechtDR:
## Library vs in-app survey (follow-up to 2026-08-27)

Quick count of **endpoint-oriented** (tap/click moves nearest thumb) vs **range-oriented** (drag the filled interval, translate both ends). This is research for a mode; **not** a resolution on the attribute name.

### Libraries (defaults)

| Library | Default | Opt-in range translate? |
| --- | --- | --- |
[...]

Issue #1339, opened 2025-11-26
[Range] Standardize visual and audio feedback for constrained handle movement.


brechtDR:
Multi-handle slider introduces complex constraints, particularly the `stepbetween` attribute, which enforces a minimum distance between thumbs. When a user tries to move a handle but is blocked by another handle or a boundary (the "max meets min and vice versa" scenario), they require explicit feedback.

I didn't really build something in here in the [prototype](https://brechtdr.github.io/enhanced-range-slider-poc/), however, I do think this needs to be discussed.

I don't have the knowledge of whether something should be done for this, and/or can be done for this. Which is exactly why I wanted to raise this issue. (Hoping on some of the A11Y experts to clarify this a bit).

My personal feeling is that it doesn't really feel broken, not hearing there is a `stepbetween`. at the moment. But it is certainly an inconvenience/annoyance. I could be wrong about this, but in the best case, this is an attribute that would later on be exposed to screen reader users. 

I'd love your insights on the right steps to take.

smhigley:
My main thought would be that this would be handled through having min/max values set on each slider. Those wouldn't ever change for the slider that is currently in focus.

Issue #1337, opened 2025-11-26
[Range] Multi-handle necessity and scalability guidelines (3+ Thumbs)


brechtDR:
While dual-handle usage (min/max price filters) is the most common use case, the foundation of the <rangegroup> API supports adding more than two handles. 

Community feedback provides specific, validated use cases where 3 or more handles are necessary, which confirms the design choice to support N handles, the use cases are however, limited.

## A few Justifications for 3+ Handles:

- **4 Handles:** Opening Hours Selector to manage separate open and close times (e.g., AM shift and PM shift), which is demonstrated in the prototype and mentioned by developers.
- **3+ Handles**: Data visualization tasks, such as setting bin sizes across a histogram, or complex category allocation in budget tracking.
- **N Handles:** Used in Color gradient selectors.

There is, however, a caveat here. Most people I talk to tell me that they haven't used more than two handles, mostly because a library didn't support it, and then another way was easier (this also resulted in the form, a lot of examples with two thumbs, barely any with more). 

However, many are excited that if it were possible, there might be new ways to visualize ranges.

So, I'd love to ask confirmation here. Based on the (currently) limited use cases. Do we want to continue with this idea for an API. 
I feel that this API opens up possibilities that weren't there yet and can be made accessible in a clean way. 

The other way would be a lot of attribute additions, which I feel could get bloated really fast.

The community is positive of the idea to have N-thumbs, but once again use cases are limited.

My main question here is: Is there actually a case **against** the capability of making it more than 2 thumbs based on the current API design?
Should there be a limit (also from an engineering standpoint)? Or is the author's responsibility?

TheJaredWilcurt:
This could help with scheduling events, such as auditions, performances, live concerts, conference talks, etc.

```
   5:30     6:00     6:30     7:00     7:30     8:00              9:00     9:30    10:00
<---|--------|--------|--------|--------|--------|-----------------|--------|--------|----->
  Staff    Band     Setup/   Venue   Opening  Concert           Set Ends  Audience  Venue
 Arrives  Arrives Sound Test Opens     Act     starts                      Leaves   Closes
```

Though a table or bulleted list would be better for communicating this information to attendees, when creating the event, it may be more natural to use a timeline, instead of a stacked series of input fields for times:

```
|  5:30 📆 | Staff Arrives
|  6:00 📆 | Band Arrives
|  6:30 📆 | Setup/Sound Test
|  7:00 📆 | Venue Opens
|  7:30 📆 | Opening Act
[...]

PR #1513, opened 2026-08-28
[Rangegroup] Explainer updates: spec alignment pass + interaction mode

brechtDR:
I did a pass through the explainer against the latest specs and recent CSSWG / Open UI discussions. The core proposal hasn't changed—most of this pass is about tightening the wording, fixing spec citations, and resolving a few internal contradictions.

## Spec Alignment & Wording Fixes

* **Switched CSS Forms links to the Editor's Draft:** We were linking to the TR snapshot while referencing `::slider-track`, `::slider-fill`, and `::slider-thumb`. That snapshot still lists the older `::track`/`::fill`/`::thumb` syntax and flags naming as unresolved. Since CSSWG resolved to keep the `slider-` prefix (csswg-drafts#9830), the explainer now points to the ED and explicitly notes why. Also clarified which pseudos come from CSS Forms versus the ones we're introducing (`::slider-segment`, `::slider-tick`, `::slider-tick-label`).
* **Aligned with the split fill model:** CSSWG recently resolved (19 Feb 2026) to split the fill into `::slider-fill-over` / `::slider-fill-under`, with `::slider-fill` aliasing `under`. Because multi-range handling came up in that discussion, the explainer now maps the two models cleanly: single-thumb controls use over/under, while multi-thumb controls use `::slider-segment(n)`. This also clears up a contradiction where we previously claimed `::slider-fill` styled the space between multi-range thumbs.
* **Clarified form association and `name`:** The `<rangegroup>` `name` attribute was vaguely described as being "for identification." To avoid confusion around form submission, we aligned it with `<fieldset>`: it's form-associated and listed, but not submittable on its own. The group name only appears in `form.elements`, while child inputs handle their own name/value pairs. We added `form` to the attribute list to match and closed the open question on group-level submission for v1 (dual submission risks payload mismatches across browsers and breaks progressive enhancement).
* **Explicit HTML change for `<legend>`:** Using `<legend>` as a direct child of `<rangegroup>` requires updating its HTML content model (which currently only allows `<fieldset>` and `<optgroup>`). We now call this out directly as an explicit spec ask, citing the `<optgroup>` update for customizable `<select>` as precedent, alongside a fallback pattern that validates today.
* **Reframed datalist snapping as proposed:** We previously claimed arrow keys snap to datalist ticks "following existing conventions," but native range controls only snap on pointer input (keyboard navigation is still strictly step-driven). This is now framed as a proposed feature, with an open question on whether it should apply strictly to `rangegroup` or any range with a `list`.

## Quick Corrections & Polish

* **`::-ms-*` prefixes:** Labeled as legacy IE/EdgeHTML (modern Edge runs Chromium and uses WebKit pseudos).
* **`accent-color`:** Noted as the only standardized tint available today.
* **`interactivity` collision:** Clarified that the name collision is with CSS UI 4's `interactivity`, since a CSS `interaction` property doesn't exist.
* **`list` attribute on groups:** Noted that moving `list` to `<rangegroup>` is an explicit ask, and flagged that fallback markup won't render ticks without duplicating the attribute onto child inputs.
* **Vertical orientation:** Since `slider-orientation` is flagged for rework in the ED (csswg-drafts#11891), we no longer present it as a settled solution.
* **Thumb custom property:** Added a note that UA-defined `--slider-thumb-position` can be overridden by authors, so it might need a different shape.
* **Fieldset disabled state:** Added the standard `<legend>` exception inside `<fieldset disabled>`.

## Single-Thumb Handling

Expanded the section on single-thumb groups to explain how they relate to standard `<input type="range">` with `appearance: base`. A group with a single thumb still behaves like a standard single control (one thumb, one filled area, one unfilled area), but gains the shared track coordinate space, group-level attributes (`min`, `max`, `list`), and the group label. This keeps the mapping to `::slider-fill-over` / `::slider-fill-under` consistent. It also means we have a baseline on which we can expand more easily.

## Interaction Modes & Naming

No functional changes here: `endpoints` remains default and `range` is opt-in. We did, however, give the naming caveat its own dedicated paragraph in **Track Click Behavior**. It explicitly lists the collisions (`interactivity`, `inputmode`), notes why `draggable` was dropped (global attribute clash), and makes clear that the core interaction behavior is what matters, regardless of the final attribute name. Discussion stays on #1460.

## Issue References
Updated all citations so claims point directly to their supporting discussions, pulling in the relevant CSSWG threads alongside our existing Open UI issues. Open UI numbers and telecon resolutions remain untouched.


site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -598,32 +738,31 @@ rangegroup::slider-segment(4) {
...
+- **Keyboard navigation** ([#1185](https://github.com/openui/open-ui/issues/1185)): Each thumb gets a tab stop. Arrow keys, Page Up/Down, Home, and End behave as they do on a standard `<input type="range">`. See the [Keyboard Interaction](#keyboard-interaction) section.
+- **Disabled attribute propagation** ([#1338](https://github.com/openui/open-ui/issues/1338)): `disabled` on `<rangegroup>` cascades to all thumbs. Individual thumbs can be disabled independently. When all thumbs are disabled, the group behaves as fully disabled. See the [Disabled State](#disabled-state) section.
+- **Screen reader announcements** ([#1339](https://github.com/openui/open-ui/issues/1339)): The group's min and max should be announced when entering the group (first thumb), plus `stepbetween` when it applies. Each thumb then announces its own min, max, and current value. See the [Screen Reader Announcements](#screen-reader-announcements) section.
```
anaskim, line 746: This issue remains open, should we close it as it resolved?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -474,6 +602,8 @@ For groups specifically we'd like to introduce a few new ones:
 - `inputs`: A property that returns a collection of the contained range input elements. A shorthand for `querySelectorAll`
 - `getRangeInput(index)`: Returns the range input at the specified index.
 - `setRangeValue(index, value)`: Sets the value for a specific handle.
+- `addThumb(value, options?)`: Adds a new handle to the group at the given `value` and returns the created `<input type="range">`. Because handles are backed by real `<input type="range">` elements, this is a convenience over manually creating and appending an input. The optional `options` object accepts `min`, `max`, `name`, and `label`. Use cases such as a gradient editor that lets users add and remove color stops on demand are a primary motivation for this method.
+- `removeThumb(index)`: Removes the handle (and its backing `<input>`) at the specified index.
```
anaskim, line 606: Do `addThumb` and `removeThumb` have valid workarounds or are these APIs absolutely necessary for a specific case? If they're just helpers that could be written in JS for now, then I would rather add this as an enhancement later.

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -331,9 +477,11 @@ The number of segments is determined by the number of thumbs.
 
 - `min`: Specifies the minimum value for the entire range group, defining the lower bound of the track.
 - `max`: Specifies the maximum value for the entire range group, defining the upper bound of the track.
-- `name`: The name of the range group, used for identification.
-- `list`: Links the range group to a `<datalist>` element, providing tick marks or predefined values.
-- `stepbetween`: Defines the minimum distance between handles in a range group.
+- `name`: The name of the range group, used in the `form.elements` API. It is not submitted, matching `name` on `<fieldset>`. See [Form Association and Naming](#form-association-and-naming).
+- `form`: Associates the range group with a form element explicitly, matching `form` on `<fieldset>`.
```
anaskim, line 481: Since fieldset’s `form` attribute does not propagate to descendant controls, should we mention that a detached group’s child ranges still need their own `form` attributes to be submitted to that form?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -2,11 +2,29 @@
...
+- [Background](#background)
+- [Goals](#goals)
+- [Current State of Range Styling](#current-state-of-range-styling)
+- [Multi-Handle Support with `<rangegroup>`](#multi-handle-support-with-rangegroup)
+- [Form Association and Naming](#form-association-and-naming)
+- [Accessibility Enhancements](#accessibility-enhancements)
+- [Disabled State](#disabled-state)
+- [Progressive Enhancement](#progressive-enhancement)
+- [Datalist Integration](#datalist-integration)
+- [Multi-Color Range Segments](#multi-color-slider-segments)
```
anaskim, line 21:
```suggestion
- [Multi-Color Range Segments](#multi-color-range-segments)
```

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -321,7 +413,61 @@ rangegroup::slider-segment(3) {
...
+A circular control is the same technique with a closed path. A single-thumb circular input behaves as a rotary knob; multi-thumb `<rangegroup>` controls become radial range selectors.
+
+```css
+rangegroup {
+    appearance: base;
+    inline-size: 240px;
+    aspect-ratio: 1;
+}
+
+rangegroup::slider-thumb {
```
anaskim, line 458: Which element owns each thumb pseudo: the `rangegroup` or the child input? Some example use `rangegroup::slider-thumb`, while this one uses `rangegroup input::slider-thumb`. From the definition later: "- `::slider-thumb`: Represents the draggable handles within the group." I was expecting this pseudoclass to apply to all handles so the host element would be `rangegroup` but looks like the host element is the `input`. Am I missing something?

anaskim (commented):
Still reading through it, but posting my initial comments :)

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: Are you talking about #1337 ? In that case we could keep it open because its resolution was to gather more 3+ thumb examples and feedback, not to settle the two-thumb-vs-N-thumb API question.

Personally I think the "more than two question" has been answered a few times. There are quite a few ideas in the proof of concept as well. 

I'm open to closing (but saying it carefully)

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: Yes, both have valid DOM workarounds because the handles are real child inputs (append() / remove()). They are not required for v1. I included them to make the potential future API concrete; would it be blocking to keep them explicitly labeled as deferred/future API so we have a place to pinpoint that discussion, or would you prefer they be removed until separately proposed?

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: Correct. The group’s form attribute associates only the group, just like `<fieldset>`; detached child ranges need their own form attributes to submit. I’ll state this explicitly and add an example

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: The intended proposal is that `<rangegroup>` owns the shared visual anatomy, including its generated thumb pseudos; the child inputs remain the semantic/form controls. The examples seem to be a bit inconsistent. I'll try to improve on that.

site/src/pages/components/enhanced-range-input.explainer.mdx
anaskim: I think we could keep them as potential future APIs and open an issue to discuss them as a future enhancement of the feature that could improve developer ergonomics.

site/src/pages/components/enhanced-range-input.explainer.mdx
anaskim: Ah sorry, was referring to #1339 which already has a resolution posted. GitHub isn't great at highlighting the commented line :(

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -331,9 +477,11 @@ The number of segments is determined by the number of thumbs.
...
 - `max`: Specifies the maximum value for the entire range group, defining the upper bound of the track.
-- `name`: The name of the range group, used for identification.
-- `list`: Links the range group to a `<datalist>` element, providing tick marks or predefined values.
-- `stepbetween`: Defines the minimum distance between handles in a range group.
+- `name`: The name of the range group, used in the `form.elements` API. It is not submitted, matching `name` on `<fieldset>`. See [Form Association and Naming](#form-association-and-naming).
+- `form`: Associates the range group with a form element explicitly, matching `form` on `<fieldset>`.
+- `list`: Links the range group to a `<datalist>` element, providing tick marks or predefined values. This extends `list`, which today is only valid on `<input>`.
+- `stepbetween`: Defines the minimum distance between **adjacent** handles. It is a single value on the group, not a per-gap list. It has no effect when the group contains fewer than two handles. A maximum distance between handles is intentionally not in v1 (see [Resolved Decisions](#resolved-decisions)).
```
anaskim, line 464: Curious, maybe I missed it somewhere along the conversations that you've already had over the years, but I'm not sure what "adjacent" means in this case. Is it just child DOM order?

anaskim: Maybe somewhat related to this, is there a way where an author can get two thumbs to cross each other?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -99,20 +128,24 @@ This opt-in approach allows developers to access enhanced styling capabilities w
 
 ## Multi-Handle Support with `<rangegroup>`
 
-Based on community feedback and discussions, we propose a new approach for multi-handle functionality. Instead of adding a new attribute to the existing `<input type="range">` element or adding a new type, we believe a new `<rangegroup>` element could offer us the correct flexibility.
-This would be a wrapper element that can contain multiple `<input type="range">` elements.
+Based on community feedback and discussions, we propose a new `<rangegroup>` element rather than a new attribute or a second range input type. It is a wrapper that contains one or more `<input type="range">` elements and paints them as a single slider on a shared track.
```
anaskim, line 112: This made me think the following: what happens when `<rangegroup>` has zero children?

anaskim (commented):
Thanks for the updates, some more questions. Feel free to point me to separate conversations that you've had. We can also continue the discussion on Discord, I don't think my comments are necessarily blocking for this PR.

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: This is a good question. Based on how it is described, it should behave the same as an empty fieldset.

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: 1 yes, and 2 is no crossing by default.
The crossing complicates labelling a lot. If a thumb has a label "min price" on the left handle and it gets crossed over to the right, it might not make sense anymore, especially for SR users.

Unless there is a way to solve this from an engineering perspective; in that case, I'd love to hear about it :)

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: I created an issue just so we don't forget:
#1515

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: Updated the explainer to reflect that

anaskim (approved):

lukewarlow:
Gonna Take Ana's review as having done the heavy lifting. I'll try to put some time aside to read the full explainer and raise issues where I think we need wider discussion. But won't block this obviously.

lukewarlow (approved):

brechtDR:
> Gonna Take Ana's review as having done the heavy lifting. I'll try to put some time aside to read the full explainer and raise issues where I think we need wider discussion. But won't block this obviously.

Hey @lukewarlow, that would be fantastic to move this forward! I'm sure many questions will rise. Thank you for this.

Issue #1419, opened 2026-03-13
[toolbar] should disabled toolbar items remain focusable


lukewarlow:
It says on MDN:

"If any of the otherwise interactive elements within the toolbar are temporarily disabled, consider letting them remain focusable so screen reader users can be made aware of their presence."

Should we provide something to achieve this?

scottaohara:
i was going to comment on [this menuitem issue](https://github.com/openui/open-ui/issues/1274) the other day, but never got around to it.  

so now i can just do it here and say that i'd personally rather there be a new feature for "soft disabling" controls, rather than have the disabled attribute work differently depending on what context its used in - and thus also making it more work for someone to achieve standard disabling behavior, if they wanted to do that.

The guidance to provide soft disabling of controls, as mentioned by the apg, also talks about how its important to provide consistent behavior for disabled states.  hence, imo why it should be the choice of the author whether or not controls can be disabled, but focusable, and also it shouldn't necessarily be limited to just elements within a menu or toolbar.  E.g., why not within a select/listbox?

sb3nder:
Maybe we could have a new HTML attribute `passive`/`inactive`... which keeps elements focusable but prevents them from being activated?

sb3nder:
I was wondering how we can achieve inoperable focusable elements.
The options I can think of:

1. nothing - devs have to prevent activation and mutation like they do now,
for example `aria-disabled="true" onclick="event.preventDefault();"`
2. a new html attribute `passive`/`inactive`...
3. an html modifier for `disabled`, `disabled=focusable`...
4. `disabled` + a new html attribute
5. `disabled` + a new html attribute placed on the parent
6. ~~allow `readonly` to be used for more elements, and change css `:read-only` to match only elements that actually have the attribute... JK I don't want to break the websites that rely on `:read-only` matching everything... 🙄~~

In the case below, one would want the button to **not** be focusable?
```html
<toolbar disabled><button softdisabledsomehow>btn</button></toolbar>
```

What others good and bad ideas I'm missing?

Similar or related things I found in the wild:
https://github.com/whatwg/html/issues/11363
https://github.com/whatwg/html/issues/10125
https://github.com/w3c/aria/issues/1309
https://github.com/w3c/aria-practices/pull/3387
https://github.com/whatwg/html/issues/2311
https://github.com/whatwg/html/pull/12219

PR #1514, opened 2026-08-31
Menu aria roles

davidsgrogan:


site/src/pages/components/menu.explainer.mdx
```diff
@@ -360,41 +354,24 @@ and best practices curated from accessibility experts as closely as possible; de
...
+   - While the `<menuitem>` element's native ARIA role defaults to `menuitem`, it is responsive to whether the element is embedded in a `<fieldset checkable=single|multiple>`. The [`menuitemradio`](https://w3c.github.io/aria/#menuitemradio) role is exposed for items in a `<fieldset checkable=single>`, and the [`menuitemcheckbox`](https://w3c.github.io/aria/#menuitemcheckbox) role is exposed for items in a `<fieldset checkable=multiple>`.
+   - `menuitemradio` and `menuitemcheckbox` elements expose a boolean `checked` state.
+5. `menuitem`s and `button`s that are `menu` invokers:
+   - have `expanded=true` or `expanded=false` state.
+   - have `haspopup=menu`.
+   - will NOT have `details` pointing to the `menu` ([past discussion on `haspopup` instead of `details`](https://github.com/openui/open-ui/issues/1297#issuecomment-3376865931)).
+6. `menubar` and `menu` have `setSize`.
+7. `menuitem` descendants of a `menubar` or `menu` have `setSize` and `posInSet`.
+8. `menuitem`, `menuitemcheckbox`, and `menuitemradio` are included in the accessibility tree even when they are disabled.
```
scottaohara, line 373: the 'even' in this makes it seem like there's something special going on here - but disabled items should always be in the a11y tree.

is this trying to allude to the soft disabling concept from the previous draft?  if so, then the fact that someone could use arrow keys to navigate to the disabled items isn't clear anymore. From my quick read that isn't mentioned in the keyboard behavior section either.  so is that behavior not going forward?

davidsgrogan: > disabled items should always be in the a11y tree.

Ah, I didn't know that, but it makes sense. This "even" sentence was here because Blink previously had a bug where these elements being disabled meant they were omitted from the a11y tree. I removed "even" and instead mention aria-disabled. WDYT? Should I just remove this item altogether?

As for soft-disabling, no, we aren't planning to do soft disabling, in large part due to your opinion at https://github.com/openui/open-ui/issues/1419#issuecomment-4054782968 :)

PR #1518, opened 2026-09-17
[rangegroup] Add non-linear scale information and empty state

brechtDR:


site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -372,6 +391,40 @@ Note that `list` on the group alone does not degrade: in a non-supporting browse
...
 Whether the group should instead inherit `list` down to its children automatically, so the duplication is unnecessary, is an open question.
 
+### Non-linear value scales
+
+A recurring request is a range whose value does not grow evenly along the track ([#1464](https://github.com/openui/open-ui/issues/1464)). A price filter from 1 to 1,000,000 is the classic case: on a linear scale everything below 500,000 sits in the left half, and a one-pixel drag near the middle jumps by thousands, so the values most people actually want are unreachable. The same shape of problem turns up in audio frequency and gain, zoom levels, file sizes, and anything spanning several orders of magnitude.
+
+The [10 September 2026 telecon](https://www.w3.org/2026/09/10-openui-minutes.html) resolved to first build this out of what a range input can already express, namely a `<datalist>` with unevenly spaced options, and to document the result here rather than to introduce a scale attribute now:
+
+```html
+<rangegroup name="budget" min="0" max="1000000" list="budget-stops">
```
anaskim, line 401: In this example, should we we repeat `min`, `max`, and `list` on the child input? Otherwise, it wouldn't preserve its domain or ticks in a non-supporting browser. This seems important because the final paragraph states the linear native fallback.

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -158,6 +158,25 @@ In a supporting browser this still renders as the `<rangegroup>` slider (one thu
...
+```html
+<rangegroup min="0" max="100">
+    <legend>Stops</legend>
+</rangegroup>
+```
+
+Specifically:
+
+- The control is **not** hidden and is **not** rendered in its disabled state. `disabled` remains an author-controlled attribute; an empty group is simply a group that has nothing to drag yet.
+- There are no thumbs, so there are no tab stops inside the group, and pointer interaction on the track does nothing. There is no nearest thumb to move and no segment to translate, so `interaction` has no observable effect.
```
anaskim, line 174: Should we also say whether the named group remains exposed to accessibility APIs with no slider children?

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -194,7 +213,7 @@ A robust labeling strategy is crucial for users of assistive technologies. We pr
...
 
-Until that change lands, a `<legend>` inside `<rangegroup>` is invalid markup and, in a non-supporting browser, is not associated with anything. Authors who need markup that validates today can wrap the group in a `<fieldset>` and put the `<legend>` there instead; the group is then named through `aria-labelledby` on the `<rangegroup>`. The examples in this explainer use the proposed form.
+Until that change lands, a `<legend>` inside `<rangegroup>` is invalid markup and, in a non-supporting browser, is not associated with anything. Authors who need markup that validates today can wrap the group in a `<fieldset>` and put the `<legend>` there instead; the group is then named through `aria-labelledby` on the `<rangegroup>`. The examples in this explainer use the proposed form.
```
anaskim, line 216:
```suggestion
Until that change lands, a `<legend>` inside `<rangegroup>` is invalid markup and, in a non-supporting browser, is not associated with anything. Authors who need markup that validates today can wrap the group in a `<fieldset>` and put the `<legend>` there instead; the group is then named through `aria-labelledby` on the `<rangegroup>`. The examples in this explainer use the proposed form.
```

site/src/pages/components/enhanced-range-input.explainer.mdx
```diff
@@ -503,11 +556,21 @@ This approach offers several benefits:
...
+- Dragging a thumb directly still adjusts that one endpoint, exactly as in endpoint-oriented mode.
+
+This was [resolved](https://www.w3.org/2026/09/10-openui-minutes.html) on 10 September 2026 ([#1460](https://github.com/openui/open-ui/issues/1460)). The point of removing the nearest-thumb fallback entirely is to make the two keywords **two cleanly distinct experiences** rather than two variations that share most of their behavior, so they can be put in front of users as an either/or choice. Which of them should be the default is a question for that testing, not something this explainer settles.
+
+A group with fewer than two thumbs has no segment, so `interaction="range"` has nothing to translate. Such a group keeps endpoint-oriented behavior: a track press moves the single thumb, and an empty group does nothing at all.
```
anaskim, line 569: I'm a bit confused by the use of "segments" here. I see two definitions:

1. Portion of the track
2. Draggable interval

The `::slider-segment()` states that empty groups have 1 segment. Here we say that groups with fewer than two thumbs have no segment. Am I misreading it?

anaskim (commented):
First pass review :)

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: That's a good point. Could that be optional? I feel for progressive enhancement reasons this would certainly be a good idea. However for times where progressive enhancement wouldn't be needed anymore it might be a bit too much?

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: I'll try to clear this up. For styling purposes a segment exists from point 0 to the first thumb as well. I'll try to state things clearer here.

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: Addressed in the latest push. The child now repeats `min`, `max`, and `list` so the fallback keeps the domain and ticks (without them `value="10000"` would clamp to 100 on a default 0–100 slider).

On whether the duplication can be optional: in a supporting browser it is already inert — identical child bounds intersect with the group's to the same range, and the group's `list` takes precedence. So the duplication is purely a progressive-enhancement technique. I extended the existing open question so that if the group eventually inherits `list` / `min` / `max` down to children that omit them, authors only need the duplication while unsupporting browsers still matter.

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: Addressed. The empty-group section now states that the named group stays in the accessibility tree as a `group` (like an empty `<fieldset>`), is not pruned, is not exposed as a slider, is not a Tab stop, and is not `aria-disabled`. The HTML `name` attribute is not an accessible name — a group that matters before it has thumbs still needs a legend. Tick marks from `list` still render because they belong to the track.

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: Accepted — trailing whitespace removed.

site/src/pages/components/enhanced-range-input.explainer.mdx
brechtDR: You were reading it correctly — the document defined "segment" twice and the two definitions disagreed.

**Resolution:** "segment" stays the CSS paint region (`::slider-segment(n)`, *N* thumbs → *N*+1 segments). The draggable two-thumb region is now called an **interval** (*N* thumbs → `max(N−1, 0)` intervals). That split is defined once in Multi-Color Range Segments with a sketch, and every interaction call site was updated.

Related fix from the same confusion: the draft said a group with fewer than two thumbs "keeps endpoint-oriented behavior", which quietly reintroduced the hybrid the 10 Sep resolution removed. `interaction="range"` with no interval (fewer than two **enabled** thumbs) is now inert — track press does nothing; direct thumb drag and keyboard stepping still work. Authors who want track-click on a single-thumb control want `endpoints`.

brechtDR: I went for interval as a region between two thumbs to clarify. If there is better wording for this, feel free to suggest something. But yes, this got confusing.

